PR #859, opened 2013-02-19
Add support for testing C API calls directly

carsonmcdonald:
I'm not sure if this is needed or not but I figured I would see. I noticed the other day that there was no way to test str_to_cstr. This adds support for compiling/linking c source in test/t into the test binary. I've added an example for str_to_cstr.


matz:
Closing without merge — the policy I stated in this thread on 2013-09-18 still holds: core mruby C API is tested via Ruby, while C-level testing belongs in mrbgems.

The mechanism this PR proposed (compile `test/*.c` and link into the test driver) is in fact already supported at the mrbgem layer: see `g.test_objs` handling in `mrbgems/mruby-test/mrbgem.rake` and the existing C-side tests in `mrbgems/mruby-{dir,error,socket,fiber,io,...}/test/*.c`. So the underlying need has been met, just outside `test/t/`.

Thanks for the contribution.

PR #874, opened 2013-02-23
Fix handling of second argument of Regexp.new.

masamitsu-murase:
I think that second argument of Regexp.new should be Integer according to ISO spec.


matz:
Option check has been added.

PR #880, opened 2013-02-23
Refere ISEQ BLOCK of binary bytecode to ROM directly(rebased).

kyab:
I've implemented idea of referring ISEQ BLOCK of binary bytecode to ROM directly. Based on (6) of http://www.csk.com/fukuoka/services/mruby/nxt02.html.

I'm not sure this kind of hack is wellcome, but in my environment, around 8kb of heap can be saved by this change.


monaka:
IMO, this issue can be closed as the second suggestion is on #1338.     @kyab WDYT?

kyab:
I think this issue can be closed.

PR #929, opened 2013-03-02
mruby-time: add check to prevent invalid time

cremno:
``` ruby
Time.at(-1)  # => ïD$♦ ░`☻ Sat -1 -1:-1:-1 1899
Time.at(100) - 101  # => ïD$♦ ░`☻ Sat -1 -1:-1:-1 1899
```

I added a check to mrb_time_make that'll raise an ArgumentError instead.


matz:
We are still making effort to use negative `time_t` if possible, to represent my birthday time (I am an old man).

PR #1319, opened 2013-06-28
Fix incorrect debug filename for bytecode from multiple *.rb

### LINKED ISSUE #1243: show a backtrace when when an error occurs on gem load
Create a gem with this ruby code:

``` ruby
ezezeze
```

This code is obviously invalid but the problem is that when you will try executing mruby you will get no backtrace at all, only a very unhelpful:

``` text
NoMethodError: undefined method 'ezezeze' for main
```

I gave it a try but I failed, as far as I understand it the generated gem_init.c is the culprit and should show a backtrace when an error occurs during loading, as a test my first try was to duplicate the showcallinfo method from main.c in gem_init.c and add calls to it on error.
I suppose it misses something since all I get is an empty backtrace...

Debugging is currently really annoying due to this problem, I could try to make a patch if someone points me in the right direction, right now I am clueless as to what is required.

/cc @bovi

kyab:
Fix #1243

With this fix bytecode from multiple *.rb with `mrbc -g` shows correct filename.
I hope someone to review.

``` sh
[koji@macbookpro:~/work/mruby]$ cat one.rb
puts "hi from one.rb"

non_defined_func

[koji@macbookpro:~/work/mruby]$ cat two.rb
puts "hi from two.rb"

[koji@macbookpro:~/work/mruby]$ mrbc -g -o prog.mrb one.rb two.rb
[koji@macbookpro:~/work/mruby]$ mruby -b prog.mrb
hi from one.rb
trace:
    [0] one.rb:3
one.rb:3: undefined method 'non_defined_func' for main (NoMethodError)
```

### DIFF
--- a/include/mruby/irep.h
+++ b/include/mruby/irep.h
@@ -32,6 +32,7 @@ typedef struct mrb_irep {
 
 mrb_irep *mrb_add_irep(mrb_state *mrb);
 mrb_value mrb_load_irep(mrb_state*, const uint8_t*);
+int mrb_irep_stop_p(mrb_state *mrb, mrb_irep *irep);
 
 #if defined(__cplusplus)
 }  /* extern "C" { */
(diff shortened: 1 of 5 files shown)


matz:
I confirmed this is fixed.

Issue #944, opened 2013-03-04
[RFC] Update .mrb file format


masuidrive:
It's proposal of new .mrb file format.
Current .mrb file is hex format. It's too fat.

New .mrb file can have sections. now it have irep section only. I'll add debug section.
It have #880 in mind. The irep section has 'endianness' field and can choice big/little endian in the irep section.

I'm working on https://github.com/masuidrive/mruby/tree/binary

![mrb file format](https://f.cloud.github.com/assets/12104/217139/40c5b978-84e1-11e2-8927-f9da055e8af8.png)

matz:
Closing — the substantive parts of your RFC are now in place. The current `.mrb` format (`RITE 0400`) is:

- Binary, not the old hex format
- Section-based: `IREP` (bytecode), `DBG\0` (debug info), `LVAR` (local variable names), `END\0` (terminator)
- Endian-neutral on disk: multi-byte fields are stored big-endian and converted on load via `bin_to_uint*`

CRC was discussed in this thread but ultimately not added. Once mruby settled on the pattern of embedding bytecode directly in C source via `mrbc -B`, the transmission scenarios where CRC matters became uncommon, and the cost wasn't worth paying for the typical ROM-embedded use case.

Thanks for the careful original RFC and the diagrams — the format you sketched is essentially what we ended up with.

PR #1338, opened 2013-07-03
ROM-able ISEQ revisited

kyab:
This is second version of #880. Related issues are #944, #964.

In my environment(using default gembox, clang, x64),
total heap allocation till end of mrb_open() was reduced around 13kb (183514bytes to 170654bytes).
- Bytecode format now has endian field.
- mrbc has new option `-E<endian>`. Only effective with -B(C-func).
- Default for `-E` is "neutral", which means copy all iseq(same as current). 
- In build_config, you can specify  `endian :little`, or `endian :big` in each target. This value is used for `-E` option on running mrbc.
- Currently CRC are checked even for C-func. I would update to skip CRC check if it is OK.
- I do not have Big-endian device. Only little endian was tested.

test code.

``` c
#include <stdio.h>
#include <stdlib.h>

#include "mruby.h"

size_t total = 0;

void *myallocf(mrb_state *mrb, void *p, size_t size, void *ud){
    if (size == 0){
        free(p);
        return NULL;
    }else{
        //printf("alloc size=%ld, p=%p\n",size,p);
        total += size;
        void *ret = realloc(p, size);
        return ret;
    }
}

int main(int argc, const char * argv[])
{
    mrb_state *mrb = mrb_open_allocf(myallocf, NULL);
    printf("total : %zu bytes\n", total);
    mrb_close(mrb);

    return 0;
}

```

### DIFF
--- a/include/mruby/dump.h
+++ b/include/mruby/dump.h
@@ -15,7 +15,7 @@ extern "C" {
 
 #ifdef ENABLE_STDIO
 int mrb_dump_irep_binary(mrb_state*, size_t, int, FILE*);
-int mrb_dump_irep_cfunc(mrb_state *mrb, size_t n, int, FILE *f, const char *initname);
+int mrb_dump_irep_cfunc(mrb_state *mrb, size_t n, int, int, FILE *f, const char *initname);
 int32_t mrb_read_irep_file(mrb_state*, FILE*);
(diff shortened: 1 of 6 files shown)


matz:
`mruby` iseq are now ROM-able.

PR #1351, opened 2013-07-07
Add cast to suppress -Wsign-compare warnings

k-tsj:


matz:
Those issues are already fixed.

PR #1456, opened 2013-08-11
trigger gc test in travis build

flaneur2020:
this pr allow us to trigger a GC test if GC_TEST is enabled while running `rake test`


zzak:
How is this different from `MRB_GC_STRESS`?

Also, why use `fprint()` over `puts()`?

PR #1457, opened 2013-08-11
I implement super in aliased methods.

suzukaze:
This test code super-in-aliased-method.rb:

``` ruby
class Name

  def name
    return "super-name"
  end

end

class Human < Name

  def name
    return "James-#{super}"
  end

  alias get_name name
  # Or
  #alias_method :get_name, :name
end

human = Human.new
puts "name:#{human.name}\n"
puts "get_name:#{human.get_name}\n"
```

Don't impelment super in aliased methods:

```
$ mruby super-in-aliased-method.rb 
name:James-super-name
trace:
    [1] super-in-aliased-method.rb:12:in Human.get_name
    [0] super-in-aliased-method.rb:22
super-in-aliased-method.rb:12: undefined method 'get_name' for #<Human:0x7fdfc4006470> (NoMethodError)
```

This issue is that super call get_name method of super class in aliased method.
Correctly super should call name method.

Impelment super in aliased methods:

```
$ mruby super-in-aliased-method.rb 
name:James-super-name
get_name:James-super-name
```

super calls name method.

### DIFF
--- a/include/mruby/proc.h
+++ b/include/mruby/proc.h
@@ -40,8 +40,8 @@ struct RProc {
 #define MRB_ASPEC_KDICT(a)        ((a) & (1<<1))
 #define MRB_ASPEC_BLOCK(a)        ((a) & 1)
 
-#define MRB_PROC_METHOD_ALIAS 64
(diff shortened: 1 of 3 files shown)


matz:
Fixed by #4718

PR #1476, opened 2013-08-15
Suggestion for different data API names

carsonmcdonald:
This is a suggestion to make the naming of the data API calls more consistent with their function. This change renames GET_PTR to GET_CHECKED_PTR and CHECK_GET_PTR to GET_UNCHECKED_PTR. It seems like those read better than the previous names since they more directly describe what you are asking to happen when you call them.

### DIFF
--- a/include/mruby/data.h
+++ b/include/mruby/data.h
@@ -38,10 +38,10 @@ struct RData *mrb_data_object_alloc(mrb_state *mrb, struct RClass* klass, void *
 #define DATA_PTR(d)        (RDATA(d)->data)
 #define DATA_TYPE(d)       (RDATA(d)->type)
 void mrb_data_check_type(mrb_state *mrb, mrb_value, const mrb_data_type*);
-void *mrb_data_get_ptr(mrb_state *mrb, mrb_value, const mrb_data_type*);
-#define DATA_GET_PTR(mrb,obj,dtype,type) (type*)mrb_data_get_ptr(mrb,obj,dtype)
-void *mrb_data_check_get_ptr(mrb_state *mrb, mrb_value, const mrb_data_type*);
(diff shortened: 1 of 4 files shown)


matz:
Thank you for this PR — looking back at it now, 12 years after filing.

Closing it without merge. The reasoning is essentially what I noted in 2013, plus the strong signal of 12 years of inactivity:

1. **`DATA_CHECK_GET_PTR` does check the type** (it just returns NULL on type mismatch instead of raising). The proposed rename to `GET_UNCHECKED_PTR` would misname the actual behavior, as I pointed out at the time.

2. **The current names accurately describe the behavior:** `DATA_GET_PTR` gets the pointer (raising on type error), `DATA_CHECK_GET_PTR` does a checked-get that returns NULL on type error. The verb-prefix "CHECK_" is conventional Ruby/CRuby vocabulary for "the safe non-raising variant" — see CRuby's `rb_check_typeddata` etc. So even outside of mruby, the existing naming is in the established idiom.

3. **Renaming public API macros is a breaking change for external embedders** — every mrbgem author and every project that embeds mruby would need to update. The cost is concrete; the benefit is subjective.

4. **Master has held the original names unchanged for 12 years**, with widespread current use across `mruby-regexp`, `mruby-random`, `mruby-time`, etc. That's the strongest possible signal that the existing names work fine in practice.

Your alternative proposal in the discussion (`DATA_HAS_PTR_OF_TYPE` + always-raising `DATA_GET_PTR`) is more interesting structurally but introduces double type-checking at every call site (check + get), and would require updating dozens of callers. That trade-off also didn't tip in favor of the change.

Closing in favor of "the names we have are the names we'll keep." Thanks again for the suggestion — these naming discussions sharpen everyone's understanding even when the conclusion is to leave things as they are.

PR #1767, opened 2014-02-28
Add mruby-sharedlib : Shared library

mattn:
mrbgem for making libmruby.so, making mruby.dll on windows.


matz:
Thank you for this PR. Closing it after re-reviewing — 11 years of inactivity and the approach has aged out of the project's preferred direction.

Two reasons:

1. **The `mruby.def` approach is fragile.** As @kyab noted in 2014, manually maintaining a 290-line export-list file means non-Windows contributors will forget to add/remove entries when they touch the API. After 11 years, this concern has only grown — the C API has changed substantially. `__declspec(dllexport)` annotations on individual `MRB_API` functions would be more robust, but that's a different design that would need its own PR.

2. **A more current effort exists** in #6239, where the discussion has moved to a `host-shared.rb`-based approach that addresses both POSIX `.so` and (eventually) Windows DLL needs. @dearblue has a PoC branch (`dearblue:poc/build/shared-libmruby`) that handles soname, rpath, version scripts, and `LD_LIBRARY_PATH` for tests — closer to what Linux distros need.

Closing in favor of the discussion in #6239. Thanks again for the original work.

kyab:
Wow, I’m glad @matz explained it so carefully.

PR #1769, opened 2014-02-28
build_mrbconf.h

take-cheeze:
With this patch mruby build system can separate `mrbconf.h` macro and compiler flag macro.
To use `mrbconf.h` macro use `MRuby::Build#enable_mrbconf` method to enable specific macro easily instead of appending to `defines` property.
(if you need to enable `MRB_GC_FIXED_ARENA` call `enable_mrbconf :gc_fixed_arena`)

Additionally when you change **mrbconf** in **build_config.rb** the whole code would be recompiled with new **mrbconf** so that it's safer.

Though it has some limitation:
- it can't specify macro with value
- it can't disable mrbconf
- it can't detect conflict of mrbconfs

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,6 +29,9 @@ load "#{MRUBY_ROOT}/tasks/libmruby.rake"
 load "#{MRUBY_ROOT}/tasks/mrbgems_test.rake"
 load "#{MRUBY_ROOT}/test/mrbtest.rake"
 
+# generate build specific mrbconf.h (build_mrbconf.h)
+load "#{MRUBY_ROOT}/tasks/build_mrbconf.rake"
+
 ##############################
 # generic build targets, rules
 task :default => :all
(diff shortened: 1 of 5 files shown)


zzak:
I'm not sure we should allow building separate configs, as this also complicates the build system.

Is there another way we could introduce this feature outside of the current scope?

matz:
Closing without merge — the design tradeoff hasn't aged well.

The author's own note in this thread (2014-03-01) made the core problem explicit: differently-mrbconfed builds are ABI-incompatible, since `mrb_state` layout depends on flags like `MRB_GC_FIXED_ARENA`. Routing those macros through a generated `build/<target>/include/build_mrbconf.h` that user code never sees makes mismatch *easier* to introduce, not harder. The PR body also lists three unresolved limitations: no value macros, no disable, no conflict detection.

mruby's build model assumes libmruby.a is built alongside the host program from a single `build_config.rb` rather than installed, so the extra indirection adds maintenance cost without solving a real problem.

Thanks for the contribution.

PR #1822, opened 2014-03-07
Add mrb_utf8_from_locale, mrb_utf8_free, mrb_locale_from_utf8, mrb_locale_free

mattn:
Add mrb_cstr_from_locale/mrb_cstr_to_locale. ARGV should be utf8 strings converted from locale strings. And __printstr__ should print locale strings.


mattn:
rebased.

PR #1835, opened 2014-03-08
Support private and protected visibility.

take-cheeze:
- This will break compatibility with old mruby irep since opcode `OP_FSEND` and `OP_FSENDB` that won't be supported by old VM would be used to check private call.
- `mrb_call_type` would be stored to `struct RClass` and `struct RProc` in 10th and 11th bit of `flags` using `0x600` bit mask. `RClass`'s `mrb_call_type` flags would be cleared if class scope is opened.
- Method such as `private_instance_methods` and `protected_instance_methods` isn't implemented yet.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -360,12 +360,12 @@ void mrb_gc_protect(mrb_state *mrb, mrb_value obj);
 mrb_value mrb_to_int(mrb_state *mrb, mrb_value val);
 void mrb_check_type(mrb_state *mrb, mrb_value x, enum mrb_vtype t);
 
-typedef enum call_type {
-    CALL_PUBLIC,
-    CALL_FCALL,
-    CALL_VCALL,
-    CALL_TYPE_MAX
-} call_type;
+typedef enum mrb_call_type {
+    MRB_CALL_PUBLIC,
+    MRB_CALL_PRIVATE, MRB_CALL_FCALL = MRB_CALL_PRIVATE,
+    MRB_CALL_PROTECTED, MRB_CALL_VCALL = MRB_CALL_PROTECTED,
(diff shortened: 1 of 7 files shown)


jbreeden:
Would love to see this support in mruby. Right now (using the IO mrbgem) I'm getting funny behavior like...

```Ruby
# If this function fails and returns nil...
stream = get_stream_somehow

# ...this blocks forever trying to read from stdin
# since nil doesn't have a `gets` method, and we hit
# the would-be-private Kernel.gets
stream.gets
```

Are there plans to support this? Or should I accept it as fact and find a way around it?

zzak:
You should be able to work-around this using IO#closed?

stream = get_stream_somehow
stream.gets unless stream.closed?

jbreeden:
True enough. I could check for nil on the returned stream, or check the state of the stream

I was more hoping for a solution that made `nil.gets` fail. In case the client of my library forgets to do a nil check, I would think that throwing an exception would make there error more obvious than having their program block or read from the wrong stream.

For this I could check self and throw if it's nil. There are probably better options as well... But if private/protected is going to be supported I'd just as soon wait for that.

(Thanks for the suggestion, btw!)

cremno:
Another workaround, that may not be acceptable to you, is:
```ruby
undef gets
```
You or your project's users then have to explicitly write the receiver:
```ruby
STDIN.gets
```

jbreeden:
That's a good idea too. If I wanted to assurance for all the kernel methods inherited by objects though, I'd have to `undef` all of them. Can't imagine any consumers of the lib would appreciate that :P

Considering defining the IO methods on NilClass to throw...

Anyway, I just wanted to +1 this request, and ask about support.

PR #2152, opened 2014-04-29
Add argument pattern test

mirichi:
dedicate to OP_ENTER


matz:
Adding obvious tests only makes test execution time longer.

PR #2309, opened 2014-05-23
Remove `LINENO` section.

take-cheeze:
Since it isn't used anymore.
It shouldn't break compatibility with 1.0.0. Though may cause problem with mruby older than 1.0.0.


matz:
`LINENO` section has removed before mruby2.0

PR #2321, opened 2014-05-25
Print ISO methods test coverage in test result.

take-cheeze:
https://gist.github.com/take-cheeze/5d616bfd6561922964df is the verbose result.

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -44,6 +44,7 @@ end
 #       assertion
 def assert(str = 'Assertion failed', iso = '')
   t_print(str, (iso != '' ? " [#{iso}]" : ''), ' : ') if $mrbtest_verbose
+  mark_iso_done iso if $iso_stats
   begin
     $mrbtest_assert = []
     $mrbtest_assert_idx = 0
@@ -228,6 +229,8 @@ def report()
   if Object.const_defined?(:Time)
     t_print(" Time: #{Time.now - $test_start} seconds\n")
   end
+
+  report_iso if $iso_stats
 end
 
 ##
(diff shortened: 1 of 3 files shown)


matz:
Thank you for this PR — looking at it now, 11 years after filing.

Closing it without merge. Master has converged on a lighter-weight approach to ISO tracking that doesn't require the maintenance burden this PR introduces:

1. **Per-assert ISO codes are already in use throughout `test/t/`.** The `assert(str, iso)` form (defined in `test/assert.rb`) lets each individual test declare which ISO reference it covers — e.g., `assert('NilClass', '15.2.4.1')`, `assert('Literals Numerical', '8.7.6.2')`. So per-test traceability already exists; what your PR adds is a top-down coverage report.

2. **The `test/iso.rb` table (~350 lines) was acknowledged as incomplete at filing** ("based on draft from IPA, seems like it's incomplete"). To make the coverage report meaningful would require continuously maintaining the table as the ISO spec evolves and as mruby's API changes — a real ongoing cost.

3. **mruby's positioning has settled as "ISO-aware but not ISO-strict."** ISO compliance is a useful informational signal but isn't a primary project goal, so a top-down coverage report is more visibility than the project actively needs. The per-test annotations are sufficient for the use cases that have actually emerged.

The work itself is interesting — a comprehensive ISO method list is a non-trivial artifact to assemble. If a future maintainer wants to revive the coverage-reporting direction, the per-assert ISO codes already in master give them most of the inputs; the missing piece is just an authoritative method list, which would need to be sourced fresh in any case (the 2014 IPA draft is now well outdated).

Closing in favor of "per-assert annotations are enough." Thanks again for the original work.

PR #2325, opened 2014-05-27
use  build/${RUBY_PLATFORM}  instead of  build/host .

windwiny:
use  build/${RUBY_PLATFORM}  instead of  build/host .

support difference platform compiler in same source directory


zzak:
I'm having doubts about changing this from "host".

Even though it may make sense if your expecting platform to change, having a static "host" build for the platform allows us to make certain assumptions when packaging applications.

/cc @hone

PR #2331, opened 2014-05-29
Improve `Struct` test.

take-cheeze:
* Use `assert_*` method in `Struct` test.
* Improve `Struct#[]` and `Struct#[]=` tests. Checks raised exceptions and negative value index.
* Optimization using `mrb_check_intern_str()` seems to conflict with `NameError` raising when it is implemented as how it's defined in specification.


suzukaze:
Look good to me.

@matz  Please look into this PR.

PR #2382, opened 2014-06-10
Revert "Fallback to `mrb_assert` instead of dirty trick."

monaka:
This reverts commit 4de468e75414eb1adb3bafba19b72caba3a17044.

mrb_assert() may not compatible with mrb_static_assert().

You can use mrb_static_assert() (strictly _Static_assert())
in out of functions.
mrb_assert() (possibly assert()) can use only in functions.


monaka:
Closes as outdated.

PR #2384, opened 2014-06-10
Build sources in {GEM}/tools/{BIN_NAME}/ with the configuration for {GEM}.

monaka:
After applied this fix, build options defined gems are passed to sources in tools/.
This will make binaries more configurable.


zzak:
I'm also unsure about this change.

Do you have an example of how this can be configurable?

This change looks like it makes the build system even more complex, which we should actively try to avoid.

monaka:
Outdated. Close.

PR #2404, opened 2014-06-18
Count skipped tests and print them in test result report.

take-cheeze:


matz:
The issue was addressed by #4314

PR #2459, opened 2014-07-10
Core support for complex numbers

AE9RB:
This is support for "fast" complex numbers in the core. Enable MRB_COMPLEX to turn it on. Normally, MRB_COMPLEX will use its own mrb_value boxing but you can also use it with MRB_WORD_BOXING.

Methods like **, abs, conj, phase, arg, and polar should probably be an mrbgem. Let me know if you agree. I will do this work after I review any changes made to this pull request.

Also, vm.c can speed up complex math in the same way as regular math. I will submit this as a separate pull request so you can decide if it should be in the core.

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -17,10 +17,13 @@
 /* add -DMRB_INT64 to use 64bit integer for mrb_int; conflict with MRB_INT16 */
 //#define MRB_INT64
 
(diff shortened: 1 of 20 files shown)


matz:
The complex and rational numbers are introduce by d67d2ae8e88b93536e71dfa41a90721ce351da7c

Issue #2132, opened 2014-04-26
Add Class.name


postmodern:
`Class.name` comes in handy when outputting exceptions.

matz:
solved by #2470

jmhmccr:
Suggest adding this to 1.3 notes: #3140

PR #2470, opened 2014-07-13
Implement Module#name.

take-cheeze:
Solves #2132.


dabroz:
Can we add this to mruby?

matz:
merged.

PR #2559, opened 2014-08-25
Use ptrdiff_t instead of int.

cubicdaiya:


matz:
f388b6d61 addressed the issue.

Issue #2557, opened 2014-08-23
+ and - overrides lose sign of small Fixnums


chasonr:
If you override + or - in any class, and the right hand side is a Fixnum from -127 to -1, the sign of the Fixnum is lost.

Example:

```
class Optest
    def +(x)
        puts "add #{x}"
        0
    end
    def -(x)
        puts "sub #{x}"
        0
    end
end

q = Optest.new
q + +5
q + -5
q - +5
q - -5
q + +500
q + -500
q - +500
q - -500
```

mruby produces this output:

```
add 5
add 5
sub 5
sub 5
add 500
add -500
sub 500
sub -500
```

Ruby 2.1 works as expected:

```
add 5
add -5
sub 5
sub -5
add 500
add -500
sub 500
sub -500
```

chasonr:
I'm going to close this issue. The bug has disappeared in 2.0.0.

PR #2579, opened 2014-08-31
Support override of + - * / for Fixnum and Float.

chasonr:
For support of mruby-bignum (https://github.com/chasonr/mruby-bignum/).

A set of flags is added to mrb_state, corresponding to each of eight numeric
methods with existing optimizations:  + - \* / for Fixnum and for Float.
If the flag is set, the method has not been overriden and the optimization
remains in effect.  If clear, the use of the operator produces a method call.
Thus these methods can be overriden, just like any other.

This pull request supersedes #2558, and addresses issue #2557.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -184,6 +184,8 @@ typedef struct mrb_state {
   mrb_atexit_func *atexit_stack;
 #endif
   mrb_int atexit_stack_len;
+
+  unsigned numeric_methods;
 } mrb_state;
 
 #if __STDC_VERSION__ >= 201112L
--- a/src/class.c
+++ b/src/class.c
@@ -15,6 +15,7 @@
 #include "mruby/variable.h"
 #include "mruby/error.h"
 #include "mruby/data.h"
+#include "methods.h"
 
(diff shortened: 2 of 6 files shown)


sgnr:
I was wondering if there are any plans to merge this?

chasonr:
As far as I can tell, no.  That's for Matz to decide.  And I'm a bit
irritated about this.  I have a bignum gem that works best with this change
and Matz has just let it sit since August.

matz:
Closing without merge — adding flag-checks to OP_ADD/SUB/MUL/DIV/ADDI/SUBI puts a fixed cost on the hottest opcodes for every mruby program, even though virtually no program redefines `Fixnum#+` etc. That goes against mruby's memory-over-performance design priority, especially for embedded targets where icache footprint matters.

Since 2014, mruby also gained a generalized method cache (`mc_clear_by_id` in `src/class.c`), which subsumes the per-operator flag mechanism this PR proposed.

The original sign-preservation issue (#2557) will be addressed separately against the current byte-stream opcode format. Thanks for the contribution.

PR #2603, opened 2014-10-03
Corrected include path for object.h in value.h

pbosetti:
Previously, #include "mruby/object.h" was breaking usability of the mruby library
within OS X frameworks. Since value.h is actually including a header that is
in its same directory, there should be no need for prepending 'mruby/' to it.

### DIFF
--- a/include/mruby/value.h
+++ b/include/mruby/value.h
@@ -97,7 +97,7 @@ enum mrb_vtype {
   MRB_TT_MAXDEFINE    /*  23 */
 };
 
-#include "mruby/object.h"
+#include "object.h"
 
 #if defined(MRB_NAN_BOXING)
 #include "boxing_nan.h"


zzak:
I am :-1: for changing this, I think the current organization works and makes sense for me.

If you found a compatibility problem, like with compiler include spec, that would be possible cause for change.

ebertolazzi:
I do not understands the problem.
If you have a file file.c that use an inclusion like

#include “A/B.h”

this means that file B.h is the directory A relative to
the file doing the inclusion. This is a C standard rule.
If the file B.h is the same directory of file file.c
this is an error unless you SPECIFY different search paths
using -I switch. The correct way to include the file is

#include “B.h”

If you WANT to maintain A/B.h the correct way to include the file
is to use

#include <A/B.h>

specifying that you are using a “system” installed file and use -I
for a correct compilation.

The fact that "the current organization works and makes sense for me”
is not a technical argument against a better way to write C code.


regards,

matz:
I think this resolved by 5c405de (#3032).

PR #2615, opened 2014-10-20
fix test so that it can be ran multiple times instead of just once

mrbrdo:
This is the only test in the whole suite that cannot be ran multiple times, because undef will fail the second time.
It is nice to be able to do this e.g. for benchmarking reasons.

### DIFF
--- a/test/t/kernel.rb
+++ b/test/t/kernel.rb
@@ -376,7 +376,7 @@ assert('Kernel#method_missing', '15.3.1.3.30') do
   end
 
   class NoInspectClass
-    undef inspect
+    undef inspect if instance_methods(true).include?(:inspect)
   end
   d = NoInspectClass.new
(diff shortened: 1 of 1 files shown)


zzak:
I'm not sure I understand the problem.

I am able to run `mrbtest` multiple times (without re-compiling).
Such as: `$ rake test && ./build/test/bin/mrbtest`

mrbrdo:
This a pretty old PR... I think what I meant was if you ran this test in a loop, it would only work properly the first time. So not running mrbtest multiple times, but running the test itself multiple times without restarting mruby.

matz:
Corresponded code has already removed.

mrbrdo:
Thanks, after 8 years I get to know that my commit won't make it into ruby after all 😁 Cheers!

Issue #2678, opened 2014-12-17
block_given? returns inconsistent value at internal block with nested call


yasuyuki:
yield2.rb

``` ruby
def a
  p block_given?
  1.times do
    p block_given?
  end
end

def b
  yield a
end

b {}
```

% ruby yield2.rb  
false
false
% mruby/bin/mruby yield2.rb
false
true

Issue #2665, opened 2014-12-10
yielding failed with nil proc


yasuyuki:
yield.rb

``` ruby
def a(*a, &b)
  b(*a, &b)
end

def b(*a)
  c(*a) do |*c|
    if block_given?
      yield(*c)
    else
      c
    end
  end
end

def c(*a)
  if block_given?
    yield(*a)
  else
    a
  end
end

a(2)
```

% ruby yield.rb # => nothing shown
% mruby yield.rb
trace:
    [4] yield.rb:8:in Object.b
    [3] yield.rb:17:in Object.c
    [2] yield.rb:6:in Object.b
    [1] yield.rb:2:in Object.a
    [0] yield.rb:23
yield.rb:8: undefined method 'call' for nil (NoMethodError)

PR #2703, opened 2015-01-16
block_given? always returns false when it calls in block

### LINKED ISSUE #2695: block_given? does not work with nested blocks
looks https://github.com/mruby/mruby/commit/44c29e88aeaf241915b3d10a44d4a3748d1a6328 misses some cases.

```Ruby
def f
  loop do
    loop do
      p block_given?
      break
    end
    break
  end
end
f
```

```
% ruby -v c.rb
ruby 2.1.2p95 (2014-05-08 revision 45877) [x86_64-openbsd5.5]
false

% mruby c.rb
true
```

suzukaze:
Fix #2695
I made sure #2665, #2678 are not destroyed.

block_given? always returns false when it calls in block.
block cannot recieve block.

Could you tell me a sample code that block_given? return true in block.

### DIFF
--- a/src/kernel.c
+++ b/src/kernel.c
@@ -185,15 +185,10 @@ mrb_f_block_given_p_m(mrb_state *mrb, mrb_value self)
     given_p = FALSE;
   }
   else {
-    /* block_given? called within block; check upper scope */
-    if (ci->proc->env && ci->proc->env->stack) {
-      mrb_value *sp = ci->proc->env->stack;
 
-      /* top-level does not have block slot (alway false) */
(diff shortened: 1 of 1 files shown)


cremno:
>Could you tell me a sample code that block_given? return true in block.

Something like this? (It doesn't return `true` anymore after applying your patch.)
```ruby
def foo; define_singleton_method(:bar) { block_given? }; end; foo {}; bar
```

suzukaze:
@cremno Thank you very much.
I close this PR.

Issue #2140, opened 2014-04-27
mrb_float_to_str does not work with DISABLE_STDIO


XanClic:
With DISABLE_STDIO defined, mruby will not include stdio.h (as expected and desired). mrb_float_to_str will however use sprintf() for the conversion which then fails during the build because of the implicit declaration.

cremno:
@matz: This issue can be closed. Since the merge of #2723 `mrb_float_to_str()` can be used regardless of `ENABLE_STDIO` / `DISABLE_STDIO`.

PR #1947, opened 2014-03-25
Use dtoa in mrb flo to str

nurse:
As #1906, current mruby's float to string logic doesn't output ideal strings.
This patch use dtoa like CRuby and other implementations for double2str logc.

You may worry about increasing the binary size of mruby.
~~I confirmed it is rather than decreased on my FreeBSD amd64 though source lines are increased by 4000 lines~~.

%  ll build/host/bin/mruby # BEFORE
-rwxr-xr-x  1 naruse  users  1952985  3月 26 02:38 build/host/bin/mruby*
%  ll build/host/bin/mruby # AFTER
-rwxr-xr-x  1 naruse  users  1938695  3月 26 02:40 build/host/bin/mruby*

%  ll build/host/lib/libmruby* #BEFORE
-rw-r--r--  1 naruse  users  4799112  3月 26 05:20 build/host/lib/libmruby.a
-rw-r--r--  1 naruse  users      267  3月 26 05:20 build/host/lib/libmruby.flags.mak
-rw-r--r--  1 naruse  users  3898684  3月 26 05:20 build/host/lib/libmruby_core.a
%  ll build/host/lib/libmruby* #AFTER
-rw-r--r--  1 naruse  users  4893466  3月 26 05:36 build/host/lib/libmruby.a
-rw-r--r--  1 naruse  users      267  3月 26 05:36 build/host/lib/libmruby.flags.mak
-rw-r--r--  1 naruse  users  3999150  3月 26 05:36 build/host/lib/libmruby_core.a

Issue #1906, opened 2014-03-21
mrb_flo_to_str() generates inexact string for float


ghost:
, even if the float value can be represented in binary digits.

OK:
- `(0.5).inspect` => `"0.5"`

NG:
- `(0.25).inspect` => `"0.24999999999999"`
- `(0.125).inspect` => `"0.12499999999999"`
- `(0.0625).inspect` => `"0.06250"`

mattn:
closable?

PR #2723, opened 2015-02-13
re-implement mrb_float_to_str()

cremno:
The new implementation is backwards incompatible, but I couldn't find any usage outside mruby and I also couldn't think of a different and good name.

All ISO C99 printf conversion specifiers for floating point numbers and an optional precision are supported.

It is largely based on code from the MIT licensed musl libc (http://www.musl-libc.org/) and its floating point printing is exact (unlike the current code behind `Float#to_s`).

--
Related issues: #2140, #1947 and #1906.

These points require further discussion:

- backwards incompatible (if this is really a problem, what's a good name?)
- increases the size by a few KB (~4 KB on x86 SSE2), but it still supports hexadecimal floating point numbers (can be made optional) and flags
- 3 of 4 format strings, especially the both used in `src/numeric.c`, are weird. I don't understand why they are `"%.7g"`/`"%.14g"` and not (at least)`"%.9g"`/`"%.17g"` for correct round-tripping. It should be okay to change them, shouldn't it?
- Does it compile with Visual Studio 2012 or 2010?

### DIFF
--- a/include/mruby/numeric.h
+++ b/include/mruby/numeric.h
@@ -17,6 +17,8 @@ extern "C" {
 
 MRB_API mrb_value mrb_flo_to_fixnum(mrb_state *mrb, mrb_value val);
 MRB_API mrb_value mrb_fixnum_to_str(mrb_state *mrb, mrb_value x, int base);
+/* ArgumentError if format string doesn't match /%(\.[0-9]+)?[aAeEfFgG]/ */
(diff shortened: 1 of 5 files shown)


mattn:
Good Job :+1:

mattn:
Currently, 5.0.to_s become '5'. Is this your expected?

matz:
Yes. In mruby float numbers and integers are interchangeable as much as possible (mostly because mruby does not have bignums).

PR #2734, opened 2015-02-26
Fix for MSVC build.

dyama:
Fix initialize a structure for MSVC build.


cremno:
That's legal C99 (and C++). Why not simply ignore or disable the warning?

matz:
MSVC does not support full C99 for years and MS has no intention to fix.
Since MSVC is an important compiler, I will merge initializer fix.
But I have no idea about 99e409d. What is this? Is this kind of merge hiccups?

cremno:
But it's supported by MSVC as an extension (just like other C99 features).

matz:
I didn't know that. How do we turned the extension on?

cremno:
They're enabled by default (like GCC) and also it's a [level 4 warning](https://msdn.microsoft.com/en-us/library/6b73z23c.aspx), but mruby uses [level 3 (`/W3`)](https://github.com/mruby/mruby/blob/414678d61af1d689f56a4d36230415df056905c9/tasks/toolchains/visualcpp.rake#L5). So if someone or mruby itself (in the future?) uses `/W4`, then this warning should just be disabled in my opinion.

matz:
@cremno Do  you mean replacing `/W3` by `/W4` works?

cremno:
@matz: No. mruby uses `'/W3` (by default), but this warning is enabled by `/W4` (since VS2003!). `/W4` enables much more warnings and some of them are as irrelevant as this one or in rare cases false positives.

@dyama: Did you use `/W4` or why did the compiler emit a warning?

PR #2757, opened 2015-04-05
JIT for mruby

furunkel:
Over the last few weeks I've been working on a JIT for mruby. Now that it can run simple scripts I'd like to hear a few comments and get some feedback.
*This code is nowhere near production ready and a only proof of concept, use at own risk*.

### Implementation
The JIT in this PR is semi-automatically generated from the C source code.
It does not do any optimizations at the moment, but it get's rid of the VM's fetch-decode main loop.

It is using Clang/LLVM to parse the VM code and turn it into LLVM IR. Then all symbols are stripped from the IR. Linking happens at runtime using a symbol table. Assembly level changes
are made by parsing the LLVM-generated (machine level) assembly and writing it back afterwards.
*Note*: There is no runtime dependency on LLVM or Clang, all this happens at compile time.

### Results
It can run `fib(39)`, and is reasonably faster that the interpreter.
Here the `time` outputs for my machine.

**Interpreter**: 48.77user 0.00system 0:48.74elapsed 100%CPU
**JIT**: 31.38user 0.00system 0:31.37elapsed 100%CPU

These numbers are to be taken with a grain of salt, since this example is dominated by method lookup which is unoptimized (in both versions), i.e. the JIT might do better in other micro benchmarks.

Consider the following contrived example:
```ruby
i,j,k,l,m = 0,0,0,0

while i < 100000000
  while j < 100000000
    while k < 100000000
      while l < 100000000
        while m < 100000000
          m += 1
        end
        l += 1
      end
      k += 1
    end
    j += 1
  end
  i += 1
end
```
**Interpreter**: 40.03user 0.00system 0:40.00elapsed 100%CPU
**JIT**: 10.77user 0.00system 0:10.76elapsed 100%CPU

Both `mirb` and `mruby` now have a `--jit` option, that enables the JIT.
Futher tests and benchmarks are pending.

### Restrictions
Currently only x86-64 with SystemV ABI is supported. Supporting other platforms, is, however, relatively easy, since LLVM does most of the work.

Also, this JIT cannot do any cross-VM-instruction optimizations. That is, every jitted VM instruction
has to reload the VM values from and to memory, even if the next instruction needs them in registers.
This is not as bad as it sounds, as most instructions are so large, that they fill the registers with other stuff anyway.

### Changes to existing code
There are two main changes. First, the code handling a specific VM instruction has been factored into separate functions. It makes the code a bit cleaner.

Second, `op_enter` no longer relies on `op_jmpif`, this gets rid of some indirection that was there before. Instead, the offsets are saved in a table in the specific irep and `op_enter` does the jump directly.

Also, string constants that are accessed from code that is run through the JIT have been given
names, so that they can be used with the symbol table.

The code is still in an *early phase* and is *likely to crash*. This pull request is meant to bring up
a discussion about a possible JIT.


matz:
Wow, very interesting challenge. Let me investigate.
It might be unable to be merged to the core for the portablity's sake, but the value of the challenge is still real.

furunkel:
Yeah. I'll investigate into that when I find some time. Resolving conflicts should not be a big deal.

furunkel:
Silly problem failing Travis...
I'm getting a linker error with the C++ ABI (when linking with `mrbc`), apparently a redefinition of `enum mrb_run_flags` which is compiled into both, `libmruby` and `mrbc`.
How does one avoid these kind of conflicts ?

furunkel:
@cremno: Thought about typedef too. Doesn't help. Strange...

cremno:
@furunkel: Change line 3041 to:
```c
      ctx.run_flags = (enum mrb_run_flags)(ctx.run_flags & ~MRB_RUN_JIT);
```
C++ is more strongly typed than C. Enumeration types aren't just another integer type. So if it isn't really an enumeration, but rather flags you probably should use some unsigned integer type instead.

furunkel:
@cremno: Done that already. But I'd like to commit a version that passes Travis. All compiling related issues are resolved, however, linking still fails. Nevermind, I do a commit.

furunkel:
Funny, the C++ ABI build fails on my machine. Whatever...

zzak:
By the way, there is another JIT for mruby:
https://github.com/miura1729/mruby

Also I think givin the platform restrictions, this should remain an optional feature. Also, there seem to be some JIT unspecific changes that might be nice.

And, I'm wondering if the codegen stuff has to be written in cpp?

furunkel:
@zzak, yeah, would be nice if at least the `mrb_context_run` splitup makes it upstream.
As for the codegen stuff, yes, since LLVM is written in C++. However, this tool needs to be run only after changes to `vm.c`. It generates C files, so you don't even need a C++ compiler to build this.

@miura1729's approach seems a bit different from mine. Seems like he's using a runtime assembler (xbyak) that actually needs a C++ compiler to build. His JIT might be able to produce more efficient code, while porting to other architectures seems to be harder (e.g. xbyak cannot handle ARM or anything but x86 and x86-64).

furunkel:
In case somebody is interested, the JIT is now relatively stable and (at least on my machine) passes
all tests. It now supports all three boxing models. Only supported platform is still Linux/x86, though.

This now contains an awful lot of things at once but splitting things up would be a lot of work.
It's here in case somebody want's to try this out.

This code also contains the improved method cache work.
Results for the JIT are okay, I guess. Average is only about 12-13% but, e.g. Mandelbrot went up 50%. Also consider that this JIT does not yet do any real optimizations.
The method cache might shave off another 10-12%.
Here is the graph (note, method caching disabled).

jit

# Profiling
Another cool thing is that with every operation living in its own function, operation overheads now
show up in profiler logs, which makes profiling much more pleasant.
This PR also contains a task to generate profile charts like this one (using oprofile):
prof

zzak:
It would be interesting to use the MRI benchmarks instead of vendoring our own separate tests.

This also means we could build some automation around benchmarking MRI vs. MRuby to compare results. In my own tests, I've found MRuby to be slower than MRI, so this would help convince me about JIT for MRuby.

furunkel:
@zzak, yes, according to my own benchmarks, MRI is considerably faster than mruby (at least most of the time).
The method cache (that's still lying around in one of my branches),  however, does help quite a bit.
I don't have time, though, to update it for master.

Also note that the approach taken here for the JIT can easily be modified to emit LLVM bitcode
instead of native machine code.  It does have LLVM as a runtime dependency then.
On the plus side, one can let LLVM optimize the bitcode on a per-proc (i.e. per-method) level.
Anyway, I think I'm pretty much done with this stuff.

Moon4u:
@furunkel But (as far as I know) LLVM bytecode can be compiled to native. Also if we have dependency on llvm then linux isn't going to be the only supported platform, right? 
I'm currently looking at your [jit](https://github.com/furunkel/mruby/tree/jit) branch (the method cache seems to be there). The performance is great. 
I'm not entirely sure that I understand the current generation process. What can be changed to generate llvm bytecode?

furunkel:
@Moon4u, as far as I remember there is a `llc` call somewhere that pre-translates the LLVM IR to machine assembly.
The assembly is then post-processed and translated to machine code, which is then stored in a header file.
During runtime, the machine code (compiled into the executable through that header)
is then copied into executable memory and run, in a nutshell.

Instead, it should be possible to store the LLVM IR (in form of bitcode) directly. During runtime one then needs to load that bitcode and pass it to the LLVM JIT.
There are a few pitfalls though, e.g. jumps and a few other things need special treatment.

Anyway, for a serious implementation, mruby needs to change.
Currently it's really hard to track what mruby's virtual instructions do.
They might do fancy stuff to the VM stack or call stack and so on.
So, IMHO, first the VM needs to be adapted, then one can think about a JIT.

AnthonySuper:
Any progress on this? It looks really cool.

furunkel:
@AnthonySuper, nope, pretty much fallen asleep, due to lack of interest. I better close this...

AnthonySuper:
Darn. This looked awesome. Oh well.

zzak:
@furunkel FWIW, there's interest but personally haven't had the time to invest in following up recently. That's my bad.

Just happy to know if I can reach you when the time comes.

furunkel:
@zzak, well this was only an experiment and, frankly, the result is that I don't consider this a good idea anymore. Mostly, because I think the mruby codebase has, unmindfully with respect to a JIT, grown too far already; so that every solution now seems like a terrible hack. These things need to be planned for from the beginning, IMO. Anyway, given that mruby is supposed to be small and embeddable, that's ok, I guess.

AnthonySuper:
@furunkel I think that there's definitely some room for a JIT version of Ruby. I know I'd like to use Ruby for scripting in applications as opposed to something like LuaJIT, mostly because I feel Ruby is a better language.

Adding JIT to Mruby might not be a great idea at this point (I'm not technically knowledgeable on the subject enough to judge myself, unlike you) but I think a JIT version of Ruby is a great idea in the general case.

furunkel:
@AnthonySuper, yeah I know, it's really frustrating.
But I think this holds for CRuby just as well (actually it's even worse there, as it has progressed even further). It's design (IMHO) makes it very difficult to integrate a JIT now, at least in a clean fashion. 

Here's what I think the Ruby community needs to do: adapt the new JRuby IR representation
across all implementations (CRuby, Opal mostly). Then write (and translate JRuby's existing) 
IR optimization passes in some sort of DSL so they can be
shared across implementations. CRuby could then, just like JRuby does with JVM bytecode,
translate further to machine code, i.e. add a JIT.

texrg:
is possible to change benchmark jit to this code?
http://taktentus.herokuapp.com/be

`#!/usr/bin/ruby
s=0
k=0

for j in 0..10001
 for i in 0..10002
   k=1-k
   s+=k
 end
 s-=1003
end

puts "#{s}"`

PR #2763, opened 2015-04-17
Integer#% speed up

miura1729:
Using bit operation.

### DIFF
--- a/src/numeric.c
+++ b/src/numeric.c
@@ -638,7 +638,14 @@ fix_mod(mrb_state *mrb, mrb_value x)
     if ((b=mrb_fixnum(y)) == 0) {
       return mrb_float_value(mrb, NAN);
     }
+#if ((-2 >> 31) == -1) && ((11 % -5) == 1) && ((-11 % 5) == -1) && ((-11 % -5) == -1)
+    mod = a % b;
+    if (mod) {
+      mod += ((a ^ b) >> 31) & b;
+    }
+#else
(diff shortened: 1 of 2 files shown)


miura1729:
Sorry . This patch can't execute on non 32bit CPU. Please do not merge it.

tsahara:
`#if` trick won't work on cross-compiled binaries in general.

mattn:
You can replace the `#if`like below.
```c
#if UINTPTR_MAX == 0xffffffff // 32bit
#elif UINTPTR_MAX == 0xffffffffffffffff // 64bit
#else
// WTF
#endif
```

miura1729:
I didn't think about cross-compile. I have no idea about cross-compile.
I think this patch needs checking specification of modulo for negative number. But most CPU doesn't need check in fact. Most CPUs have modulo instruction I assume.

beoran:
The modulus operator % is guaranteed to work and exist by virtue of the C standard on all architectures, including embedded 8 or 16 bits architectures or long word 128 bits or 256 bits architectures (which this PR all ignores). Most optimizing compilers can and do replace the modulus by a bit operation if possible. See: http://stackoverflow.com/questions/22446425/do-c-c-compilers-such-as-gcc-generally-optimize-modulo-by-a-constant-power-of

In other words, this pull request is a bad idea. "Premature optimization is the root of all evil" in programming. We can and should use C's % and let the compiler optimize it for us.

matz:
Close #2763

PR #2764, opened 2015-04-17
Simple Method Cache

furunkel:
This is an implementation of a simple method cache. The cache is inherently local. Every proc gets a small cache (default is 4 entries). 
At the same time, the cache is precise, in the sense that it is always up-to-date and doesn't require any complicated invalidation logic. Instead, upon definition (or undefinition) of a method we walk through all procs and their caches and update them (a bit like a garbage collector walks the heap).
Thus, the cost of defining and undefining a method increases, whilst lookup becomes faster.

Now some numbers: the `fib(39)` benchmark yields an improvement of about 6-8% on my machine.
Of course, the cache shines if there are deep inheritance chains or tight loops with lookups and is essentially a waste of memory for simple scripts.

If you consider merging this, I'd vote for not enabling by default.

### DIFF
--- a/benchmark/ao-render.rb
+++ b/benchmark/ao-render.rb
@@ -5,8 +5,8 @@
 # mruby version by Hideki Miura
 #
 
-IMAGE_WIDTH = 256
-IMAGE_HEIGHT = 256
+IMAGE_WIDTH = 128
+IMAGE_HEIGHT = 128
 NSUBSAMPLES = 2
 NAO_SAMPLES = 8
 
--- a/build_config.rb
+++ b/build_config.rb
@@ -8,7 +8,7 @@ MRuby::Build.new do |conf|
     toolchain :gcc
   end
 
-  enable_debug
+  #enable_debug
 
   # Use mrbgems
   # conf.gem 'examples/mrbgems/ruby_extension_example'
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -69,6 +69,10 @@ typedef struct {
   int argc;
   int acc;
   struct RClass *target_class;
+#ifdef MRB_ENABLE_JIT
+  uint32_t send_idx;
+  uint32_t rescue_idx;
+#endif
 } mrb_callinfo;
 
 enum mrb_fiber_state {
@@ -186,6 +190,8 @@ typedef struct mrb_state {
   mrb_atexit_func *atexit_stack;
 #endif
   mrb_int atexit_stack_len;
+
+  struct RProc *proc_list;
 } mrb_state;
 
 #if __STDC_VERSION__ >= 201112L
@@ -381,6 +387,7 @@ MRB_API void mrb_print_error(mrb_state *mrb);
    + those E_* macros requires mrb_state* variable named mrb.
    + exception objects obtained from those macros are local to mrb
 */
+
 #define E_RUNTIME_ERROR             (mrb_class_get(mrb, "RuntimeError"))
 #define E_TYPE_ERROR                (mrb_class_get(mrb, "TypeError"))
 #define E_ARGUMENT_ERROR            (mrb_class_get(mrb, "ArgumentError"))
@@ -398,6 +405,10 @@ MRB_API void mrb_print_error(mrb_state *mrb);
 #define E_FLOATDOMAIN_ERROR         (mrb_class_get(mrb, "FloatDomainError"))
 
 #define E_KEY_ERROR                 (mrb_class_get(mrb, "KeyError"))
+
+#define E_FIBER_ERROR (mrb_class_get(mrb, "FiberError"))
+
+
 
 MRB_API mrb_value mrb_yield(mrb_state *mrb, mrb_value b, mrb_value arg);
 MRB_API mrb_value mrb_yield_argv(mrb_state *mrb, mrb_value b, mrb_int argc, const mrb_value *argv);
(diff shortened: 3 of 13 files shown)


furunkel:
@matz, any thoughts on this ?

matz:
This is nice idea. Two concerns.

1. for smaller devices, we need to be able to turn off method cache.
2. we need to invalidate cache, when methods are redefined.

We'll investigate.

furunkel:
1. That's already possible. See the `MRB_ENABLE_METHOD_CACHE` definition.
2. Invalidation should already work properly. See `mcache_update`.

benchmark/ao-render.rb
```diff
@@ -5,8 +5,8 @@
 # mruby version by Hideki Miura
 #
 
-IMAGE_WIDTH = 256
-IMAGE_HEIGHT = 256
+IMAGE_WIDTH = 128
+IMAGE_HEIGHT = 128
```
takahashim: Why do you change IMAGE_WIDTH and IMAGE_HEIGHT?  This PR conflict because of this change.

build_config.rb
```diff
@@ -8,7 +8,7 @@
     toolchain :gcc
   end
 
-  enable_debug
+  #enable_debug
```
takahashim: Why do you want to be disable debug?

furunkel: @takahashim: first, there is a separate debug build. Second, it is still compiled with -g.
`enable_debug ` sets `-O0`, which is a bad idea (for the default build) IMHO.

benchmark/ao-render.rb
furunkel: Ok, should not be in there...

build_config.rb
takahashim: OK, I have no opinion which is better, but disabling debug is independent from your method caching code, right?
If so, IMHO you should not include this change in your mcache PR, and make another PR to remove `enable_debug` in build_config.rb.

include/mruby.h
```diff
@@ -399,6 +404,10 @@ MRB_API void mrb_print_error(mrb_state *mrb);
 
 #define E_KEY_ERROR                 (mrb_class_get(mrb, "KeyError"))
 
+#define E_FIBER_ERROR (mrb_class_get(mrb, "FiberError"))
+
+
+
```
zzak: :scissors:

include/mruby/class.h
```diff
@@ -18,6 +18,21 @@ struct RClass {
...
+#define MRB_METHOD_CACHE_SIZE 4
+struct mrb_mcache {
+  struct mrb_mcache_entry entries[MRB_METHOD_CACHE_SIZE];
+  struct RClass *classes[MRB_METHOD_CACHE_SIZE];
+  struct RProc* procs[MRB_METHOD_CACHE_SIZE];
+  int16_t head;
+  int16_t tail;
+};
+
+
```
zzak: :scissors:

include/mruby/proc.h
```diff
@@ -66,6 +76,7 @@ MRB_API mrb_value mrb_proc_cfunc_env_get(mrb_state*, mrb_int);
 /* old name */
 #define mrb_cfunc_env_get(mrb, idx) mrb_proc_cfunc_env_get(mrb, idx)
 
+
```
zzak: :scissors:

src/class.c
```diff
@@ -8,14 +8,16 @@
...
-#include "mruby/class.h"
 #include "mruby/numeric.h"
 #include "mruby/proc.h"
+#include "mruby/class.h"
 #include "mruby/string.h"
 #include "mruby/variable.h"
 #include "mruby/error.h"
 #include "mruby/data.h"
 
+#include "class_inline.h"
```
zzak: Is this public? It should be `mruby/class_inline`

src/class_inline.h
```diff
@@ -0,0 +1,19 @@
+static inline struct RProc *
+mrb_method_search_vm_proc(mrb_state *mrb, struct RProc *p, struct RClass **cp, mrb_sym mid)
```
zzak: Why does this need to be in a separate file?

src/class.c
```diff
@@ -350,17 +458,7 @@ mrb_define_method(mrb_state *mrb, struct RClass *c, const char *name, mrb_func_t
...
-  struct RProc *p;
-
-  if (!h) h = c->mt = kh_init(mt, mrb);
-  k = kh_put(mt, mrb, h, name);
-  p = mrb_proc_ptr(body);
-  kh_value(h, k) = p;
-  if (p) {
-    mrb_field_write_barrier(mrb, (struct RBasic *)c, (struct RBasic *)p);
-  }
+  mrb_define_method_raw(mrb, c, name, mrb_proc_ptr(body));
```
zzak: Is this a necessary refactor?

furunkel: @zzak: not a necessary one if you like code duplication :)

src/class_inline.h
furunkel: @zzak: So it can be inlined in `vm.c`

src/class.c
furunkel: Nah, this one lives in `src`.

build_config.rb
furunkel: @takahashim: Yes, you are right, shouldn't be there either.

src/class.c
zzak: I mean, is it necessary for this feature?

src/class_inline.h
zzak: It just seems like an anti-pattern for mruby's current code organization, since we usually put headers in `include` if they're meant to be shared.

src/class.c
cremno: It even might make sense to get rid of `mrb_define_method_vm()` in favor of `mrb_define_method_raw()` (+ occasional `mrb_proc_ptr()`). But let's do that with a separate PR (assuming it does).

furunkel:
Hmm, I'm not really content with performance yet. Turned out that a constant cache size is worse than I thought. Unfortunately, currently, `RProc`s must have a constant size.
It might be possible to store the cache directly in the bytecode, just like an ordinary inline cache.

mcache

furunkel:
I have now reimplemented the cache with a different strategy. I introduced a new `OP_MCACHE` operation, which by itself is a no-op, that, however, stores the index to a cache entry. The cache size
now adapts to the number of (potential) send sides in an `irep`.
The code (https://github.com/furunkel/mruby/tree/mcache2) is currently not compatible with master, though.
Results are better now. Average speedup is 12.8%. For some benchmarks it's even 20%-30%.
mcache2

matz:
OP_MCACHE looks promising.  Looking forward to pull-request.

PR #2772, opened 2015-04-24
Add rake task for benchmarking

furunkel:
This pull request add a rake task to plot benchmarks in `/benchmark`.
The few existing benchmarks in mruby have been renamed and the problem sizes adjusted so
that they roughly take the same time to run.

## Plots

The output looks like this (these include benchmarks from MRI, which this PR does not include (are there possible licensing issues ?):
boxing

cc

## Usage
Use (or create a new) build config file in the `/benchmark` directory with the desired toolchain and options.
Then, to run the benchmarks and plot the results (requires `gnuplot`) use e.g.
`MRUBY_CONFIG=benchmark/build_config_boxing.rb  rake benchmark`,
which will build the required binaries if necessary and output the plots in a 
corresponding subdirectory in `/benchmark`.

To modify the appearance of the plots you can modify `/benchmark/plot.gpl`

## Requirements
Apart from Gnuplot,  a GNU-compatible `time` command (i.e. with support for the `-f` option) needs to be available on your system. On OSX, for instance, you might want to install `gnu-time` via homebrew and set a symlink.


Asmod4n:
Doesn't seam to work with gnuplot 5
```set yrange [0:];set terminal pngcairo font 'Sans, 8' lw 1 size 1400,1024;set xtics rotate by -45;set style histogram errorbars gap 2 lw 1;set style fill solid border -1;set output 'mruby/benchmark/build.png';plot ;
                             ^
line 0: unknown or ambiguous terminal type; type just 'set terminal' for a list```

PR #2775, opened 2015-04-28
remove SIZE_ERROR macro

cremno:
It's only used once and in that case it isn't useful anyway, since the code requires `size_t` to be at least an `uint32_t` and the length of a filename is stored as `uint16_t`.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -22,10 +22,8 @@
 
 #if SIZE_MAX < UINT32_MAX
 # define SIZE_ERROR_MUL(x, y) ((x) > SIZE_MAX / (y))
-# define SIZE_ERROR(x) ((x) > SIZE_MAX)
 #else
 # define SIZE_ERROR_MUL(x, y) (0)
-# define SIZE_ERROR(x) (0)
 #endif
 
 #if UINT32_MAX > SIZE_MAX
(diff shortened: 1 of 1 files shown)


cremno:
Merging this should be fine, but maybe hold on as there is something fishy. The current `src/load.c` contains these lines:

```c
#if SIZE_MAX < UINT32_MAX
# define SIZE_ERROR_MUL(x, y) ((x) > SIZE_MAX / (y))
# define SIZE_ERROR(x) ((x) > SIZE_MAX)
#else
# define SIZE_ERROR_MUL(x, y) (0)
# define SIZE_ERROR(x) (0)
#endif

#if UINT32_MAX > SIZE_MAX
# error This code cannot be built on your environment.
#endif
```

These two `#if` directives together don't make much sense, do they? `SIZE_ERROR_MUL` is definitely needed to prevent integer overflows, yet it's only defined at the same time as the `#error` directive causes the compilation to fail. The second check (`UINT32_MAX > SIZE_MAX`) is correct, isn't it? Then this code could be simplified to (or something similar):

```c
#define SIZE_ERROR_MUL(x, y) ...

#if UINT32_MAX > SIZE_MAX
# error This code cannot be built on your environment.
#endif
```

PR #2784, opened 2015-05-07
fix splat without assignment

### LINKED ISSUE #2781: unknown lhs 86
I have the following Ruby code:

```ruby
def complete(key, icase = false, pat = nil)
  candidates = candidate(key, icase, pat, &method(:each)).sort_by {|k, v, kn| kn.size}
  if candidates.size == 1
    canon, sw, * = candidates[0]
  end
end
```

It seems this line won't compile in mruby, and generates a `unknown lhs 86` in `gem_init.c` when compiling this code in an mrbgem.

`canon, sw, * = candidates[0]`

However, if I change it to this it works:

`canon, sw, _ = candidates[0]`

Is this a bug in mruby parser?

cremno:
Fixes #2781.

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -964,6 +964,9 @@ gen_assignment(codegen_scope *s, node *tree, int sp, int val)
       genop_peep(s, MKOP_AB(OP_MOVE, cursp(), sp), val);
     }
     break;
+  /* splat without assignment */
+  case NODE_NIL:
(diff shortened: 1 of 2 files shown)


zzak:
Sweet, thanks! <3

tsahara:
It works well on OpenBSD/MacOS.  Great!

PR #2792, opened 2015-05-15
let the C compiler validate the identifier

cremno:
Some C compilers may allow other characters in identifiers such as `$`. They may also implement C99's extended identifiers (`\u30EB\u30D3\u30FC`, `ルビー`).

https://gcc.gnu.org/onlinedocs/gcc/Dollar-Signs.html
https://gcc.gnu.org/wiki/FAQ#What_is_the_status_of_adding_the_UTF-8_support_for_identifier_names_in_GCC.3F
http://llvm.org/releases/3.3/tools/clang/docs/ReleaseNotes.html#extended-identifiers-unicode-support-and-universal-character-names

### DIFF
--- a/src/dump.c
+++ b/src/dump.c
@@ -1020,22 +1020,6 @@ mrb_dump_irep_binary(mrb_state *mrb, mrb_irep *irep, uint8_t flags, FILE* fp)
 }
 
 static mrb_bool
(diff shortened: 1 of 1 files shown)


cremno:
By the way, luajit and gperf also don't validate.

PR #2797, opened 2015-05-20
String#split fails to split on second or later separator

tsahara:
String#split fails to split some strings on their second or later separators.

```ruby
a = "a.b.c.d"; p a
b = a.gsub('.', '-'); p b
c = b.split('-'); p c
```

```
% mruby a.rb
"a.b.c.d"
"a-b-c-d"
["a", "b-c-d"]

% ruby a.rb
"a.b.c.d"
"a-b-c-d"
["a", "b", "c", "d"]
```

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -1768,22 +1768,21 @@ mrb_str_split_m(mrb_state *mrb, mrb_value str)
   result = mrb_ary_new(mrb);
   beg = 0;
   if (split_type == awk) {
-    char *ptr = RSTRING_PTR(str);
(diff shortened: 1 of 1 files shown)


schmurfy:
was this ever fixed ?
I just got this issue on 1.1.0, sometimes it split correctly and sometimes it split the first and stop there, if I call split twice it works...

matz:
We will release 1.2.0 in November.

schmurfy:
nice, thanks :)

PR #2798, opened 2015-05-20
Fix build mruby for NetBSD

tsahara:
There are 2 compilation issues on NetBSD.

1. NetBSD 6.1.5 (latest release) does not have fexpl(3).
2. NetBSD does not have libreadline but readline(3) is in libedit.

The patch attached fixes them.

### DIFF
--- a/mrbgems/mruby-bin-mirb/mrbgem.rake
+++ b/mrbgems/mruby-bin-mirb/mrbgem.rake
@@ -12,7 +12,11 @@ MRuby::Gem::Specification.new('mruby-bin-mirb') do |spec|
         spec.linker.libraries << 'termcap'
       end
     end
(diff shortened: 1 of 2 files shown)


zzak:
Do we know that `RUBY_PLATFORM` is always available during build?

I'm worried we may need to test this against other platforms..

Also notice `MRUBY_BUILD_HOST_IS_CYGWIN` so perhaps we should do something similar for other supported platforms?

tsahara:
> Do we know that RUBY_PLATFORM is always available during build?

Yes.  It is used on the top of the Rakefile: https://github.com/mruby/mruby/blob/master/Rakefile#L5

zzak:
マジかよー

PR #2808, opened 2015-05-29
Added String#ord method as for standard Ruby

pbosetti:


cremno:
Why not add it to `mruby-string-ext`? `String#chr` is part of it too.

matz:
As @cremno pointed out, implement this in `mruby-string-ext` gem. If you want, I'd volunteer to implement UTF-8 version in `mruby-string-utf8`, but of course, you can do it by yourself.

Issue #2769, opened 2015-04-22
Segfault in new_scope


tongson:
Found through afl-fuzz. Thousands of crashes, going to report obvious syntax errors.

VERSION AFFECTED: mruby 1.1.0

TESTCASE:
```
Proc.new do
  do
end
def x==x
end
```

BACKTRACE:
```
#0  0x000000000043c245 in new_scope (p=0x6fbbe0, body=0x6fc468) at src/parse.y:277
#1  0x000000000043e83f in yyparse (p=0x6fbbe0) at src/parse.y:1172
#2  0x000000000044ba11 in mrb_parser_parse (p=0x6fbbe0, c=0x6be9a0) at src/parse.y:5377
#3  0x000000000044bfbf in mrb_parse_file (mrb=0x6a2010, f=0x6dd7f0, c=0x6be9a0) at src/parse.y:5523
#4  0x000000000044c45d in mrb_load_file_cxt (mrb=0x6a2010, f=0x6dd7f0, c=0x6be9a0) at src/parse.y:5608
#5  0x00000000004024b9 in main (argc=2, argv=0x7fffffffe468) at /root/mruby-1.1.0/mrbgems/mruby-bin-mruby/tools/mruby/mruby.c:225
```

cremno:
>Thousands of crashes

I guess they aren't all fixed by now, are they? Do you mind uploading the findings folder of afl to somewhere, so someone can investigate them? Plain Ruby code shouldn't cause any segmentation faults.

cremno:
@matz: What about similar code (`new_module()`, `new_def()`)?

```ruby
module A
  Proc.new do
    def x==x end
  end
end
```
```ruby
def a
  Proc.new do
    def x==x end
  end
end
```
```diff
diff --git a/src/parse.y b/src/parse.y
index 336b20a..0571fe0 100644
--- a/src/parse.y
+++ b/src/parse.y
@@ -270,12 +270,17 @@ local_add(parser_state *p, mrb_sym sym)
   }
 }
 
+static node*
+locals_node(parser_state *p)
+{
+  return p->locals ? p->locals->car : NULL;
+}
+
 /* (:scope (vars..) (prog...)) */
 static node*
 new_scope(parser_state *p, node *body)
 {
-  node *n = p->locals ? p->locals->car : NULL;
-  return cons((node*)NODE_SCOPE, cons(n, body));
+  return cons((node*)NODE_SCOPE, cons(locals_node(p), body));
 }
 
[...]

PR #2810, opened 2015-05-29
failing multiple assignment with rest tests

cremno:
They all fail at the moment. At least one of them may even cause a segfault.

### DIFF
--- a/test/t/syntax.rb
+++ b/test/t/syntax.rb
@@ -219,6 +219,21 @@ assert('Splat without assignment') do
   assert_equal 1, a
 end
 
+assert('multiple assignment (rest)') do
+  *a = 0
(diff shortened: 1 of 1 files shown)


cremno:
By the way, this is the cause of some crashes posted in #2769. The backtrace is misleading as this is a codegen issue:
~~~ruby
*a, _ = []
p a
~~~
~~~
    1 000 OP_ARRAY      R1      R3      127     ; R1:a    (OOPS!)
    1 001 OP_LOADSELF   R3
    1 002 OP_MOVE       R4      R1              ; R1:a
    1 003 OP_MOVE       R5      R2              ; R2:b
    1 004 OP_SEND       R3      :p      2
    1 005 OP_STOP
~~~

PR #2424, opened 2014-06-25
Create `Undocumented` part to each C API document file.

take-cheeze:
- Document `mrb_open_core`.
- Document regexp mrbgems in mgem-list.

PR #2818, opened 2015-06-01
Added more documentation for some of the functions found in mruby.h and value.h

Mav7:
I wrote more documentation for some of the functions found in mruby.h and value.h. I also wrote examples of how they can be used.


take-cheeze:
I've wrote some documents before: https://github.com/mruby/mruby/pull/2424
Having examples seems great idea.

Mav7:
Oh I didn't see sorry take-cheeze and thank you.

PR #2828, opened 2015-06-08
gsub/sub supports back references in substitutes.  fixes #2816.

### LINKED ISSUE #2816: String#sub incompatibility with CRuby
```sh
% cat e.rb
puts '\n'.sub("\\", "\\\\\\\\")

% ruby e.rb
\\n

% bin/mruby e.rb
\\\\n
```

Note: `bin/mruby` above is built without Regexp gems.

tsahara:
This implementation is compatible with CRuby's String#gsub/sub
except \1 ... \9 and \+.  They are useless without Regexp library.

### DIFF
--- a/mrblib/string.rb
+++ b/mrblib/string.rb
@@ -20,6 +20,30 @@ class String
     self
   end
 
+  # private method for gsub/sub
+  def __sub_replace(pre, m, post)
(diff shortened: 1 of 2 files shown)


matz:
thank you.

PR #2829, opened 2015-06-09
Detect if ncurses' backend is terminfo or termcap.

### LINKED ISSUE #2662: Using terminfo instead of termcap
Hi there, everyone!

Don't know how feasible this is. Not all Unix-likes have termcap available to them, but most have terminfo by default. CRuby doesn't seem to depend on termcap when compiling. Is this somehow doable?

hiroshiyui:
Detect if ncurses' backend is terminfo or termcap. fixes #2662

Borrowed from @mattn's code at:
https://github.com/mruby/mruby/issues/2662#issuecomment-65535705

Signed-off-by: Huei-Horng Yo <hiroshi@ghostsinthelab.org>

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,7 @@
 # basic build file for mruby
 MRUBY_ROOT = File.dirname(File.expand_path(__FILE__))
 MRUBY_BUILD_HOST_IS_CYGWIN = RUBY_PLATFORM.include?('cygwin')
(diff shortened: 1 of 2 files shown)


take-cheeze:
(Just to make sure.)
I think there isn't `has_library` method for linker class.
And @mattn says

> Hmm, we should add search_library_path. Or has_libraray

In #2662.

mattn:
has_library is pseudo name for explain.

hiroshiyui:
My fault, I forget to implement `has_library` method.

PR #2831, opened 2015-06-10
Update TODO with recent approved features

zzak:


zzak:
> On 6/12/15, at 12:12 PM, matz/まつもとゆきひろ wrote:
> I think we should update it

applied.

zzak:
@matz ping, wdyt of this?

PR #2836, opened 2015-06-16
Added a check for 64 bit time_t overflow

kext:
This should check for overflow on platforms where time_t is 64 bit.

### DIFF
--- a/mrbgems/mruby-time/src/time.c
+++ b/mrbgems/mruby-time/src/time.c
@@ -193,7 +193,6 @@ mrb_time_wrap(mrb_state *mrb, struct RClass *tc, struct mrb_time *tm)
   return mrb_obj_value(Data_Wrap_Struct(mrb, tc, &mrb_time_type, tm));
 }
 
-
 /* Allocates a mrb_time object and initializes it. */
 static struct mrb_time*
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-time/src/time.c
```diff
@@ -201,12 +200,15 @@ time_alloc(mrb_state *mrb, double sec, double usec, enum mrb_timezone timezone)
...
-  else if ((sec > 0 && tm->sec < 0) || (sec < 0 && (double)tm->sec > sec)) {
-  out_of_range:
+  tm->sec = (time_t)sec;
+  if (
+    /* Check for overflow of 32 bit time_t */
+    (sizeof(time_t) == 4 && (sec > (double)INT32_MAX || (double)INT32_MIN > sec)) ||
+    /* Check for overflow of 64 bit time_t */
+    (sizeof(time_t) == 8 && (sec > (double)INT64_MAX || (double)INT64_MIN > sec)) ||
+    /* Check for wrapping of tm->sec */
+    ((sec > 0 && tm->sec < 0) || (sec < 0 && (double)tm->sec > sec))
```
kext: I'm not sure, but this line might be a no-op, because signed overflow is undefined behavior.

PR #2845, opened 2015-06-21
Implement `mrb_protect`.

take-cheeze:
This pull request implements `mrb_protect`, `mrb_ensure`, `mrb_rescue`, `mrb_rescue_exceptions`.
(`mrb_rescue_exceptions` is implementation of `rb_rescue2`.)
The 4 APIs are placed to "mruby-error" mrbgem which is included to 'default' gembox for now.

The changes to `Build#compile_as_cxx` is for missing directory when generating C++ file in some cases.

### DIFF
--- a/include/mruby/error.h
+++ b/include/mruby/error.h
@@ -35,7 +35,8 @@ MRB_API mrb_value mrb_ensure(mrb_state *mrb, mrb_func_t body, mrb_value b_data,
 MRB_API mrb_value mrb_rescue(mrb_state *mrb, mrb_func_t body, mrb_value b_data,
                              mrb_func_t rescue, mrb_value r_data);
 MRB_API mrb_value mrb_rescue_exceptions(mrb_state *mrb, mrb_func_t body, mrb_value b_data,
-                                        mrb_func_t rescue, mrb_value r_data, ...);
+                                        mrb_func_t rescue, mrb_value r_data,
+                                        mrb_int len, struct RClass **classes);
 
 #if defined(__cplusplus)
 }  /* extern "C" { */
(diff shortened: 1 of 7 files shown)


mattn:
@take-cheeze ++

cremno:
Great! But I see one issue. Your test code even contains an example:

```c
return mrb_rescue_exceptions(mrb, protect_cb, b, protect_cb, r, E_TYPE_ERROR, NULL);
```

In C `NULL` is allowed to and in C++ it's required to expand to `0`, so one always has to write at least `(struct RClass*)0`. Why not change the function to take a pointer to an array and its length instead? Then it'd be type-safe and mruby can't and doesn't have to be compatible to CRuby anyway. However there's also a new downside: an extra variable or, if a C99 compiler is used, a compound literal is needed . There's also the `mrb_funcall()` compromise (length + varargs).

jbreeden:
Hurray! Regarding @cremno's comment, +1 for the array argument idea. It could still be NULL terminated, without sacrificing type safety (or go the length + array route... either way).

Thanks, @take-cheeze!

take-cheeze:
@cremno 
Array of exception classes seems nice.
The variadic API is confusing and I've changed the API name too.
I'm thinking of following API:
```C
MRB_API mrb_value mrb_rescue_exceptions(
    mrb_state *mrb, mrb_func_t body, mrb_value b_data,
    mrb_func_t rescue, mrb_value r_data,
    mrb_int classes_len, struct RClasss **classes);
```

take-cheeze:
Done changing to array of classes API. ( https://github.com/mruby/mruby/commit/15c869f65222562a2a85aeac4e923131c271231a )

Asmod4n:
yay

matz:
merged

PR #2846, opened 2015-06-21
Add :mgem gem loading.

take-cheeze:
This pull request adds `:mgem` gem loader to load mrbgem from mgem-list.
And it will support auto mrbgem dependencies solving from mgem-list and core mrbgems too.


bovi:
wow :+1:

PR #2848, opened 2015-06-22
fix arity of lambdas with optional arguments

cremno:
From [the CRuby 2.2.2 Proc#arity documentation](https://github.com/ruby/ruby/blob/v2_2_2/proc.c#L790-L793):

>If the block has optional arguments, returns -n-1, where n is the number of mandatory arguments, with the exception for blocks that are not lambdas and have only a finite number of optional arguments; in this latter case, returns n.


cremno:
I don't know how to fix `Proc#curry`.

Issue #2847, opened 2015-06-22
mrb_string_value_cstr crashes if not given a string


jbreeden:
`mrb_string_value_cstr` does horrible things if given an argument that isn't actually a string (well, it tends to crash, anyway). That's all fine and well, since in C you should know what you're doing with your pointers. However, `mrb_string_value_ptr` avoids this by calling `mrb_str_to_str` on its argument. I'm just wondering if there's a reason that they behave differently. (Until now, I thought the _cstr version just stopped at the first null character, and _ptr was more suited to arbitrary char buffers. It wasn't clear to me that one is more type sensitive.)

I didn't see any issues/discussions on this yet. Maybe I'm the only one that's confused... O_o

cremno:
You should create a PR since I agree that `mrb_str_to_str()` has to be called. I think most mruby users call `mrb_str_to_cstr()` instead which is very strict (only `MRB_TT_STRING`) and (or but?) creates a copy. Also this function is only used once in mruby and there it's preceded by a check. That's probably why nobody did find this bug before.

However there's something else: `mrb_str_to_str()` does much more than `rb_str_to_str()`. Is this intended, @matz?

https://github.com/mruby/mruby/blob/e344c6ab6d7b3d6a8ffcec6f7f96aeba6c5fdeda/src/string.c#L628-L641
https://github.com/ruby/ruby/blob/a02a3f46496210ca401fd74585993c8754cbe91c/string.c#L1087-L1091

jbreeden:
Sounds good. I'm heading to work, I'll make the PR when I get back.

PR #2852, opened 2015-06-23
Mrb string value cstr patch

jbreeden:
All existing tests passing. Not sure how to write a test for the failure case (suggestions welcome), since the failure behavior is to crash. Also I only see tests in Ruby, and this function isn't exposed via any bindings. In fact...

I only see `mrb_string_value_cstr` used internally in 
 - `mrb_str_to_inum` : with a flag that controls whether it should be converted to a string, which this function now does by default.
 - `mrb_get_args` : in case of the 'z' format specifier. I'm guessing we want this conversion. Otherwise we need a way to not crash, and alert the user of the type mismatch.

### DIFF
--- a/mrbgems/mruby-sprintf/src/sprintf.c
+++ b/mrbgems/mruby-sprintf/src/sprintf.c
@@ -808,7 +808,7 @@ retry:
             if (mrb_fixnum_p(val)) goto bin_retry;
             break;
           case MRB_TT_STRING:
-            val = mrb_str_to_inum(mrb, val, 0, FALSE);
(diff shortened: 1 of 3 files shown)


jbreeden:
I went overboard with 9553e20d34c204826fde86620d2f433b3d59f1c1 (the call wasn't really redundant, it would still affect the else case that followed) and 264a059aef01b079d9788c8569182482edeab82a, so I reverted those changes. Even though the tests still passed, I'm not sure I understood the intent of the code at first. Going to look more at mrb_str_to_inum... something seems fishy. But the remaining one-liner fixes #2847 and keeps the tests happy.

PR #2854, opened 2015-06-23
Move deprecated macros and functions to dedicated header file

furunkel:
This PR introduces code to properly handle deprecation. For now, it moves the deprecated `ARGS_` macros into a separate file. Unfortunately, it is not possible to deprecate macros via a compiler directive.

Are there currently any other deprecated functions or structures ?

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -192,10 +192,13 @@ typedef struct mrb_state {
 # define mrb_noreturn _Noreturn
 #elif defined __GNUC__ && !defined __STRICT_ANSI__
 # define mrb_noreturn __attribute__((noreturn))
+# define mrb_deprecated __attribute__((deprecated))
 #elif defined _MSC_VER
 # define mrb_noreturn __declspec(noreturn)
+# define mrb_deprecated __declspec(deprecated)
 #else
 # define mrb_noreturn
+# define mrb_deprecated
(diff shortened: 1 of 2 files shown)


cremno:
>Unfortunately, it is not possible to deprecate macros via a compiler directive.

`#pragma GCC poison` (GCC) and `#pragma deprecated` (MSVC) but like the deprecated attributes they are standard C.

>Are there currently any other deprecated functions or structures ?

https://github.com/mruby/mruby/wiki/List-of-deprecated-and-obsolete-C-API

Moving them all into a new header doesn't make sense if their replacements are declared in other header files. They are all just macros anyway.

furunkel:
@cremno Don't the `#pragma`s work on identifiers too ?, also `#pragma poison` seems to
throw errors, not warnings and has special treatment for macros ?
I don't see a problem with not being standard C; GNUC compatible compilers and MSVC probably
cover 99% of what users have.

So what do you suggest, leaving this altogether ? Or replacing macros with proper stub functions and attribute-deprecating those ?
It bothers me a bit that no warnings are given for deprecated functions.

matz:
Adding `mrb_deprecated` is OK. And I think it's OK now to remove old `ARG_*` macros.  We don't need to preserve them.

PR #2859, opened 2015-06-24
add_dependency doesn't work

mattn:
```
rake aborted!
undefined local variable or method `root' for #<MRuby::Gem::List:0x002b4f6123c3b0>
```


hone:
I ran into this. I had to do `spec.add_dependency 'foo', :core => 'foo'`, basically pass in the 2nd param.

PR #2863, opened 2015-06-27
Fix typo; Replace extensional with extended

suzukaze:
Adjective of `extension`(noun) is not `extensional` but `extended`

extenstional means 外延的 in Japanese.

refer:
http://ejje.weblio.jp/content/extensional


zzak:
How about "Use <Name> extension" instead?

suzukaze:
@zzak 
Thanks you for advice.
I will change them.

PR #2865, opened 2015-06-28
Add Float#angle

suzukaze:


matz:
I am not sure why we should add `#angle` here, especially without complex number support.
Any use-case?

suzukaze:
@matz 
No particular reason.

PR #2868, opened 2015-06-29
Avoid a narrowing cast in flo_round under MRB_INT64.

sgnr:
`flo_round` is producing the following warning under both Darwin and Linux x86_64

```
mruby/src/numeric.c:454:7: warning: absolute value function 'abs' given an argument of type 'mrb_int' (aka 'long long') but has parameter of type 'int' which may
      cause truncation of value [-Wabsolute-value]
  i = abs(ndigits);
```

An alternative would be to not rely on `MRB_INT64` and have a check such as
```c
if (sizeof(num) == sizeof(long long)) {
  return llabs(num);
} else {
  return abs(num);
}
```

### DIFF
--- a/src/numeric.c
+++ b/src/numeric.c
@@ -436,7 +436,7 @@ flo_round(mrb_state *mrb, mrb_value num)
 {
   double number, f;
   mrb_int ndigits = 0;
-  int i;
+  mrb_int i;
 
   mrb_get_args(mrb, "|i", &ndigits);
   number = mrb_float(num);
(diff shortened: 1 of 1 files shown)


cremno:
What do you think about using sth. like `i = ndigits > 0 ? ndigits : -ndigits;` instead of calling either `abs()`or `llabs()`? It has the same issue as the functions (undefined behavior when `ndigits == MRB_INT_MIN`) but it's simpler since no `#ifdef` and `llabs()` is likely not defined by older Visual Studio versions.

sgnr:
Good idea. I just updated the PR.

PR #2879, opened 2015-07-06
call always frexp() instead of frexpl()

cremno:
That `#if` was getting a bit ridiculous. There are no plans for a `long double` version of `mrb_float` anyway or are there?


cremno:
Rebased because of #2889.

matz:
I will merge this to see if it works well on all platforms (I assume it works).
Report if anyone face the problem.

PR #2880, opened 2015-07-08
implement llround. some platform (ex: mingw32) doesn't have llround.

mattn:
if you want, I'll make block this part with `#ifdef _WIN32`.

### DIFF
--- a/mrbgems/mruby-time/src/time.c
+++ b/mrbgems/mruby-time/src/time.c
@@ -11,6 +11,10 @@
 #include "mruby/class.h"
 #include "mruby/data.h"
 
+#if !defined(__MINGW64__) && defined(_WIN32)
+# define llround(x) round(x)
+#endif
+
 #if defined(__MINGW64__) || defined(__MINGW32__)
 # include <sys/time.h>
 #endif


matz:
travis-ci test failed due to lack of `LONG_LONG_MAX` in `mrbgems/mruby-time/src/time.c`

tsahara:
+1 for `#ifdef _WIN32` to reduce binary size.

cremno:
But MinGW-w64 (which also targets 32-bit Windows) and VS 2013/2015 have `llround()`.

mattn:
@cremno llround is macro which is possible to check with `#if defined(llround)`?

mattn:
rebased. this should work.

cremno:
@mattn: Does it have `round()`?

mattn:
@cremno yes, but maybe not support long long.

cremno:
Well, then let's use `round()`.

The body of a conforming definition of `llround()` can be as simple as `return (long long)round(x);` since [the range error is optional](http://port70.net/~nsz/c/c99/n1256.html#7.12.9.7p2).

mattn:
@cremno please take a look.

PR #2881, opened 2015-07-08
fix checking of Regexp

mattn:


zzak:
I get the following test failure with this patch:

```
NotImplementedError: Array#slice [15.2.12.5.29] => Regexp class not implementedNotImplementedError: Regexp class not implemented
```

PR #2885, opened 2015-07-14
Module#prepend

### LINKED ISSUE #2082: Module#prepend
How hard would it be to add Module#prepend? Is it possible to do it as a mruby-module-ext?

I was looking at the ruby MRI implementation of it, as I really need it, but I saw it also affected some other methods, like `class_instance_method_list`

archSeer:
Hi, we requested Module#prepend, so about a year later we figured what the heck and went ahead and implemented it.

We've gone and ported over the original ruby commits by ko1 and nobu over to mruby, all tests currently pass. RClass struct size is now bigger, because we needed to add `origin`, but that seems unavoidable.

/cc @matz @cremno @zzak

Closes #2082.

### DIFF
--- a/include/mruby/class.h
+++ b/include/mruby/class.h
@@ -16,6 +16,7 @@ struct RClass {
   struct iv_tbl *iv;
   struct kh_mt *mt;
   struct RClass *super;
+  struct RClass *origin;
 };
 
 #define mrb_class_ptr(v)    ((struct RClass*)(mrb_ptr(v)))
--- a/include/mruby/class.h
+++ b/include/mruby/class.h
@@ -49,8 +49,11 @@ mrb_class(mrb_state *mrb, mrb_value v)
   }
 }
 
-#define MRB_SET_INSTANCE_TT(c, tt) c->flags = ((c->flags & ~0xff) | (char)tt)
-#define MRB_INSTANCE_TT(c) (enum mrb_vtype)(c->flags & 0xff)
+// TODO: figure out where to put user flags
+#define MRB_FLAG_IS_ORIGIN (1 << 20)
+#define MRB_FLAG_IS_INSTANCE (0xFF)
+#define MRB_SET_INSTANCE_TT(c, tt) c->flags = ((c->flags & ~MRB_FLAG_IS_INSTANCE) | (char)tt)
+#define MRB_INSTANCE_TT(c) (enum mrb_vtype)(c->flags & MRB_FLAG_IS_INSTANCE)
 
 MRB_API struct RClass* mrb_define_class_id(mrb_state*, mrb_sym, struct RClass*);
 MRB_API struct RClass* mrb_define_module_id(mrb_state*, mrb_sym);
--- a/include/mruby/class.h
+++ b/include/mruby/class.h
@@ -51,9 +51,9 @@ mrb_class(mrb_state *mrb, mrb_value v)
 
 // TODO: figure out where to put user flags
 #define MRB_FLAG_IS_ORIGIN (1 << 20)
-#define MRB_FLAG_IS_INSTANCE (0xFF)
-#define MRB_SET_INSTANCE_TT(c, tt) c->flags = ((c->flags & ~MRB_FLAG_IS_INSTANCE) | (char)tt)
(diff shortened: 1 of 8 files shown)


include/mruby/class.h
```diff
@@ -48,8 +49,11 @@ mrb_class(mrb_state *mrb, mrb_value v)
   }
 }
 
-#define MRB_SET_INSTANCE_TT(c, tt) c->flags = ((c->flags & ~0xff) | (char)tt)
-#define MRB_INSTANCE_TT(c) (enum mrb_vtype)(c->flags & 0xff)
+// TODO: figure out where to put user flags
+#define MRB_FLAG_IS_ORIGIN (1 << 20)
+#define MRB_FLAG_IS_INSTANCE (0xFF)
```
take-cheeze: I think `MRB_FLAG_IS_INSTANCE` should be renamed to `MRB_INSTANCE_TT_MASK` since it's a mask for instance type tag.

IceDragon200: @take-cheeze Ah, right, okay, I've changed it, can you review again?

take-cheeze: @IceDragon200 LGTM

matz:
Thank you for the implementation. I will merge this. But I need at least a few days to review the patch, sinch the change is huge.

archSeer:
@matz Alright, thank you! For reference, here's our fork PR where we discussed the implementation commit-by-commit, if it's of any help: https://github.com/polyfox/mruby/pull/1

src/class.c
```diff
@@ -767,47 +771,133 @@ boot_defclass(mrb_state *mrb, struct RClass *super)
...
+mrb_include_module(mrb_state *mrb, struct RClass *c, struct RClass *m)
+{
+  int changed = include_module_at(mrb, c, c->origin, m, 1);
+  if (changed < 0) {
+    mrb_raise(mrb, E_ARGUMENT_ERROR, "cyclic include detected");
+  }
+}
+
+MRB_API void
+mrb_prepend_module(mrb_state *mrb, struct RClass *c, struct RClass *m)
```
cremno: This function should be either `static` or declared in a header (`mruby.h` like `mrb_include_module()` or maybe even `mruby/class.h` because it won't be used that often).

src/class.c
```diff
@@ -767,47 +771,133 @@ boot_defclass(mrb_state *mrb, struct RClass *super)
...
+  if (m->tt == MRB_TT_ICLASS) {
+    ic->c = m->c;
+  } else {
+    ic->c = m;
+  }
+  return ic;
+}
+
+MRB_API int
+include_module_at(mrb_state *mrb, struct RClass *c, struct RClass *ins_pos, struct RClass *m, int search_super)
```
cremno: `static`?

src/class.c
IceDragon200: Ah, we missed that one, thanks!

archSeer:
Hi @matz, sorry for the bother, any updates on this?

matz:
Finally I took time to review this pull request. Sorry for being so late.

The PR seems fine except for on thing. You have added `origin` member to `struct RClass` but due to `union RVALUE` system in gc.c, adding a extra member to the struct would increase 1 word memory space for every mruby object. This is unacceptable for small footprint implementation.

If you are willing to address this issue, I'd love to wait. If you are not, I will take time to fix, but you may have to wait longer in this case.

archSeer:
Hi @matz, in the original pull request description we pointed out "RClass struct size is now bigger, because we needed to add origin, but that seems unavoidable."

I'd be happy to try and implement an alternative, however I have no idea how that would look like.

furunkel:
A golden opportunity to remove `RVALUE` and introduce something like a segregated free list :wink: .

matz:
@furunkel it's trade-off, RVALUE makes GC simpler and sweeping faster.

@archSeer I think I can succeeded to remove origin by adding one bit flag.  I will commit soon.

archSeer:
:tada: :balloon: :tada: thanks matz!

furunkel:
@matz, sure, but segregating into a *small* number of object types, say, `small`, `medium`, `large` wouldn't be very complicated and allow to simplify a few other things while not really hurting sweeping performance.

PR #2886, opened 2015-07-16
Add -fPIC flag

yoshida-mediba:


mattn:
AFAIK, `-fPIC` do effects to the location of symbols in the binary. though you want enable this but someone may not want. if you want enable this on linux or OSX, you should specify it just for them at least.

yoshida-mediba:
OK

PR #2899, opened 2015-07-27
Shallow clone mgem source

zzak:


matz:
What will happen with "--depth 1" option?  Performance?

franckverrot:
Cloning will be much faster as it fetches the history as a single revision (the only commit `git log` will show is the latest commit of the given branch). (We do that on [`mruby-redis`](https://github.com/matsumoto-r/mruby-redis/blob/master/Rakefile#L9) and [`holycorn`](https://github.com/franckverrot/holycorn/blob/master/Rakefile#L9) and this speeds up the builds).

zzak:
Yeah, it will only make the fetch for the mgem-list faster.

It looks like the only other `git.run_clone` already does this:
https://github.com/mruby/mruby/blob/a41a649e8370a8a0dfd7e5de4833be28842fdfb9/tasks/mruby_build_gem.rake#L96

PR #2907, opened 2015-08-02
be able to source with :path for mrbgems

hone:
Allow people in `build_config.rb` to override a github repo or mgem with a local working copy, so you don't need to push to github to see the changes. This would like this:

```ruby
conf.gem :path => '/path/to/gem_fork`
```


zzak:
How is this different than

```ruby
conf.gem '../path/to/my/gem'
```

joans321:
may be more conform to convention

PR #2914, opened 2015-08-12
Change to return self if array length is 1 and less in mrb_ary_reverse

gkta:
* No need to make new object if array length is 1 and less.


matz:
Array#reverse of CRuby copies the array if its length is 1. 
```
a=[1]
p a.object_id
b=a.reverse
p b.object_id
```
We don't want to introduce incompatibility here.

PR #2915, opened 2015-08-15
Remove MRB_DEBUG in favor of ENABLE_DEBUG to do both

zzak:
Open to suggestions to keep MRB_DEBUG for compatibility.


cremno:
Interesting but those are two different features: e.g., you wouldn't be able to build `mrdb` without enabled assertions. If you think their names are confusing, then let's fix them (especially since `ENABLE_DEBUG` isn't prefixed with `MRB_`).

zzak:
@cremno Hello!

I'm not sure why they need to be separate, in my opinion they both are useful for debugging.

The naming of the constants is irrelevant to me, but having the `ENABLE_DEBUG` constant match the `enable_debug` option in build config seems ok.

My questions is: why not both? hehe

cremno:
I agree that both can be useful for debugging, but they're used for different things. I think a more descriptive name for the current `ENABLE_DEBUG` would be `MRB_DEBUG_HOOKS`. This macro is (usually) defined to enable building a mruby-based debugger for Ruby code. However `MRB_DEBUG` is used to debug/to verify the correct behavior of (some parts of) mruby itself.

matz:
As @cremno pointed out `ENABLE_DEBUG` is to enable instruction hook for debugger (to debug Ruby programs), whereas `MRB_DEBUG` is to enable assert to debug mruby itself.  Renaming suggestions are welcome, but I don't think we are going to unify them.

zzak:
What does assert do exactly?

I would think `ENABLE_DEBUG` can be defined by `enable_debug` flag. Since this is most likely used by developers, who are not debugging mruby internals. In my opinion the `MRB_DEBUG` flag sounds like it would be used for debugging internals (e.g. mrb, the interpreter)

cremno:
>What does assert do exactly?

It is similar to rspec's `expect(actual).to be true` or minitest's `assert_equal true, actual`. The significant difference is that it is not in a different file and intermixed with the implementation code.

>I would think `ENABLE_DEBUG` can be defined by `enable_debug` flag.

Yeah, I probably would too. This is why I suggested renaming it. What about `MRB_VM_HOOKS`? This feature is also used by [mruby-profiler](https://github.com/miura1729/mruby-profiler).

PR #2926, opened 2015-08-31
Fix MRUBY_RELEASE_TEENY and MRUBY_RELEASE_NO, refactor version macros, and add a global const variable.

suzukaze:
* Fix MRUBY_RELEASE_TEENY and MRUBY_RELEASE_NO
mruby version is "1.1.0" but MRUBY_RELEASE_TEENY is 1 and MRUBY_RELEASE_NO is 10101.
It is correct that MRUBY_RELEASE_TEENY is 0 and MRUBY_RELEASE_NO is 10100.

* Refactor version.h macros
It is redundant MRUBY_VERSION ,MRUBY_RELEASE_MAJOR, MRUBY_RELEASE_MINOR and MRUBY_RELEASE_TEENY are set.
MRUBY_RELEASE_MAJOR, MRUBY_RELEASE_MINOR and MRUBY_RELEASE_TEENY should generate MRUBY_RELEASE_TEENY.
It is redundant MRUBY_RELEASE_DATE, MRUBY_RELEASE_YEAR,  MRUBY_RELEASE_MONTH and MRUBY_RELEASE_DAY are set.
MRUBY_RELEASE_YEAR,  MRUBY_RELEASE_MONTH and MRUBY_RELEASE_DAY should generate MRUBY_RELEASE_DAY.

* Add a global const variable MRUBY_RELEASE_NO

```
$ mirb

> MRUBY_VERSION
 => "1.1.0"
> MRUBY_RELEASE_NO
 => 10100
> MRUBY_RELEASE_DATE
 => "2014-11-19"
> MRUBY_DESCRIPTION
 => "mruby 1.1.0 (2014-11-19) "
> MRUBY_COPYRIGHT
 => "mruby - Copyright (c) 2010-2014 mruby developers"
```


kazuho:
IMO we should better not modify the version number now, since existing programs might already be checking if `MRUBY_RELEASE_NO` is `10101` to see if the mruby being used provides a feature that is provided by the 1.1.0 release.

Changing `MRUBY_RELEASE_NO` (and `MRUBY_RELEASE_TEENY`) would break such assumptions.

Instead, I think it would be better to fix the issue in the next bug-fix release (should there be any), by skipping 1.1.1 and going to 1.1.2 instead.

suzukaze:
I got it. Now we shouldn't change MRUBY_RELEASE_NO (and MRUBY_RELEASE_TEENY)

matz:
could you re-submit PR without modifying TEENY?

suzukaze:
I got it.

PR #2937, opened 2015-09-02
Distribute MRuby's source as a Ruby gem

sagmor:
I've been playing around with some tooling for MRuby and think that having MRuby's source as a gem could be a good Idea.

Right now, the simplest way to build MRuby is to clone the repo -> edit `build_config.rb` -> `./minirake`

The cool thing of bundling MRuby's source as a gem is that you can specify it as a Bundler dependency and use it with other tooling.

To make it clearer, here's a simple `mruby-build` command that can be used anywhere with any config file with a shared MRuby source: https://github.com/sagmor/mruby-build-command
This would allow you create a stand alone `build_config.rb` file and build it with a simple
```
mruby-build [FILE]
```

Another more complex example I'm experimenting with is a build tool that can crossbuild an application to multiple platforms: https://github.com/loicarb/loica-build
The idea behind this is to eventually build a whole app with MRuby and crossbuild it to multiple platforms like Android, iOS and the Web.

These are currently just experiments I'm working on but thought somebody might be interested and having a simple way of working with MRuby might foster more tooling around it.


joonjeong:
I think this is really cool feature.

sagmor:
I've done some more experiments around this tooling and is working pretty well.
In fact, I've improved upon the build command idea and built a [small command line tool](https://github.com/sagmor/mruby-cli) that can build and test configs and I'm even adding generators (sagmor/mruby-cli#1) to bootstrap projects quickly.

Let me know if you like the approach.

sagmor:
Awesome.

I can backport this to make '1.0.0' and '1.1.0' releases on rubygems (for completeness)

PR #2950, opened 2015-09-11
mirb should take filename and arguments.

mattn:
works as same as irb

```
$ cat test.rb
puts "foo"

$ mirb test.rb hello

foo
 => nil
> p ARGV
["hello"]
 => ["hello"]
```

### DIFF
--- a/mrbgems/mruby-bin-mirb/tools/mirb/mirb.c
+++ b/mrbgems/mruby-bin-mirb/tools/mirb/mirb.c
@@ -194,6 +194,7 @@ is_code_block_open(struct mrb_parser_state *parser)
 }
 
 struct _args {
+  FILE *rfp;
   mrb_bool verbose      : 1;
   int argc;
   char** argv;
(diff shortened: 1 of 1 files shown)


mattn:
oops, test failed

matz:
The behavior is different from irb (irb terminates after reading from script file). what is the use-case for this change?

zzak:
Should mirb terminate too?

matz:
I just don't know. I cannot imagine the use case.

mattn:
Ah, sorry. I seemed to have seen a vision.

matz:
may I close this issue, or should we make mirb consistent with irb?

mattn:
I already modified change to make behavior as same as irb.

matz:
the point is whether we have to make it compatible to irb or not. I feel the current irb behavior has no benefit.

strcmp:
I think it is more useful to not terminate at the end. At the moment `irb foo.rb` sounds `Kernel#p` and `ruby foo.rb` on steroids. it doesnt seem useful.

sagmor:
Given the purpose of irb it seems natural to pass a file name to "preload" an IRB session. You could be working on a new script and use that functionality to test things out as you go or to fast-forward a tedious setup (DB connection or setup some constants for example) I've came up with times hacking with some new/unknown API when I had to copy and paste multiple times some initial setup script that could have benefited from such functionality. Keeping the session open could help on those scenarios.

On the other hand I see no benefit of closing it on finish. Just use `ruby filename.rb` instead right?

pbosetti:
I second @sagmor comment: `irb filename.rb` can prove useful in preloading setups, which implies that it should not exit on finish.

PR #2953, opened 2015-09-16
support String#[pos, len]= val

takahashim:
cf. http://docs.ruby-lang.org/en/2.2.0/String.html#method-i-5B-5D-3D

### DIFF
--- a/mrblib/string.rb
+++ b/mrblib/string.rb
@@ -154,13 +154,48 @@ class String
   end
 
   ##
-  # Modify +self+ by replacing the content of +self+
-  # at the position +pos+ with +value+.
-  def []=(pos, value)
-    pos += self.length if pos < 0
-    b = self[0, pos]
-    a = self[pos + 1..-1]
-    self.replace([b, value, a].join(''))
+  # Modify +self+ by replacing the content of +self+.
+  # The portion of the string affected is determined using the same criteria as +String#[]+.
+  def []=(*args)
+    anum = args.size
+    if anum == 2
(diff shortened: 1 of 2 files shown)


mrblib/string.rb
```diff
@@ -154,13 +154,48 @@ def each_byte(&block)
...
+          a = self[(posnum + pos.length)..-1]
+          self.replace([b, value, a].join(''))
+          return value
+        else
+          raise IndexError, "string not matched"
+        end
+      else
+        pos += self.length if pos < 0
+        if pos < 0 || pos > self.length
+          raise IndexError, "index #{args[0]} out of string"
```
franckverrot: do you think we could we rephrase “... out of bounds” ?

takahashim: I'm not sure which is better, but it' from CRuby:

```ruby
$ ruby -e '"foo"[4,1]="X"'
-e:1:in `[]=': index 4 out of string (IndexError)
	from -e:1:in `<main>'
```

franckverrot: Yes, I figured :-)

On a somehow related note, I can't see where in the spec `String#[]` (15.2.10.5.6) is supposed to raise anything in those cases. :-/

takahashim: I guess `String#[]` and `String#[]=` are different methods, and the latter is not defined in ISO Ruby :worried:

franckverrot: Whoops... indeed :+1:

PR #2956, opened 2015-09-17
Problem: Hash#fetch doesn't raise KeyError when a key cannot be found

Asmod4n:
Solution: change the Exception class raised to KeyError when a key
cannot be found.


Asmod4n:
How can i get rid of that remote branch merge commit?

takahashim:
I guess it might be easy to re-make pullrequest from the begining.

```
$ git clone git@github.com:mruby/mruby.git
$ cd mruby
$ git checkout -b fix-hash-fetch

(... edit mruby-hash-ext/mrblib/hash.rb and mruby-hash-ext/test/hash.rb ...)

$ git commit
$ git remote add asmod4n git@github.com:Asmod4n/mruby.git
$ git push asmod4n

(when you see https://github.com/mruby/mruby, you found PR button, so push it to make PR)
```
Also you should read https://github.com/mruby/mruby/blob/master/CONTRIBUTING.md.
This file says `Create a branch which is dedicated to your change.`  You seems not to do it (your PR use `master` branch).  You can do it with `git checkout -b fix-hash-fetch`  as above.

mattn:
Do not mind please. @matz will merge this with cherry pick.

PR #2959, opened 2015-09-21
C API documentation

sagmor:
I started to work on documenting MRuby's C API.

While working on a gem project, I started to get locked quite often on "whats the function that does x?" moments so I started to document the methods I found on the process.

I noticed there are some methods already documented at [doc/api](https://github.com/mruby/mruby/tree/master/doc/api) but I believe it's better to have the API documented on the headers, so I've been taking those docs, and merging them into the headers as well.

I'm leaving this pull request to get some feedback before putting more effort into it.

Some of the things I've experimented with related with documenting MRuby's C API are:

* Generate a simple API website: the generated site is available right now at http://sagmor.com/mruby-c-api/ and its source at https://github.com/sagmor/mruby-c-api. eventually with a "c-api" repo it could be automated and live in `mruby.org/c-api` for example.
* Grabbed some macros like `MRB_API`, `mrb_deprecated` and moved them to `mruby/common.h`
* Add a couple macros `MRB_BEGIN_DECL` and `MRB_END_DECL` to get rid of those ugly `extern "C"` all over the headers.
* Replacing some `#define` aliases with inline function, this is just an experiment/proposal, they look more consistent while documenting (see [mrb_class_new_instance](http://sagmor.com/mruby-c-api/group__mruby.html#ga05117cc9d150df3921cdade39a74d59e))
* typedeffed [mrb_get_args](http://sagmor.com/mruby-c-api/group__mruby.html#ga6659ff35e634812fff48bb6a7f66730e) format string to document it separately (See [mrb_args_format](http://sagmor.com/mruby-c-api/group__mruby.html#gad46e8c7921f4ad0181f527670eeb1e27)

As I said, I'm leaving this here to see what you think of these approaches before putting more effort into them.

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -100,15 +100,4 @@
 # define TRUE 1
 #endif
 
-#if defined(MRB_BUILD_AS_DLL)
-
-#if defined(MRB_CORE) || defined(MRB_LIB)
-#define MRB_API __declspec(dllexport)
-#else
-#define MRB_API __declspec(dllimport)
-#endif
-#else
-#define MRB_API extern
-#endif
-
 #endif  /* MRUBYCONF_H */
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1,28 +1,30 @@
 /*
-** mruby - An embeddable Ruby implementation
-**
-** Copyright (c) mruby developers 2010-2015
-**
-** Permission is hereby granted, free of charge, to any person obtaining
-** a copy of this software and associated documentation files (the
-** "Software"), to deal in the Software without restriction, including
-** without limitation the rights to use, copy, modify, merge, publish,
-** distribute, sublicense, and/or sell copies of the Software, and to
-** permit persons to whom the Software is furnished to do so, subject to
-** the following conditions:
-**
-** The above copyright notice and this permission notice shall be
-** included in all copies or substantial portions of the Software.
-**
-** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
-** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
-** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
-** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
-** SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-**
-** [ MIT license: http://www.opensource.org/licenses/mit-license.php ]
+ * mruby - An embeddable Ruby implementation
+ *
+ * Copyright (c) mruby developers 2010-2015
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be
+ * included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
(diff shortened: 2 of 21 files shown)


matz:
Looks good. But this pull-request has several conceptual changes in one.  Next time, could you separate pull-requests for each concept (and gain more contribution measurement)?

bovi:
Hi @sagmor,

this is great work. I did some time ago work on the mrbdoc tool and we integrated it into the official website (https://github.com/mruby/mruby.github.io/blob/master/gen_mrbdoc.rb). Your generated stuff looks much better and I would encourage you (of course only if you have time and fun) to consider a pull request to exchange the mrbdoc integration in the mruby.github.io repo [https://github.com/mruby/mruby.github.io] with your variant. In this way we can keep the documentation via your tool up-to-date on the official website.

sagmor:
@matz sorry for the "several conceptual changes in one" but things came out as I was playing with the headers. Glad you merged it though.

@bovi, That was my idea. I'm pushing the generation script into mruby.org repo right now.

cremno:
Why do we need `MRB_INLINE`?

takahashim:
I guess that:

* Macro has no type information (without preprocessor), so use function declarations
* Function call has overhead, so use inline functions
* Inline definition has compatibility issue, so define and use `MRB_INLINE`

furunkel:
Should probably be named `mrb_inline` for consistency, though ?

sagmor:
The problem is that MRB_API (`extern`) is mutually exclusive with MRB_INLINE (`inline`)
I tried to place them together and failed miserably :stuck_out_tongue: 

```
CC    src/array.c -> build/host/src/array.o
In file included from /Users/sagmor/Development/mruby/mruby/src/array.c:7:
/Users/sagmor/Development/mruby/mruby/include/mruby.h:288:9: error: cannot combine with previous 'extern' declaration specifier
MRB_API MRB_INLINE mrb_value mrb_class_new_instance(mrb_state *mrb, mrb_int argc, const mrb_value *argv, struct RClass *c)
        ^
/Users/sagmor/Development/mruby/mruby/include/mruby/common.h:51:21: note: expanded from macro 'MRB_INLINE'
# define MRB_INLINE static inline
                    ^
1 error generated.
In file included from /Users/sagmor/Development/mruby/mruby/src/array.c:7:
/Users/sagmor/Development/mruby/mruby/include/mruby.h:288:9: error: cannot combine with previous 'extern' declaration specifier
MRB_API MRB_INLINE mrb_value mrb_class_new_instance(mrb_state *mrb, mrb_int argc, const mrb_value *argv, struct RClass *c)
        ^
/Users/sagmor/Development/mruby/mruby/include/mruby/common.h:51:21: note: expanded from macro 'MRB_INLINE'
# define MRB_INLINE static inline
                    ^
1 error generated.
rake aborted!
```

sagmor:
As @takahashim says, inline functions keeps the fact that It's a function without masking it under a "constant" that makes a lot easier to document C Headers consistently (all while keeping the performance of the `#define`).

Right now I'm playing with that fact to further improve the docs with a [yard plugin](https://github.com/sagmor/yard-mruby/) it's a bit early yet but [here's a demo](http://sagmor.com/mruby/) of it's current state when executed on MRuby's core

cremno:
>The problem is that MRB_API (`extern`) is mutually exclusive with MRB_INLINE (`inline`)

`extern` and `static` are mutually exclusive. `extern` and `inline` aren't.

Let me rephrase my question: what's the rationale behind the introduction of `MRB_INLINE`? Just consistency with `MRB_API`? Or why is directly writing `static inline` not acceptable?

>Right now I'm playing with that fact to further improve the docs with a yard plugin it's a bit early yet but here's a demo of it's current state when executed on MRuby's core

That's great! (I like yard.)

sagmor:
@cremno Actually I took the inspiration from [libgit2](https://github.com/libgit2/libgit2/blob/cf716beed2f4729cf01a9822d75b4f1ebf7ccbf1/src/common.h#L13-L18) and to be honest `MRB_INLINE` was a suggestion to turn `defines` into actual functions for API consistency but it's open for debate if anyone has a better idea :+1: 

I will make a few experiments defining `mrb_inline` as just `inline` and see how it plays along and see, not sure how the `static`, `extern` and `inline` keywords behave together.

And by the way, The docs are getting better (See: [sagmor.com/mruby](http://sagmor.com/mruby/headers/mruby.h.html))

cremno:
http://www.greenend.org.uk/rjk/tech/inline.html is a good explanation.

I think the linked libgit2 code is a bit too cautious. I don't know its history though. Maybe there was a reason to define it that way.

However mruby already provides `inline` (as macro which expands to `__inline`) for older MSVC versions. So in my opinion `MRB_INLINE` doesn't improve or solve anything. Would it be okay to remove it and use `static inline` again instead?

sagmor:
@cremno Further investigating I found out there's an `__attribute__((always_inline))` that

> According to the GCC User Manual one can force a function to be inlined using the always_inline attribute

But I don't have enough info to make an informed choice.
Now from what I see these are the options we have:

```c
1:
#define mrb_foo /* code */
/*
 * What we have now, fast and easy.
 *  but in a way inconsistent (for documentation purposes)
 */

2:
#define MRB_INLINE static inline __attribute__((always_inline))
MRB_INLINE void mrb_foo( void ) { /* code */ }
/*
 * We turn the define into a function and get more context for the documentation
 */

3:
static inline void mrb_foo( void ) { /* code */ }
/*
 * Pretty much the same as above.
 * Not sure if the  __attribute__ has a real impact, and the need of a 
 * MRB_INLINE define seems like a keyword to keep it visually consistent with MRB_API
 */

4:
#define mrb_inline inline __attribute__((always_inline))
MRB_API mrb_inline void mrb_foo( void ); /* and code goes on some .c file */
/*
 * I like how this looks.
 * mrb_inline is consistent with other "flags" like mrb_deprecated or mrb_noreturn
 * but it forces the function's code to live on a different .c file. Which means more
 * work but also means the implementation get's hidden/encapsulated
 */
```

From how it looks, I like the fourth option but I have no idea if there's a real performance difference on each option and how far we should move to replace those `define`s

Issue #2968, opened 2015-09-28
Array#index is missing its block overload


IceDragon200:
Array#index does not take a block

eg.
```ruby
ary = [1, 2, 3, 4, 5]
# works with an object
ary.index(2) #=> 1

# now with a block
ary.index { |n| n == 3 }

# CRuby - expected
# => 2

# mruby - actual
# => ArgumentError
```

PR #2970, opened 2015-09-29
use Enumerable#find_index as Array#index

takahashim:
note: With this patch, Array includes Enumerable.

cf. #2968


takahashim:
Sorry, I fixed tests.

PR #2977, opened 2015-10-06
Add how to use  mrb_gc_arena_save()/mrb_gc_arena_restore() doc

tatsuhiro-t:
This adds documentation of mrb_gc_arena_save()/mrb_gc_arena_restore() functions.
As the documentation says, this is basically the English translation of Matz's blog post in Japanese back in 2013.  So first we have to ask @matz the permission to translate his post.
As far as I know, there is no official documentation about these functions in this repository.
Also I think that Matz's blog post is very well written.  I understand what these functions really do and their purpose.
It has 2 problems though.  First one is that it is written in Japanese, never get into official doc.  Second problem is that it is now outdated.  I read the post and learned that the size of stack is limited to 100, but the reality is it is now unlimited (true?).
Some say that they are really thin, and no overhead, so surround the every single mrb API call.  But this cannot handle the situation where mrb_gc_protect() can solve, and could even be dangerous.  We need proper understanding of them.
I think these functions are really important when writing C program using mruby API, but lacking documentation is really a sad situation.
I hope this PR improve it.
I really don't know how this kind of documentation fits in this repository.  I thought that this is indeed public API functions, so I put this into doc/api.
Any suggestions are welcome.

### DIFF
--- a/doc/api/gc-arena-howto.md
+++ b/doc/api/gc-arena-howto.md
@@ -1,0 +1,171 @@
+# How to use `mrb_gc_arena_save()`/`mrb_gc_arena_restore()`
+
+This is basically English translation of [Matz's blog post](http://www.rubyist.net/~matz/20130731.html) written in Japanese.
+Some parts are updated to reflect recent changes.
+
+When you are extending mruby using C language, you may encounter
+mysterious "arena overflow error" or memory leak or very slow
+execution speed.  This is an error indicating overflow of "GC arena"
+implementing "conservative GC".
+
+GC (garbage collector) must ensure that object is "alive", in other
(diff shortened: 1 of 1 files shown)


matz:
Of course you are OK to translate any of my blog posts. Thank you!

Asmod4n:
The example given is a bit compilcated, am i doing it right here? https://github.com/Asmod4n/mruby-czmq/blob/master/src/mrb_czmq.c#L654

i first create a array with a capa, then store the arena idx and after i have pushed a value into the array i restore the arena.

schmurfy:
thanks for translating this, really helpful :)

PR #2983, opened 2015-10-08
WIP - New Android NDK toolchain

sagmor:
Following the merge of #2978 I wanted to better approach Android cross-compilation.

This PR let's you configure android builds in the following way:
```ruby
MRuby::CrossBuild.new("build-name") do |conf|
  toolchain :androidndk, arch: 'x86'

  # The rest of your setup
end
```

The main advantage is that it let's you build multiple architectures at once.
For example you could cross-compile to every Android architecture with something like:
```ruby
MRuby::Toolchain::AndroidNDK::ARCHITECTURES.each do |arch|
  MRuby::CrossBuild.new("android-#{arch}") do |conf|
    toolchain :androidndk, arch: arch

    # The rest of your setup
  end
end
```

The available options are:
 * `:arch`: Target architecture (defaults to 'armeabi')
 * `:toolchain`: The underlying toolchain can be `:gcc` or `:clang` (defaults to `:gcc`)
 * `:toolchain_version`: Toolchain version. automatically finds the latest available.
 * `:platform`: Android platform (defaults to `'android-14'`)
 * `:ndk_home`: Path to the Android NDK, defaults to `ENV['ANDROID_NDK_HOME']` or search in known locations like `/usr/local/opt/android-ndk`

Some missing point's still under development:
- [ ] Test clang toolchain
- [ ] Build libs as shared objects (`libmruby.so`) to directly add into android packages

I'm leaving this here so others working with the `:androideabi` toolchain could test it in their projects.

### DIFF
--- a/tasks/toolchains/androidndk.rake
+++ b/tasks/toolchains/androidndk.rake
@@ -3,9 +3,10 @@ class MRuby::Toolchain::AndroidNDK
   DEFAULT_ARCH = 'armeabi'
   DEFAULT_PLATFORM = 'android-14'
   DEFAULT_TOOLCHAIN = :gcc
-  KNOWN_NDK_HOMES = %w{
+  DEFAULT_NDK_HOMES = %w{
     /usr/local/opt/android-ndk
   }
+  TOOLCHAINS = [:gcc, :clang]
   ARCHITECTURES = %w{
     armeabi armeabi-v7a arm64-v8a
     mips mips64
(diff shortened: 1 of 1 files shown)


furunkel:
Opinionated nitpick, I'd call the module `Android` (and the toolchain too) and drop the `NDK`, looks nicer that way, IMO.

sagmor:
@furunkel I like opinionated :stuck_out_tongue:

And it makes sense to call it just android.

matz:
After you rename AndroidNDK to Android, I will merge.

PR #2986, opened 2015-10-09
fix `rake doc` error

takahashim:
`doc/language/generator.rb` has been removed, so use `mrbdoc`


takahashim:
@mattn Oops,  thanks for your comment!   fixed.

mattn:
:+1:

PR #2988, opened 2015-10-11
Update androidndk.rake to add support 64bit architectures.

crimsonwoods:
To support 64bit architectures (arm64, mips64 and x86_64), androidndk.rake have to be modified.
We can build libmruby.a for 64bit architecture like below:

```Ruby
# add codes below into your build_config.rb
MRuby::CrossBuild.new('android-arm64') do |conf|
  toolchain :androidndk, arch: 'arm64', platform: 'android-21' # 'android-19' or before cannot be used to build for 64bit architecture
end
```

### DIFF
--- a/tasks/toolchains/androidndk.rake
+++ b/tasks/toolchains/androidndk.rake
@@ -62,6 +62,8 @@ Set ANDROID_NDK_HOME environment variable or set :ndk_home parameter
           'mips64el-linux-android-*'
         when /mips/
           'mipsel-linux-android-*'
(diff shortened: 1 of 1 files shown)


crimsonwoods:
@matz should I reduce/integrate my commits to suppress the cost of reviewing?

matz:
Please minimize your cosa

crimsonwoods:
@matz thank you! I would prefer to continue your review without changes.

PR #2993, opened 2015-10-15
Updated mruby.h YARD doc.

Mav7:


build_config.rb
```diff
@@ -21,7 +21,7 @@
 
   # include the default GEMs
   conf.gembox 'default'
-
+  conf.gem '/home/thamav/tests/mrbgems_test'
```
zzak: How did this get here?

Mav7: Oops sorry my mistake thanks for catching that!

PR #2994, opened 2015-10-15
updated YARD docs on mruby.h

Mav7:


sagmor:
I'm not sure I get the examples.

`mrb_obj_new` and `mrb_class_new_instance` seem to be quite equivalent (just a change in parameters order)

Shouldn't it be more like?:
```c
example_class = mrb_define_class(mrb, "ExampleClass", mrb->object_class);
// The same as class ExampleClass; end

obj1 = mrb_obj_new(mrb, example_class, 0, NULL);
// The same as ExampleClass.new

obj2 = mrb_class_new_instance(mrb, 0, NULL, example_class);
// Also the same as ExampleClass.new
```

Not sure if I got the method right, havent had time to play with it yet

Mav7:
Not exactly when I evaluate the output of them respectively I get:
```C
mrb_obj_new(mrb, mrb->object_class, 1, argv);
ExampleClass
```
```C
mrb_class_new_instance(mrb, 0, argv, example_class);
#<ExampleClass:0x89734f8>
```
Also I'm not sure if NULL should be used in there it's probably better to use mrb_nil_value() instead since mrb_obj_new's last parameter is an mrb_value.

cremno:
Different results are not surprising when the examples aren't equivalent. Just compare the variables passed to `mrb_funcall()`:
```c
mrb_funcall(mrb, obj, "puts", 1, example_class);  // puts ExamplesClass
```
```c
mrb_funcall(mrb, obj, "puts", 1, obj_inst); // puts ExampleClass.new
```

The first one is also wrong since the type of `example_class` is `struct RClass*` and not `mrb_value`. It just happens to work (which isn't good at all).

Mav7:
Ok that sounds good I'll do that.

PR #2997, opened 2015-10-19
GC cleanup

furunkel:
This restructures and refactors the GC code. This is part of something bigger, but is useful on its own.
I would have run the GC tests if I'd known how to do that ?

All GC related data is now encapsulated in the `mrb_gc` struct. Related declarations, definitions and macros have been moved to `gc.h` or replaced with something better.
In particular, the ugly `is_dead`, `is_black` etc. macros are no longer public and there are no more
GC related unprefixed structure declarations.

This code makes use of C99 flexible array members. Apparently these are even supported by recent
versions of MSVC. However, if you want to support older versions, we might need to add a hack.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -65,10 +65,6 @@ struct mrb_state;
  */
 typedef void* (*mrb_allocf) (struct mrb_state *mrb, void*, size_t, void *ud);
 
-#ifndef MRB_GC_ARENA_SIZE
-#define MRB_GC_ARENA_SIZE 100
-#endif
-
 #ifndef MRB_FIXED_STATE_ATEXIT_STACK_SIZE
 #define MRB_FIXED_STATE_ATEXIT_STACK_SIZE 5
 #endif
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -35,6 +35,7 @@
 #include "mrbconf.h"
 #include "mruby/common.h"
 #include "mruby/value.h"
+#include "mruby/gc.h"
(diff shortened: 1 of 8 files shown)


furunkel:
@matz, any complaints or comments about this ?

matz:
Thank you.,
First, I don't think `mrb_gc` is an appropriate struct name. Maybe `mrb_gc_state`? Ah but we have already used that name.
Second, it conflicts with the latest.  Can you rebase?
Third, I worry a little about performance. But probably it's OK.

furunkel:
@matz, the GC test do fail, **even in master** ? I guess they are not relevant anymore ?
I found `mrb_gc` to be quite natural because literally, `mrb->gc_field` now becomes `mrb->gc.field`
and this struct really represents the garbage collector, but of course I am open to suggestions,
any ideas anyone ?
Switching names with what is now `mrb_gc_state` would probably be confusing ?

As for performance, these are mostly just renamings, anything particular you worry about ?

cremno:
>BTW, anyone using MSVC who can ensure flexible array members are no problem ?

They're not a problem: https://msdn.microsoft.com/en-us/library/b6fae073%28v=vs.140%29.aspx.

matz:
I just worried `mrb->gc.field` requires one more redirection, but it seems to be optimizable (thus ignorable) probably.  Ok, I will merge.  If someone come up with better name for `mrb_gc`, we can update.

furunkel:
Thanks !

PR #2999, opened 2015-10-20
More Docs

sagmor:
Now docs are covering most of the sources including mani fixes to parse some special cases and include all mrbgems in the generated docs.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -1,3 +1,4 @@
+--markup markdown
 --plugin mruby
 --plugin coderay
 --output-dir doc/api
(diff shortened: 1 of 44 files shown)


matz:
I will merge this soon.  But I am not fully satisfied with enormous `@mrbgem` tags.
Is there any other way?

sagmor:
I could try to figure out a way.
I could use the fact that those files live inside `mrbgems` directory to tag them "automagically" :sparkles: 

What if:
I revert 5cdcce8dbddd94ecb9503a0a1d47370c4ef97177 and give it more thought time to skip the abundant `@mrbgem` tags?

matz:
@sagmor that's better. I will merge this after your revert.

sagmor:
@matz :heavy_check_mark:

PR #3011, opened 2015-11-14
Added more YARD docs to mruby.h

Mav7:
Wrote reference documentation for mrb_obj_dup, and mrb_obj_respond_to. I Also fixed a mistake I made on mrb_class_defined.

### DIFF
--- a/build_config.rb
+++ b/build_config.rb
@@ -21,7 +21,6 @@ MRuby::Build.new do |conf|
 
   # include the default GEMs
   conf.gembox 'default'
-  
   # C compiler settings
   # conf.cc do |cc|
   #   cc.command = ENV['CC'] || 'gcc'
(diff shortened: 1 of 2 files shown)


strcmp:
@Mav7 should valid YARD documentation not include a type annotation also, like:
`@param [Foo] name desc` ? or is it somehow different in mruby?

Mav7:
@strcmp  I think your right my friend! I'll start fixing the docs to fit this. Maybe you can review my new docs later, and tell me if it looks right.

Mav7:
@strcmp When I write `@param [mrb_state *] mrb The current mruby state.` I get `mrb (mrb_state *) — The current mruby state.` In the generated YARD doc. Does that look right?

strcmp:
yeah that looks correct @Mav7, I think everything within "[]" is free form so it should not matter if you are describing ruby or C code, as long as it makes sense to programmers of either :)

Mav7:
@strcmp  Perfect thanks a lot! This is the power of open source!

strcmp:
happy to help.

PR #3014, opened 2015-11-16
rename ENABLE_DEBUG to MRB_DEFINE_HOOKS

cremno:
Here are some reasons why it should be renamed:
- `ENABLE_DEBUG` isn't prefixed (possible name clash with other libs)
- the `enable_debug` build config method is totally unrelated to this
- the new name MRB_DEFINE_HOOKS is more precise

`ENABLE_DEBUG` and `DISABLE_DEBUG` will still work for backward compatibility but mruby users should update their code. E.g. to `#if defined MRB_DEFINE_HOOKS || defined ENABLE_DEBUG`.

### DIFF
--- a/doc/guides/mrbconf.md
+++ b/doc/guides/mrbconf.md
@@ -23,14 +23,15 @@ You can use mrbconfs with following ways:
   * Printing features in **src/print.c**.
 
 ## Debug macros.
-`ENABLE_DEBUG`
+`MRB_DEFINE_HOOKS`
 * When defined code fetch hook and debug OP hook will be enabled.
 * When using any of the hook set function pointer `code_fetch_hook` and/or `debug_op_hook` of `mrb_state`.
 * Fetch hook will be called before any OP.
 * Debug OP hook will be called when dispatching `OP_DEBUG`.
+* Defines `ENABLE_DEBUG` and vice versa (backward compatibility).
 
 `DISABLE_DEBUG`
-* Will be define automatically if `ENABLE_DEBUG` isn't defined.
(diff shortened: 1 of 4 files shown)


matz:
It's OK to rename it, but I don't think `MRB_DEFINE_HOOK` is not precise. It sounds like the hook itself, besides that, it doesn't specify which hook we are going to add. How about 'MRB_ENABLE_CODE_FETCH_HOOK`. The only concern is it's too long.

matz:
And what do  you think about another `ENABLE_` macro, `ENABLE_STDIO`?

zzak:
In any case, I think it should be prefixed with `MRB_` to avoid conflicts with other libraries.

cremno:
What about `MRB_DEBUG_HOOKS`? MY only concern is that people might still think it's related to `MRB_DEBUG`.

What do you think about `MRB_USE_STDIO`?

cremno:
Or `MRB_STATE_DEFINE_HOOKS` or just `MRB_STATE_HOOKS`?

matz:
I picked `MRB_ENABLE_DEBUG_HOOK` for consistency with `(MRB_)DISABLE_STDIO`.

cremno:
Okay. By the way, I've proposed `MRB_NO_STDIO` because mruby already has `MRB_NO_INIT_ARRAY_START`.

zzak:
@matz @cremno The only problem I have with the naming `MRB_ENABLE_DEBUG_HOOK` is it's similar to `MRB_DEBUG`. This is probably `MRB_DEBUG`s fault for being too generic though.

Thoughts?

matz:
@zzak probably I will rename `MRB_DEBUG` later, but not this time.

PR #3015, opened 2015-11-16
fix comment

murasesyuka:
fix_lshift & fix_rshift can return float.


matz:
Since mruby does not have bignums, overflowed integer operations can create float numbers as results.

murasesyuka:
Yes. I see

PR #3021, opened 2015-11-17
Fix MRB_ENABLE_DEBUG_HOOK config typo left from 4440566

zzak:
/cc #3014


zzak:
Also this wasn't caught by travis because we don't have a debug enabled config case.

Should we have one?

PR #3027, opened 2015-11-19
Fix check for VisualStudio Version in mruby-math gem

pbosetti:
erf() and other similar mathematical functions are available in MSVC starting from VS2012 (_MSC_VER == 1700).


cremno:
I don't have 2012 installed anymore but I believe you're wrong. For example, http://blogs.msdn.com/b/vcblog/archive/2013/07/19/c99-library-support-in-visual-studio-2013.aspx clearly says they were added in 2013.

PR #3039, opened 2015-11-30
Reject next, break and return with blocks.

sgnr:
Statements such as `next [] do; end` cause a segfault on latest master. This is due to the fact that it is parsed as a block_call, yet `call_with_block `cannot handle these nodes. `yield` with a block was already rejected, so I added a bunch of similar checks for `next`, `break` and `return`.

### DIFF
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -2555,6 +2555,15 @@ block_call      : command do_block
                       if ($1->car == (node*)NODE_YIELD) {
                         yyerror(p, "block given to yield");
                       }
+                      else if ($1->car == (node*)NODE_NEXT) {
+                        yyerror(p, "block given to next");
+                      }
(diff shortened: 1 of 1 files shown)


sgnr:
hum it seem it rejects code such as
```ruby
def a(n)
  n + yield(7)
end

def b
  return a 2 do |x|
      x * 10
  end
end

puts b
```
that MRI handles correctly.

clayton-shopify:
MRI does reject `a 2 do |x|` following a `yield` (with "block given to yield") so maybe it's justified for mruby to do so after `next`, `break` and `return` even if it's inconsistent with MRI. It's better than crashing at least.

clayton-shopify:
In case they are helpful for testing, some other statements that presently crash mruby:

`next a 0 do ; end`
`break a 0 do ; end`
`return a 0 do ; end`
`next {} do ; end`
`a a+a next yield do ; end`

PR #3044, opened 2015-12-10
sprintf may throw a exception with wrong message when argument is a too long shared string.

tsahara:
```
% bin/mruby -e 'a="1"*100; t=a[0,30]; sprintf("%u", t)'
trace:
	[0] -e:1
-e:1: string (1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111) too big for integer (ArgumentError)
```

The number of "1" characters should be only 30, not 100.  Another case is:

```
% bin/mruby -e 'a="a"*100; t=a[0,30]; sprintf("%u", t)'
trace:
	[0] -e:1
-e:1: invalid string for number("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa") (ArgumentError)
```

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -2163,7 +2163,7 @@ mrb_str_len_to_inum(mrb_state *mrb, const char *str, size_t len, int base, int b
     n *= base;
     n += c;
     if (n > MRB_INT_MAX) {
-      mrb_raisef(mrb, E_ARGUMENT_ERROR, "string (%S) too big for integer", mrb_str_new_cstr(mrb, str));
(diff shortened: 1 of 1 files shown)


tsahara:
The 2nd part of the patch was not applied and the 2nd case of this issue is not resolved.  Why was it removed?

PR #3048, opened 2015-12-16
converting string representation of MRB_INT_MIN raises an exception

tsahara:
Without this patch, `Integer("-2147483648")` raises an exception.  But it can be conveted to a Fixnum.

Before:
```
% bin/mruby -e 'Integer("-2147483648")'
trace:
        [0] -e:1
-e:1: string (-2147483648) too big for integer (ArgumentError)
```

After:
```
% bin/mruby -e 'p Integer("-2147483648")'
-2147483648
```


matz:
I will fix this with some code cleanup.

PR #3049, opened 2015-12-16
Integer("1_0") raises an exception

tsahara:
`"1_0"` is a string to be converted to a number but `Integer("1_0")` raises an exception.

```
% bin/mruby -e 'p Integer("1_0")'
trace:
        [0] -e:1
-e:1: invalid string for number("1_0") (ArgumentError)

% ruby -e 'p Integer("1_0")'
10
```


matz:
I fixed it with some code cleanup. Will check in soon.

PR #2917, opened 2015-08-17
Mark backtrace referenced by exception

kou:
The following code crashes without this change:

    def a
      [1].each do
        [2].each do
          [3].each do
            raise "XXX"
          end
        end
      end
    end

    begin
      a
    rescue => exception
      GC.start
      exception.backtrace
    end

GDB backtrace:

    Program received signal SIGSEGV, Segmentation fault.
    strlen () at ../sysdeps/x86_64/strlen.S:106
    106	../sysdeps/x86_64/strlen.S: No such file or directory.
    (gdb) bt
    #0  strlen () at ../sysdeps/x86_64/strlen.S:106
    #1  0x00000000004252cd in mrb_str_new_cstr (mrb=0x69f010,
        p=0x101 <error: Cannot access memory at address 0x101>)
        at mruby/src/string.c:290
    #2  0x00000000004183fe in get_backtrace_i (mrb=0x69f010, loc=0x7fffffffd940,
        data=0x6a7410) at mruby/src/backtrace.c:72
    #3  0x0000000000418793 in output_backtrace (mrb=0x69f010, ciidx=8,
        pc0=0x71940c, func=0x4183af <get_backtrace_i>, data=0x6a7410)
        at mruby/src/backtrace.c:140
    #4  0x0000000000418862 in exc_output_backtrace (mrb=0x69f010, exc=0x6a5be0,
        func=0x4183af <get_backtrace_i>, stream=0x6a7410)
        at mruby/src/backtrace.c:157
    #5  0x000000000041894c in mrb_exc_backtrace (mrb=0x69f010, self=...)
        at mruby/src/backtrace.c:199
    #6  0x000000000040dbaf in mrb_context_run (mrb=0x69f010, proc=0x6a61b0,
        self=..., stack_keep=0) at mruby/src/vm.c:1126
    #7  0x00000000004131d8 in mrb_toplevel_run_keep (mrb=0x69f010, proc=0x6a61b0,
        stack_keep=0) at mruby/src/vm.c:2422
    #8  0x000000000043a46c in load_exec (mrb=0x69f010, p=0x6f6450, c=0x6c9320)
        at mruby/mrbgems/mruby-compiler/core/parse.y:5619
    #9  0x000000000043a4e2 in mrb_load_file_cxt (mrb=0x69f010, f=0x6f61f0,
        c=0x6c9320)
        at mruby/mrbgems/mruby-compiler/core/parse.y:5628
[...]

PR #3065, opened 2015-12-29
Support backtrace after method calls

### LINKED ISSUE #2902: broken exception backtraces
```
def raise_func
  raise "noise"
end

def err_func
  raise_func
end

def use_some_more_stack
  puts "used some more stack"
end

def use_some_stack
  puts "used some stack"
  use_some_more_stack
end

def func1
  begin
    err_func
  rescue => e
    use_some_stack
    e.backtrace.each { |entry|
      p entry
    }
  end
end

def func2
  begin
    err_func
  rescue => e
    e.backtrace.each { |entry|
      p entry
    }
  end
end
      

func1
func2
```
$>ruby broken_backtrace.rb 
used some stack
used some more stack
"broken_backtrace.rb:2:in raise_func"
"broken_backtrace.rb:6:in err_func"
"broken_backtrace.rb:20:in func1"
"broken_backtrace.rb:40:in main"
"broken_backtrace.rb:2:in raise_func"
"broken_backtrace.rb:6:in err_func"
"broken_backtrace.rb:31:in func2"
"broken_backtrace.rb:41:in main"

$>mruby broken_backtrace.rb 
used some stack
used some more stack
"broken_backtrace.rb:10:in Object.use_some_more_stack"
"broken_backtrace.rb:23:in Object.func1"
"broken_backtrace.rb:40"
"broken_backtrace.rb:2:in Object.raise_func"
"broken_backtrace.rb:33:in Object.func2"
"broken_backtrace.rb:41

I see several discrepancies:
1. err_func is not mentioned in the backtraces given by mruby
2. line numbers differ for 'func1' and 'func2' entries between mruby and ruby
3. in backtrace for 'func1',  'raise_func' is replaced by 'use_some_more_stack'  even though it was not on the call stack when the exception was thrown, but was called right before printing the backstack

kou:
GitHub: fix #2902, #2917

The current implementation traverses stack to retrieve backtrace. But
stack will be changed when some operations are occurred. It means that
backtrace may be broken after some operations.

This change (1) saves the minimum information to retrieve backtrace when
exception is raised and (2) restores backtrace from the minimum
information when backtrace is needed. It reduces overhead for creating
backtrace Ruby objects.

The space for the minimum information is reused by multiple
exceptions. So memory allocation isn't occurred for each exception.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -113,6 +113,14 @@ struct mrb_context {
 
 struct mrb_jmpbuf;
 
+typedef struct {
+  const char *filename;
+  int lineno;
+  struct RClass *klass;
+  const char *sep;
+  mrb_sym method_id;
+} mrb_backtrace_entry;
+
 typedef void (*mrb_atexit_func)(struct mrb_state*);
 
 typedef struct mrb_state {
@@ -125,6 +133,12 @@ typedef struct mrb_state {
   struct mrb_context *root_c;
 
   struct RObject *exc;                    /* exception */
+  struct {
+    struct RObject *exc;
+    int n;
+    int n_allocated;
+    mrb_backtrace_entry *entries;
+  } backtrace;
   struct iv_tbl *globals;                 /* global variable table */
 
   struct RObject *top_self;
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -14,6 +14,15 @@
 #include <mruby/error.h>
 #include <mruby/numeric.h>
 
+struct backtrace_location_raw {
+  int i;
+  int lineno;
+  const char *filename;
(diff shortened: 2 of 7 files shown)


matz:
Travis failed.

kou:
Do you know how to disable backtrace related tests when `-g` isn't passed to `mrbc`?
Backtrace related tests are meaningless when we use `mrbc` without `-g` because `mrbc` without `-g` doesn't generate file/lineno information.

kou:
0ebac02813d6506f92c9aaceaa00c6f902a56a03 is an implementation of disabling backtrace related tests when backtrace isn't available.

matz:
Looks reasonable.

PR #3068, opened 2015-12-31
use _setjmp/_longjmp on OS X

kazuho:
setjmp/longjmp are very slow on OS X; _setjmp/_longjmp should be used for implementing TRY-CATCH.

The reason is as stated in the man pages of OSX, quoting:
> ```
> The setjmp()/longjmp() pairs save and restore the signal mask while _setjmp()/_longjmp() pairs save and restore only
> the register set and the stack.  (See sigprocmask(2).)
> ```

With the change, benchmark/bm_ao_render.rb became 20% faster (11.791 seconds => 9.668 seconds) on my MacBook Pro.

Shown below is the profile result of the master branch running mb_ao_render.rb on OS X.
Profile result of master running bm_ao_render.rb

### DIFF
--- a/include/mruby/throw.h
+++ b/include/mruby/throw.h
@@ -20,11 +20,19 @@ typedef mrb_int mrb_jmpbuf_impl;
 
 #include <setjmp.h>
 
-#define MRB_TRY(buf) do { if (setjmp((buf)->impl) == 0) {
+#if defined(__APPLE__)
+#define MRB_SETJMP _setjmp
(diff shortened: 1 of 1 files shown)


kazuho:
I assume it is the case for all BSD-based operating systems, since https://www.freebsd.org/cgi/man.cgi?query=setjmp&sektion=3 and http://netbsd.gw.com/cgi-bin/man-cgi?setjmp+3+NetBSD-1.5 shared the same statement.

PR #3069, opened 2016-01-02
Original2

miura1729:
see https://github.com/mruby/mruby/issues/3917


miura1729:
see https://github.com/mruby/mruby/issues/3917

PR #3075, opened 2016-01-07
first argument to `%.*s` must be int

kazuho:
Result type of `RSTRING_LEN` is `mrb_int`; it needs to be casted to `int` before passed to sprintf as the length argument of `%.*s` in case `sizeof(mrb_int)` is not equal to `sizeof(int)`.


syohex:
This is same as https://github.com/mruby/mruby/pull/3074 and it was merged.

kazuho:
Oh!  Sorry for the fuss.

Issue #3085, opened 2016-01-15
Add more info to README


cllns:
Hi there,

As someone who doesn't know much about `mruby`, there are three things I'm curious about:
- What pieces of Ruby are missing
- Benchmarks (*Approximate or [basic](https://youtu.be/XZNfQTw1IVw?t=16m39s) is fine*)
- Benefits, i.e. what embedded means and how it's done, from user's perspective (*This is already there, sort of*) 

It'd be great if all of that was on the README. It could also be good info for the homepage.
Just thought I'd offer that feedback :) 

Thanks!

matz:
Among your proposals, benchmark is difficult to add, because working platforms varies micro computers to UNIX machines with very different performance characteristics. Others are very good to add. I will try (or if someone come up with a pull-request, I'd be more than happy).

strcmp:
i think writing such a thing requires confidence in how the project works and the philosophy behind it. to me that seems difficult, even if english is your primary language but i could be wrong, that's my general feeling.

strcmp:
maybe adapting some of the content in http://zzak.io/log/2016-01-15-the-pitch-for-mruby.html to the readme would be useful.

shevegen:
I think an overview might be nice for proposal #1 - "What pieces of Ruby are missing". For instance, I believe that Encoding in mruby is not 100% the same as in MRI; to what extent I do not know myself since I use almost exclusively MRI (I did play around with mruby/mirb though).

This could be like a yearly snapshot - at around the time of XMas when the year is over (for those who count in that year-format),  when a new MRI ruby will be released, mruby could also update on what has changed, the major steps from year to year - if only for those people interested to keep track of how far along mruby has proceeded by now :)

matz:
Closing — README info-suggestion from 2016. README has since gained a Table of Contents, install paths via package managers, and links to doc/guides/ (capi.md, getting-started.md, language.md). For benchmark and language-coverage info, those guides plus per-release docs (doc/mruby*.md) cover the original request.

PR #3098, opened 2016-02-10
Limitations

bovi:
As suggested in issue #3085 it would be interesting to get an overview which limitations of mruby are "by design". I started here with these two:
  - Kernel.raise in rescue
  - Array passed to puts

If the overall style is accepted by @matz I will proceed adding.


zzak:
No objections here, thanks for taking this on!

Issue #3114, opened 2016-02-18
Sending dup to a Struct raises a TypeError


sgnr:
How to reproduce:
```
> A = Struct.new(:a, :b)
 => #<Class:0xf6d0c0>
> A.new(1, 2).dup
(mirb):2: struct size mismatch (TypeError)
```

PR #3115, opened 2016-02-19
Fix Struct#dup issue

syohex:
Remove needless assertion. The argument 'copy' is always empty array so this assertion is failed with 'struct' which has more than one 'members'.

This is related to #3114.
CC: @sgnr, @mattn


matz:
Sorry I fixed this issue by myself in cf5ca03 without noticing this PR.

PR #3121, opened 2016-02-26
The original code crashed when mrb->backtrace.n grew to 16.

asfluido:
It looks like the logic to reallocate the backtrace was flawed,
based on the wrong variable (loc_raw->i, which, as I have verified,
decreases from 16 to 0 instead of increasing)

I am not sure if this is the correct fix

### DIFF
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -330,7 +330,7 @@ save_backtrace_i(mrb_state *mrb,
 {
   mrb_backtrace_entry *entry;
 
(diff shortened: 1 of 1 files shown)


matz:
@kou may know the detail, but your assumption looks right to me.

kou:
I'll look into it and comment about it later.

kou:
I've confirmed that my patch had a bug. (Sorry. :<)
This fix is right.

PR #3123, opened 2016-03-01
Allow to configure integer division behavior

mame:
Hello,

This PR allows a user to change the behavior of integer division.
By adding the following line to config.rb,

```
conf.cc.defines = %w(MRB_INTEGER_DIVISION)
```

mruby will evaluate `1/2` to `0` rather than `0.5`.

What do you think?

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1936,7 +1936,29 @@ RETRY_TRY_BLOCK:
           mrb_int x = mrb_fixnum(regs[a]);
           mrb_int y = mrb_fixnum(regs[a+1]);
 #ifdef MRB_INTEGER_DIVISION
-          SET_INT_VALUE(regs[a], x / y);
+          if (y == 0 || (x == MRB_INT_MIN && y == -1)) {
+             SET_FLOAT_VALUE(mrb, regs[a], (mrb_float)x / (mrb_float)y);
+          }
+          else {
+            mrb_int div, mod;
+            if (y < 0) {
+              if (x < 0)
+                div = -x / -y;
+              else
+                div = - (x / -y);
+            }
+            else {
+              if (x < 0)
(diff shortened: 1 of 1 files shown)


cremno:
Your change doesn't care about the integer division issue:  `y == 0` or `x == MRB_INT_MIN &&  y == -1`.

And if I remember correctly, `Fixnum#divmod` does integer division in mruby (if possible). Yeah, it's longer and unpleasant but in my opinion so is the fixnum/float situation in mruby.

mame:
@cremno, thank you for the comment!

> Your change doesn't care about the integer division issue: y == 0 or x == MRB_INT_MIN && y 
== -1.

Good point, I think I could fix.  BTW, you've find [`Fixnum#divmod`'s possible bug](https://github.com/mruby/mruby/issues/3124) too :-)

> Yeah, it's longer and unpleasant but in my opinion so is the fixnum/float situation in mrub
y.

My hope is compatibility.  I'd like to run an existing program written for MRI.  I don't hasitate to write a shim, but it is impossible as mruby does not allow redefining `Fixnum#/`.
Since few people may want to redefine `Fixnum#/` dynamically, and since a dynamic check will cause extra overhead, build configuration is the mruby's way.  IMO.

bouk:
The Ruby ISO spec actually says that integer division should always have this behavior

> 15.2.8.3.6 Integer#/
> a) If other is an instance of the class Integer:
> 1) If the value of other is 0, raise a direct instance of the class ZeroDivisionError.
> 2) Otherwise, let n be the value of the receiver divided by the value of other. Return
an instance of the class Integer whose value is the largest integer smaller than or
equal to n.
> NOTE The behavior is the same even if the receiver has a negative value. For example,
-5 / 2 returns -3.

Issue #3130, opened 2016-03-06
bintest doesn't work on Visual Studio toolchain


bggd:
minimal build_config.rb

```
MRuby::Build.new do |conf|
  if ENV['VisualStudioVersion'] || ENV['VSINSTALLDIR']
    toolchain :visualcpp
  else
    toolchain :gcc
  end

  enable_debug
  conf.enable_bintest
  conf.enable_test

  conf.gembox 'default'
end
```

On mingw-w64's CLI (toolchain :gcc)

```
.............
Total: 13
   OK: 13
   KO: 0
Crash: 0
 Time: 0.68 seconds
```

On Visual Studio's CLI (toolchain :visualcpp)

```
Total: 0
   OK: 0
   KO: 0
Crash: 0
 Time: 0.0 seconds
```

Should I add ```conf.file_separator = '/'``` on ```:visualcpp``` toolchain?

matz:
If it works, please send us a pull request.

bggd:
bintest worked when ```conf.file_separator = '/'``` instead of ```:visualcpp```'s default ```conf.file_separator = '\\'```.
Is it ideal use case for ```conf.file_separator``` with ```:visualcpp``` toolchain?

matz:
working is better than perfect. but if you want to improve toolchain, it's OK.

zzak:
@bggd I'm glad you got mruby to build with Visual Studio! Would you be able to teach me how? I'd like to help with future testing/debugging of mruby bugs on Windows.

Thanks!

bggd:
[test/bintest.rb](https://github.com/mruby/mruby/blob/41d6ba23b512147497589efbe87a445256d5b213/test/bintest.rb#L23) call Dir.glob with a paths from [Mruby::Build#run_bintest](https://github.com/mruby/mruby/blob/41d6ba23b512147497589efbe87a445256d5b213/tasks/mruby_build.rake#L276).
These paths's separator is converted ```'/'``` to ```'\\'```(backslash) by [Mruby::Build#filename](https://github.com/mruby/mruby/blob/41d6ba23b512147497589efbe87a445256d5b213/tasks/mruby_build.rake#L221) when ```:visualcpp``` toolchain.
Dir.glob doesn't work this case.
This patch (#3131) reconvert ```'\\'``` to ```'/'``` for windows platform with Visual Studio.

matz:
May I close this issue?

bggd:
I forgot to close.

PR #3131, opened 2016-03-07
Fix doesn't work bintest on :visualcpp toolchain

bggd:
```:visualcpp``` toolchain set ```conf.file_separator = '\\'``` by default.
Dir.glob return empty array by backslash separator.


bggd:
#3130

PR #3132, opened 2016-03-07
Added example for cross-compiling on Intel Edison

pbosetti:


zzak:
@pbosetti How about wrapping this into a toolchain?

PR #3139, opened 2016-03-22
deprecate and fix VS2010/2012 build

cremno:


jmhmccr:
Should this be in 1.3 notes? #3140

Issue #3149, opened 2016-04-23
mrb_int_mul_overflow is missing


Asmod4n:
Currently multiplications aren't checked for overflowing, there is __builtin_mul_overflow in GCC/Clang etc. I just don't know how to implement it in compilers which lack it.

zzak:
Is it available in C99?

Asmod4n:
It would be mostly the same as https://github.com/mruby/mruby/blob/master/include/mruby/numeric.h#L40, so i guess its not in C99 too.

zzak:
I'm not familiar enough with the builtin overflows, maybe matz can advise

cremno:
I'll add it. The difficult part here is the fallback for compilers without these built-ins. It needs to be portable and fast. Of course, only to some degree. With 16/32 b `mrb_int` it's actually quite simple: 32/64 b multiplication + `FIXABLE()` (which btw. should be renamed to `mrb_fixable` or even `mrb_fixable_p`). But that's not an option with the 64b one…

PR #3154, opened 2016-05-09
improve checked mrb int arithmetic implementation

cremno:
Somewhat related to #3149 which will be fixed by a future PR.


cremno:
Oh and if somebody here has a better name than the long `MRB_HAVE_TYPE_GENERIC_CHECKED_ARITHMETIC_BUILTINS` - let's hear it!

PR #3159, opened 2016-05-14
disable d098d82 for shortcut operator temporarily

mame:
d098d82 introduces a stack inconsistency when it is a shortcut operator.

  a = { :foo => { :bar => 0 } }
  p((a[:foo] ||= {})[:bar] ||= 42)
    #=> expected: 0, actual: nil

Note that this is just a temporary fix to disable d098d82; `a` in `a[b] ||= c` is evaluated twice.


matz:
29d3a40 causes TravisCI fail. need to fix.

PR #3165, opened 2016-06-25
buffer overflow fix buf is sized 24 (ary member of RString)

devnexen:


cremno:
Can you post Ruby code and if necessary any other steps that are needed to reproduce the bug?

Also a string returned by `mrb_str_buf_new()` never uses the embedded array to store its content. So your fix seems incorrect to me.

PR #3174, opened 2016-07-14
Should raise LocalJumpError when no block given

### LINKED ISSUE #3173: raise NoMethodError when call yield without block on method
I think, this case should raise the `LocalJumpError`.

t.rb

```rb
def a
  yield
end
a
```

### Actual

```
$ mruby t.rb
trace:
	[0] t.rb:2:in Object.a
	[1] t.rb:4
t.rb:2: undefined method 'call' for nil (NoMethodError)
```

### Expect

```
$ mruby t.rb
trace:
	[1] t.rb:2:in Object.a
	[0] t.rb:4
LocalJumpError: unexpected yield
```

### Env

```
$ git rev-parse HEAD
d4d807b774de0e776d2749acbcb22a1b38f1c50b

$ uname -v
Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64

$ clang --version
Apple LLVM version 7.3.0 (clang-703.0.31)
Target: x86_64-apple-darwin15.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Thanks!

ksss:
Fix #3173

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1732,6 +1732,10 @@ RETRY_TRY_BLOCK:
         }
         stack = e->stack + 1;
       }
+      if (mrb_nil_p(stack[m1+r+m2])) {
+        localjump_error(mrb, LOCALJUMP_ERROR_YIELD);
+        goto L_RAISE;
+      }
       regs[a] = stack[m1+r+m2];
       NEXT;
     }
--- a/test/t/syntax.rb
+++ b/test/t/syntax.rb
@@ -38,6 +38,13 @@ assert('yield', '11.3.5') do
   assert_raise LocalJumpError do
     yield
   end
+  assert_raise LocalJumpError do
+    o = Object.new
+    def o.foo
+      yield
+    end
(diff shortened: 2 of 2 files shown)


test/t/localjumperror.rb
```diff
@@ -3,6 +3,13 @@
 
 assert('LocalJumpError', '15.2.25') do
   assert_equal Class, LocalJumpError.class
+  assert_raise LocalJumpError do
```
yui-knk: I think ISO 11.3.5 defines when LocalJumpError should be raised.
15.2.25 defined which methods LocalJumpError should have.

ksss: Thank you for reviewing.
I've moved to `yield [11.3.5]` in `test/t/syntax.rb`.

Issue #3177, opened 2016-07-22
Segmentation fault on Struct


ksss:
This code failed.

```rb
Struct.new(:foo) {
  def initialize(*)
    self.foo
  end
}.new
```

```
$ lldb mruby t.rb
(lldb) target create "mruby"
Current executable set to 'mruby' (x86_64).
(lldb) settings set -- target.run-args  "t.rb"
(lldb) r
Process 52549 launched: '/Users/yuki/bin/mruby' (x86_64)
Process 52549 stopped
* thread #1: tid = 0xcfdafb, 0x0000000100050527 mruby`mrb_struct_ref0(mrb=0x00000001002029f0, obj=mrb_value @ 0x00007fff5fbfd700) + 23 at struct.c:128, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x0000000100050527 mruby`mrb_struct_ref0(mrb=0x00000001002029f0, obj=mrb_value @ 0x00007fff5fbfd700) + 23 at struct.c:128
   125 	  return mrb_struct_getmember(mrb, obj, mrb->c->ci->mid);
   126 	}
   127
-> 128 	static mrb_value mrb_struct_ref0(mrb_state* mrb, mrb_value obj) {return RSTRUCT_PTR(obj)[0];}
   129 	static mrb_value mrb_struct_ref1(mrb_state* mrb, mrb_value obj) {return RSTRUCT_PTR(obj)[1];}
   130 	static mrb_value mrb_struct_ref2(mrb_state* mrb, mrb_value obj) {return RSTRUCT_PTR(obj)[2];}
   131 	static mrb_value mrb_struct_ref3(mrb_state* mrb, mrb_value obj) {return RSTRUCT_PTR(obj)[3];}
```

PR #3178, opened 2016-07-22
Fix segmentation fault

ksss:
Fix https://github.com/mruby/mruby/issues/3177

But, This PR have some bad points.

- `mrb_instance_alloc` copied by `src/class.c`
- `mrb_struct_instance_new` nearly the same as `mrb_instance_new`

How do you think?


ksss:
Test code dependent on https://github.com/mruby/mruby/pull/3179

PR #3181, opened 2016-07-23
Fix crash in mrb_parser_free() due to freed memory access

sdottaka:


matz:
Do you have any info to reproduce the issue?

matz:
I found the issue by using `valgrind`.

sdottaka:
Thank you for the merge.
It was always reproducible when executing mrbc.exe which was built with MSVC2015.
Sorry for no information.

PR #3183, opened 2016-07-23
yard docs for string.h

Mav7:
Please review this documentation for string.h


matz:
Looks reasonable to me.

PR #3184, opened 2016-07-26
Support to call without block to String#each_char

ksss:
`String#char` and `String#each_char` aren't same methods.

This patch support like this (same as CRuby).

```rb
"foo".each_char.with_index do |ch, index|
  # ...
end
```

### DIFF
--- a/mrbgems/mruby-string-ext/mrblib/string.rb
+++ b/mrbgems/mruby-string-ext/mrblib/string.rb
@@ -362,7 +362,15 @@ class String
       self.split('')
     end
   end
-  alias each_char chars
+
+  def each_char(&block)
+    return to_enum :each_char unless block
+
+    split('').map do |i|
+      block.call(i)
+    end
+    self
+  end
 
   def codepoints(&block)
     len = self.size


ebertolazzi:
This patch introduce the following error:

Before

mirb
> a = "hello"
 => "hello"
> a.upcase.each_char.join(' ')
 => "H E L L O"

now

mirb
> a = "hello"
 => "hello"
> a.upcase.each_char.join(' ')
(mirb):4: undefined method 'join' for #<Enumerator: HELLO:each_char> (NoMethodError)

ksss:
@ebertolazzi Thank you for reporting.
I'm sorry break compatibility.
But, This fix follow to CRuby's behavior.

http://ruby-doc.org/core-2.3.0/String.html#method-i-chars
http://ruby-doc.org/core-2.3.0/String.html#method-i-each_char

```rb
$ irb
irb(main):001:0> "hello".upcase.each_char.join(' ')
NoMethodError: undefined method `join' for #<Enumerator: "HELLO":each_char>
```

In this case, I recommend that you use `chars` or `each_chars.to_a` like this.

```rb
$ mirb
> "hello".upcase.chars.join(' ')
 => "H E L L O"
> "hello".upcase.each_char.to_a.join(' ')
 => "H E L L O"
```

Thanks!

ebertolazzi:
Now its clear, thanks for explanation.
Another possibility is to use 

chars.join(' ')

shorthand for seach_char.to_a.join(' ')

regards,

ksss:
:sparkles:

PR #3186, opened 2016-07-27
Update mirb.c

rmalizia44:


matz:
Next time, we want to see more descriptive message than `Update mirb.c`,
for example, `fixed a typo in comment`.

Issue #3140, opened 2016-03-23
1.3.0 Changes


zzak:
Let's use this ticket as a catch-all for features/changes in the upcoming release of MRuby.

It will help me, or whoever, writing the release announcement ;)

If you see a commit or pull request get merged which can be considered a feature, just comment a reference to this ticket!

Thank you!!

jmhmccr:
Looked through commits for some end user visible changes (was curious myself) and the below stood out as summarizable takeaways.

Enumerable#take, #cycle and #first improvements #3293 #3271 #3246

Hash#[] should always call #default #3272
Copy default_proc by Hash#dup #3328

Multiple C++ and Visual Studio compatibility improvements.

matz:
@jmhmccr Thank you for aggregating entry candidates here. It will help to edit the change entries later.

felixjones:
Any changes to the C API should probably be included.

I'd like to see the fixes to the supplied Android NDK toolchain mentioned as I imagine mobile platforms to be a good point of promotion for mruby's potential application. Of course, I am biased on this one.

dabroz:
#3251 is also a new C API feature.

beoran:
Sorry if this the wrong place to ask, but around what time could we expect 1.3.0 to come out? I 'd like to know for a certain project where I can only use "official" release versions. Is anything still blocking the next release?

matz:
Ah, we still don't know. I wanted to implement keyword arguments before 1.3 but have been busy fixing other bugs. Options are
* wait until keyword arguments implementation
* postpone keyword arguments to 1.4. Release it after fixing several remaining bugs

If we choose the latter, we can release it by April/May.

dabroz:
I vote for the latter - 1.3 would be a big step forward even without keyword arguments.

BTW - since we are using GitHub to manage development, we could also benefit from milestones and labels. It would be clear what is the scope of each release.

beoran:
I would also appreciate it if we could have a 1.3 release this spring. It's
been a long time since 1.2 and I eagerly await all the fixes already in
development now. For me, keyword arguments can wait until 1.4. The Ruby ISO
standard doesn't mention those anyway so I am not able to use those for my
work project.

felixjones:
As much as I'd love key-word arguments to make an appearance, the version number is not going to speed that up.

It seems a number of people are relying on a clear cut version that will hard define that no active changes will occur, so with this knowledge, perhaps a named 1.3 release is required.

[...]

PR #3189, opened 2016-08-01
make mrb_hash_values() a public API function

wrl:


cremno:
Why isn't `mrb_funcall(mrb, hash, "values", 0)` good enough for you? Speed?

wrl:
Ah, I suppose I could do that. I just figured that since `mrb_hash_keys()` was a public API function, it would make sense to have values as a public function as well.

jmhmccr:
Should this be in 1.3 notes? #3140

Issue #3208, opened 2016-09-14
android.rake toolchain outdated


felixjones:
The android.rake toolchain needs updating.

Issues with the current version;
- Does not work for Windows x64 NDK as it searches for "windows" rather than "windowsx-86_64" in the NDK selected toolchain's prebuilts
- No real clang support (can only target a single architecture, which is unspecified)
- Default compiler is still GCC, Android NDK now defaults to clang
- Defaults to Android platform 'android-14', even if it's not installed
- NDK no longer uses llvm-ar (GCC's is used), causing clang to fail as the executable no longer exists

I've seen discussions that the [build system may be replaced](https://github.com/mruby/mruby/issues/3113), is this still the case? Is it worth updating the current provided toolchains?

Is it worth removing the android.rake if keeping it up to date with the NDK is too much work?

Not a desperate issue for myself as I have already written a new android.rake for my own Android projects, but having the build system not work for the current NDK platform is quite a discouragement.

matz:
I still want to replace the build system, but currently we have no concrete plan.
So please just update the Android rules.

felixjones:
With mips support being added https://github.com/mruby/mruby/pull/3222 this issue is "resolved" as the Android rules are now updated, however Android GCC support would still be a 'nice thing to have' until GCC is removed from the Android platform.

https://android.googlesource.com/platform/ndk.git/+/master/CHANGELOG.md
GCC for Android is currently deprecated, with its removal being likely once libc++ matures; 

The lost Android GCC support should be its own issue at this point (the goal of updating Android's rules so they are no-longer outdated is met).

PR #3209, opened 2016-09-15
Re-written android.rake task for latest Android NDK clang. Currently …

felixjones:
…loses support for GCC and mips, mips64. Addresses issue #3208


matz:
Thank you for the contribution.

PR #3213, opened 2016-09-21
Visual C++ support of UTF-8 hack on mruby-print

unak:


mrbgems/mruby-print/src/print.c
```diff
@@ -13,7 +13,7 @@ printstr(mrb_state *mrb, mrb_value obj)
 {
   if (mrb_string_p(obj)) {
 #if defined(_WIN32)
-    if (isatty(fileno(stdout))) {
+    if (_isatty(_fileno(stdout))) {
```
mattn: put

unak: I'm not familiar with mingw, but underbar'ed names are real names in MSVCRT, aren't them?
Then I fell we should use them when using them does not make problem with other platforms.

mattn: I don't have strong opinion. However, I prefer to use non-underbar-ed name APIs even though the codes runs only on the Windows (if it's possible).

unak: OK.  I don't have strong opinion, too.
Please wait a moment.

PR #3215, opened 2016-09-23
Add missing MRB_API

kou:
MRB_API is used implementation in load.c:

    MRB_API mrb_irep*
    mrb_read_irep_file(mrb_state *mrb, FILE* fp)


matz:
I considered mrb_read_irep_file() to be an internal function. Do you have any use-case of the function?

kou:
I don't have any use-case. I just want to build (link) error.
Could you remove `MRB_API` in load.c?

kou:
I just want to "fix" build (link) error...

matz:
I see inconsistency.  Will fix.

kou:
Thanks. But I found more same cases. I created a new pull request as #3216.

PR #3216, opened 2016-09-25
Remove needless MRB_API

kou:
ref #3215

If a function (such as mrb_read_irep_file()) is declared without MRB_API
in header file (such as include/mruby/dump.h), implementation of the
function in source file (such as src/load.c) should also defined without
MRB_API.

If MRB_API is mismatch, Visual C++ reports link error with C2375 error
code: https://msdn.microsoft.com/en-us/library/5k6kw95a.aspx

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -646,7 +646,7 @@ mrb_load_irep(mrb_state *mrb, const uint8_t *bin)
 
 #ifndef MRB_DISABLE_STDIO
 
-MRB_API mrb_irep*
+mrb_irep*
 mrb_read_irep_file(mrb_state *mrb, FILE* fp)
 {
   mrb_irep *irep = NULL;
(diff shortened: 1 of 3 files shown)


matz:
merged. But `mrb_str_strlen()` should be `MRB_API`. I will fix.

kou:
Thanks.
BTW, how about removing all `MRB_API` from `**/*.c`? They are optional. If we have `MRB_API` in `**/*.h`, we don't need to put `MRB_API` into `**/*.c`. If we only have `MRB_API` in `**/*.h`, we don't need to care about this problem.

zzak:
:+1: We shouldn't need extern only for declaration, not definition. But, I wonder if changing them all at once will have side-effects :thought_balloon:

PR #3222, opened 2016-10-07
Android mips/mips64 support

felixjones:
Added macOS NDK search paths.

The Android SDK now has NDK installation built in, so the task will now favour that if no ndk_home is set.
~/Library/Android is the default for macOS.

The gcc toolchain path has quotes now, this is a minor change but it fixes the possibility of spaces in that path - keeps it consistent with the use of quotes for other paths in the task.


felixjones:
This pull request goes towards resolving issue https://github.com/mruby/mruby/issues/3208 with  the fixed mips/mips64 support

PR #3229, opened 2016-10-23
Make `class` & `module` return a symbol

franckverrot:
This commit makes the behavior of `class` and `module` consistent with how `def`
is currently working:

    class A end      # => :A
    class B < A; end # => :B
    module M end     # => :M


matz:
Interesting idea, but it's inconsistent with CRuby, so we have to reject now.
We had a similar proposal for CRuby in the past, but we rejected we could not define appropriate behavior for the return value from `class Foo::Bar; end`.

PR #3234, opened 2016-11-06
Add mrb_class_defined_under

felixjones:
Works the same way as mrb_class_defined but uses mrb_const_defined_at with a given outer class to test if an inner class is defined.

Example code for checking if an inner class of a module is defined:
```c
void
mrb_example_gem_init(mrb_state* mrb) {
  struct RClass *example_outer, *example_inner;
  mrb_bool cd;

  example_outer = mrb_define_module(mrb, "ExampleOuter");

  example_inner = mrb_define_class_under(mrb, example_outer, "ExampleInner", mrb->object_class);
  cd = mrb_class_under_defined(mrb, example_outer, "ExampleInner");

  // If mrb_class_under_defined returns 1 then puts "True"
  // If mrb_class_under_defined returns 0 then puts "False"
  if (cd == 1){
    puts("True");
  }
  else {
    puts("False");
  }
}
```

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -567,6 +567,37 @@ MRB_API mrb_bool mrb_class_defined(mrb_state *mrb, const char *name);
 MRB_API struct RClass * mrb_class_get(mrb_state *mrb, const char *name);
 
 /**
+ * Returns an mrb_bool. True if inner class was defined, and false if the inner class was not defined.
+ *
+ * Example:
+ *     void
+ *     mrb_example_gem_init(mrb_state* mrb) {
+ *       struct RClass *example_outer, *example_inner;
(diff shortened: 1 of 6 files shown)


felixjones:
One point to talk about is if this should be "mrb_class_under_defined" or "mrb_class_defined_under". The latter (which is not used in this PR) makes more sense when spoken aloud as "if mrb_class_defined_under", however the former treats classes defined under other classes as their own type as "class_under".

matz:
I'd rather name it `mrb_class_defined_under()` according to CRuby naming convention.
Shall I modify your PR, or will you do it for yourself?

felixjones:
I'll go ahead and do it

PR #3251, opened 2016-11-16
Inline structure data type for mruby (MRB_TT_INLINE)

### LINKED ISSUE #3237: Intrusive storage for small structures?
I am writing a game development oriented mruby gem for small vectors (primarily 4x float). I need at most 16 bytes per object to store its data.

I can use `mrb_data_type` to store this data, and this works fine. However I would like to avoid the extra allocation and store this data along the object if possible. Can I do it somehow? Would I need a new `MRB_TT_XXX` to correctly garbage collect such object?

dabroz:
New data type `MRB_TT_INLINE` is introduced.

Inline structures have no instance variables, no finalizer, and offer as much space as possible in RBASIC object. This means **24** bytes on 64-bit platforms and **12** bytes on 32-bit platforms. Inline structures are always copied by value. 

Only additions to core are 3 simple inline methods in `inline.h`. No extra classes are introduced. No C and/or Ruby gem code is added.

`mruby-inline-struct` gem is only provided for testing.

Fixes #3237.

### DIFF
--- a/include/mruby/inline.h
+++ b/include/mruby/inline.h
@@ -1,0 +1,47 @@
+/*
+** mruby/inline.h - Inline structures
+**
+** See Copyright Notice in mruby.h
+*/
+
+#ifndef MRUBY_INLINE_H
+#define MRUBY_INLINE_H
+
+#include "common.h"
+#include <string.h>
+
+/**
+ * Inline structures that fit in RVALUE
+ *
+ * They cannot have finalizer, and cannot have instance variables.
+ */
+MRB_BEGIN_DECL
+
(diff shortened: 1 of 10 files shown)


matz:
Nice idea&implementation. I will merge this after small modifications.

matz:
merged. thank you!

dabroz:
Great! Thanks for including these changes so quickly.

I also measured performance boost in `mruby-float4`. In 64-bit mode (in 32-bit the space is too small to use), it executes `bm_ao_render_float4` benchmark 15% faster using istruct!

matz:
nice.

dabroz:
cc #3140

PR #3256, opened 2016-11-18
Hash object_id to prevent programs from knowing pointers

bouk:
While running the [Shopify bug bounty for mruby](https://hackerone.com/shopify-scripts) we noticed that a lot of exploits utilize the fact that `object_id` is directly related to the memory address of an object, which makes it very easy to construct arbitrary objects in memory.

This PR fixes the issue by applying a randomly seeded hash to object_id, resulting in object_ids being nondeterministic.

## Results

Running this file after compiling with `MRB_INT64` and `MRB_WORD_BOXING`

```
# id.rb
puts Object.new.object_id
puts Object.new.object_id
puts Object.new.object_id
```

### Before this PR

```
~/src/github.com/mruby/mruby (master) bin/mruby ./id.rb
140338777833864
140338777833576
140338777833000
~/src/github.com/mruby/mruby (master) bin/mruby ./id.rb
140526011564424
140526011564136
140526011563560
~/src/github.com/mruby/mruby (master) bin/mruby ./id.rb
140254254219656
140254254219368
140254254218792
```

### After this PR

```
~/src/github.com/mruby/mruby (hash-object-id) bin/mruby ./id.rb
-4501018806046184518
-2341758151096772472
-785186630034095801
~/src/github.com/mruby/mruby (hash-object-id) bin/mruby ./id.rb
552799609413634798
3966797781753601314
4531261121971834197
~/src/github.com/mruby/mruby (hash-object-id) bin/mruby ./id.rb
-962803763787832197
2853411265111984547
2087943745143658397
```

## Choice of hash code

I decided to go with the [libsodium](https://github.com/jedisct1/libsodium) implementations of siphash and cryptographically secure bytes, because `libsodium` is very widely used, and the code is released under [ISC](https://opensource.org/licenses/ISC), which is compatible with MIT.

For review @fbogsany @matz

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -184,6 +184,8 @@ typedef struct mrb_state {
   struct RClass *eStandardError_class;
   struct RObject *nomem_err;              /* pre-allocated NoMemoryError */
 
+  unsigned char key[16]; /* siphash key */
+
   void *ud; /* auxiliary data */
 
 #ifdef MRB_FIXED_STATE_ATEXIT_STACK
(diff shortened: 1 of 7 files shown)


isra17:
Does it really need to be hashed though? Generating a random seed stored in the mrb_value state and XORing it to an object pointer should be sufficient to avoid easy memory leak. At worst, it might allow an attacker to get hint about an object location relative to another, but should leave many ASLR bits unleakable. Faster and simpler implementation than hashing (With improbable, but possible collision on 32bits)

`Class.new.to_s` and `Proc.new{}.to_s` should also be addressed as they leak heap address as well.

PR #3257, opened 2016-11-19
Fix Range#size results

ksss:
# Actual

```rb
$ mirb

> (1.0..15.9).size
 => 16
> (0..Float::INFINITY).size
 => -2147483648
```

# Expect

```rb
$ irb
irb(main):001:0> (1.0..15.9).size
=> 15
irb(main):002:0> (0..Float::INFINITY).size
=> Infinity
```

This fix passed all specs of that https://github.com/ruby/spec/blob/66daf290cb49707190481b4cf24e700010cd53bf/core/range/size_spec.rb

### DIFF
--- a/mrbgems/mruby-range-ext/src/range.c
+++ b/mrbgems/mruby-range-ext/src/range.c
@@ -139,7 +139,7 @@ mrb_range_size(mrb_state *mrb, mrb_value range)
 {
   struct RRange *r = mrb_range_ptr(range);
   mrb_value beg, end;
-  double beg_f, end_f;
+  mrb_float beg_f, end_f;
   mrb_bool num_p = TRUE;
   mrb_bool excl;
 
@@ -147,7 +147,7 @@ mrb_range_size(mrb_state *mrb, mrb_value range)
   end = r->edges->end;
   excl = r->excl;
   if (mrb_fixnum_p(beg)) {
-    beg_f = (double)mrb_fixnum(beg);
+    beg_f = (mrb_float)mrb_fixnum(beg);
   }
   else if (mrb_float_p(beg)) {
     beg_f = mrb_float(beg);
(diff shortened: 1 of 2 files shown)


mrbgems/mruby-range-ext/src/range.c
```diff
@@ -162,14 +164,27 @@ mrb_range_size(mrb_state *mrb, mrb_value range)
...
-    double f;
-
-    if (beg_f > end_f) return mrb_fixnum_value(0);
-    f = end_f - beg_f;
-    if (!r->excl) {
-      return mrb_fixnum_value((mrb_int)ceil(f + 1));
+#ifndef DBL_EPSILON
+#define DBL_EPSILON 2.2204460492503131e-16
+#endif
+    double n = end_f - beg_f;
```
dabroz: Why not `mrb_float`?

mrbgems/mruby-range-ext/src/range.c
```diff
@@ -162,14 +164,27 @@ mrb_range_size(mrb_state *mrb, mrb_value range)
     num_p = FALSE;
   }
   if (num_p) {
-    double f;
-
-    if (beg_f > end_f) return mrb_fixnum_value(0);
-    f = end_f - beg_f;
-    if (!r->excl) {
-      return mrb_fixnum_value((mrb_int)ceil(f + 1));
+#ifndef DBL_EPSILON
```
dabroz: We already include <float.h> (ANSI C), which includes this constant.

mrbgems/mruby-range-ext/src/range.c
ksss: I did'n know that.
Should we use `mrb_float` instead of `double` and `float` in all cases?

matz:
mrb_float would be better, although I saw a lot of doubles in old code.

ksss:
Thank you, I fixed it!

PR #3261, opened 2016-11-21
Reimplements Range#first

ksss:
`Range#first` with argument works like `Enumerable#take`.
`Range#first` shouldn't call `Range#to_a` on infinity range.

```rb
$ mirb

> (0..Float::INFINITY).first(3)
# ...infinite loop...
```

So, I reimplemented `Range#first` with Ruby code.
Because it hard to write `break` in iterator on C functions.
And this implementation passed specs of that https://github.com/ruby/spec/blob/66daf290cb49707190481b4cf24e700010cd53bf/core/range/first_spec.rb
(I know that it's a bit too strictly)

### DIFF
--- a/mrbgems/mruby-range-ext/mrblib/range.rb
+++ b/mrbgems/mruby-range-ext/mrblib/range.rb
@@ -1,0 +1,31 @@
+class Range
+  ##
+  # call-seq:
+  #    rng.first    -> obj
+  #    rng.first(n) -> an_array
+  #
+  # Returns the first object in the range, or an array of the first +n+
+  # elements.
+  #
+  #   (10..20).first     #=> 10
+  #   (10..20).first(3)  #=> [10, 11, 12]
+  #
+  def first(*args)
+    return self.begin if args.empty?
+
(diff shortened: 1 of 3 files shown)


mrbgems/mruby-range-ext/mrblib/range.rb
```diff
@@ -0,0 +1,31 @@
...
+  #
+  #   (10..20).first     #=> 10
+  #   (10..20).first(3)  #=> [10, 11, 12]
+  #
+  def first(*args)
+    return self.begin if args.empty?
+
+    raise ArgumentError, "wrong number of arguments (given #{args.length}, expected 1)" unless args.length == 1
+    nv = args[0]
+    raise TypeError, "no implicit conversion from nil to integer" unless nv.nil?.!
```
dabroz: We could simplify that to `if nv.nil?`

ksss: I just wanted to unify `unless` with `raise`.

PR #3263, opened 2016-11-21
Use xor with secret value to obfuscate object_id

bouk:
As an alternative to https://github.com/mruby/mruby/pull/3256 we can just hash the object_id with a randomly generated `mrb_int`, thanks for the suggestion @isra17

We will still need to fix `mrb_ptr_to_str` to not leak the pointer however


dabroz:
How would this work on embedded systems without `/dev/` filesystem?

bouk:
This wouldn't work on embedded systems, no

Asmod4n:
Libsodium works just fine without a /dev filesystem if the kernel is new enough

PR #3258, opened 2016-11-20
Travis tests for multiple configurations

dabroz:
mruby has many configuration options and I wanted to check how these configurations behave. So I've added support for multiple configurations in Travis testing, including:
- 32/64 bit architecture
- float/double
- 16/32/64 bit int size
- none/NaN/word boxing
- GCC/clang
- Linux/OSX

with unsupported configurations like `float` + NaN boxing removed from the pool.

The results are not so good. Most non-standard configurations crash or fail tests. Of all 96 possible configurations, only 13 build and test correctly.

I'll try to submit patches where possible, and create issues for more complex problems.

Issue #2832, opened 2015-06-12
On ruby error, "terminate called after throwing an instance of 'int'" when adding an mrbgem with C++ code on Raspberry


pbosetti:
I found a strange behavior when building mruby on Raspberry with mrbgems having C++ source code (there are several of them on github, e.g. `take-cheeze/mruby-marshal`)
In this case, any error that would normally call a ruby exception crashes the executable (either mirb or mruby) with the message `terminate called after throwing an instance of 'int'`.
This can be replicated even with an empty gem written in C++, and propagates to every class of the interpreter (e.g. evaluating `Object.not_a_method` crashes rather than raising `NoMethodError`), so I don't think it is related to the named gem, but rather to the build process on Raspberry.

By the way, I tried to set `conf.enable_cxx_abi` without any difference.

I am building with GCC 4.6.3-14+rpi1.

Any idea about how to fix it?

take-cheeze:
According to this [blog(sorry to paste Japanese)](http://d.hatena.ne.jp/yupo5656/20051215/p1) adding `-fexceptions` flag seems to solve this.

cremno:
Shouldn't it be added to the default CXXFLAGS? See the MSVC toolchain file (`/EHs` is the equivalent).

take-cheeze:
Well seems like it's a flag for C to support C++ exceptions.

cremno:
I see. I should have read the description:
>If you do not specify this option, GCC enables it by default for languages like C++ that normally require exception handling, and disables it for languages like C that do not normally require it. However, you may need to enable this option when compiling C code that needs to interoperate properly with exception handlers written in C++.

matz:
Closing — C++ "throwing int" terminate, raspberry-pi/old-mruby specific, no follow-up since 2015. `MRB_TRY`/`MRB_CATCH` machinery has been reworked since; please re-open with a current reproducer if you still hit this.

PR #3267, opened 2016-11-23
Compiling mruby as C++ code

dabroz:
This is an artifact discovered during #3258. 

mruby offers a way to use C++ exceptions instead of C jumps. This is a great idea, especially given that C++ exceptions cost nothing on modern compilers.

However, this is made in such a way that most of mruby code is compiled as C, and only some special files (`vm.c`, `error.c`, etc.) are complied as C++, while still being defined as "`extern C`".

This is unfortunately an undefined behavior, and many compilers won't generate proper exception propagation code in functions defined as such. That includes both GCC and Clang on different platforms.

The proper solution is to compile all of mruby code as C++. This pull request deals with this issue in a following manner:
- fixed few issues with code that was proper C but not C++ (ie. implicit `void*` casts)
- added macros for C-style definitions like `INT32_MAX`, `SIZE_MAX`, etc.
- added export declarations to MRBC outputs
- changed the method of compiling mruby with C++ ABI to use `-x c++`

The last point should be reworked in the future (hacks for marking some files as C++ would no longer be needed), but it serves well for now, and increases the compatibility of mruby significantly.

### DIFF
--- a/tasks/toolchains/gcc.rake
+++ b/tasks/toolchains/gcc.rake
@@ -6,7 +6,7 @@ MRuby::Toolchain.new(:gcc) do |conf, _params|
     cc.option_include_path = '-I%s'
     cc.option_define = '-D%s'
     cc.compile_options = '%{flags} -MMD -o %{outfile} -c %{infile}'
-    cc.cxx_compile_flag = '-x c++'
+    cc.cxx_compile_flag = '-x c++ -std=c++03'
   end
 
   [conf.cxx].each do |cxx|
@@ -16,7 +16,7 @@ MRuby::Toolchain.new(:gcc) do |conf, _params|
     cxx.option_include_path = '-I%s'
     cxx.option_define = '-D%s'
     cxx.compile_options = '%{flags} -MMD -o %{outfile} -c %{infile}'
-    cxx.cxx_compile_flag = '-x c++'
+    cxx.cxx_compile_flag = '-x c++ -std=c++03'
   end
 
   conf.linker do |linker|
--- a/tasks/toolchains/gcc.rake
+++ b/tasks/toolchains/gcc.rake
@@ -6,6 +6,7 @@ MRuby::Toolchain.new(:gcc) do |conf, _params|
     cc.option_include_path = '-I%s'
     cc.option_define = '-D%s'
     cc.compile_options = '%{flags} -MMD -o %{outfile} -c %{infile}'
+    cc.cxx_compile_flag = '-x c++'
   end
 
   [conf.cxx].each do |cxx|
@@ -15,6 +16,7 @@ MRuby::Toolchain.new(:gcc) do |conf, _params|
     cxx.option_include_path = '-I%s'
     cxx.option_define = '-D%s'
     cxx.compile_options = '%{flags} -MMD -o %{outfile} -c %{infile}'
+    cxx.cxx_compile_flag = '-x c++'
   end
 
   conf.linker do |linker|
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -28,9 +28,25 @@
 #ifndef MRUBY_H
 #define MRUBY_H
 
+#ifdef __cplusplus
+#define __STDC_LIMIT_MACROS
+#define __STDC_CONSTANT_MACROS
+#define __STDC_FORMAT_MACROS
+#endif
+
 #include <stdint.h>
 #include <stddef.h>
 #include <limits.h>
+
+#ifdef __cplusplus
+#ifndef SIZE_MAX
+#ifdef __SIZE_MAX__
+#define SIZE_MAX __SIZE_MAX__
+#else
+#define SIZE_MAX std::numeric_limits<size_t>::max()
+#endif
+#endif
+#endif
 
 #ifdef MRB_DEBUG
 #include <assert.h>
--- a/include/mruby/boxing_nan.h
+++ b/include/mruby/boxing_nan.h
@@ -53,7 +53,7 @@ typedef struct mrb_value {
 #define mrb_float_pool(mrb,f) mrb_float_value(mrb,f)
 
 #define mrb_tt(o)       ((enum mrb_vtype)(((o).value.ttt & 0xfc000)>>14)-1)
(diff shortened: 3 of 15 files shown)


matz:
PR merged. Thank you!

kjunichi:
これ、なんで取り込んだんですか？
既存のC++とCが混在しているmrbgemでリンクエラーになるんですけど。。

Why did you incorporate this?
I get a link error with mrbgem where existing C ++ and C are mixed. .

matz:
Ah, actually I thought it improve C++ linkage. If it makes handling C++ mrbgem worse, I will revert the PR.

kjunichi:
I'd like to use C ++ exceptions conveniently, but I would like you to revert.

That's for the following two reasons.

- The author of mrbgem using existing C ++ needs to modify the declaration of extern "C".
- If you want to put mrbgem using C ++, all other mrbgem written in C also needs to be C ++ compatible code.（As well as the modifications in the C code of the mruby itself in this patch）

dabroz:
You don't need to use extern "C". Mruby has a macro MRB_BEGIN_DECL that adds an appropriate block if needed. Other than that you only need to ensure that the gem's code is also valid C++ - but that is usually simple to do (sometimes need to add type casts). 

Without this patch, mruby is broken on many platforms. 

You can also use mruby in C++ without using C++ exceptions, in such scenario mruby and gems' code would be compiled as C code.

kjunichi:
？

I already made mrbgems .your patch makes me change my code!

tasks/toolchains/gcc.rake
```diff
@@ -6,6 +6,7 @@ MRuby::Toolchain.new(:gcc) do |conf, _params|
     cc.option_include_path = '-I%s'
     cc.option_define = '-D%s'
     cc.compile_options = '%{flags} -MMD -o %{outfile} -c %{infile}'
+    cc.cxx_compile_flag = '-x c++ -std=c++03'
```
zzak: I guess this change makes this flag required now?

Is there some way to do this in a backwards compatible way? i.e. default flag which supports both versions

dabroz:
@kjunichi If you used previous version of mixed C/C++ mruby, and had C++ exceptions cross the language barrier (which is likely if you mix Ruby with C(++)), then you get _undefined behavior_. Some compilers permit it, but other compilers enforce it, and will result in segmentation fault in such case.

I'm not saying that the current handling of C/C++ mode is perfect, and likely should be refactored, but we need to choose a **correct** (in terms of language correctness) solution; even if it would require small code adjustments for gem authors.

kjunichi:
@dabroz 

I already ran several games using multiple C ++ on multiple platforms such as Linux, macOS, windows without crashing. 
From this experience I was unaware of the "undefined behavior" you said here.
This is because I have not written any code using any C ++ exception.

> and had C++ exceptions
Now I understood.

I think we cannot stop using your patch even though there is fix for C only gem.

thanks for taking your time.

matz:
@dabroz could you elaborate the situation this PR fixes? What kind of problems on which platforms?
By seeing the info, we may be able to make a new workaround.

dabroz:
@matz: Sure.

Let's use following program, build with 2 source files. One of them is C, the other one is C++. To be able to use each others functions, C++ functions are declared as `extern "C"`.

This is of course a simplified version of mruby code, where (before this patch) most of the codebase was C, and exception-related parts were C++ with C linkage.

```c++
// cpp_code.cpp
#include <stdio.h>

extern "C" void c_func();

extern "C" void cpp_func()
{
	throw 1;
}

int main()
{
	try
	{
		c_func();
	}
	catch (int exception)
	{
		fprintf(stderr, "Exception correctly caught: %d\n", exception);
	}
}
```

```c
// c_code.c
void cpp_func();

void c_func()
{
	cpp_func();
}
```

Let's build them like this:
```
clang++ ${FLAGS} -c cpp_code.cpp -o cpp_code.o
clang   ${FLAGS} -c c_code.c     -o c_code.o
clang   ${FLAGS} c_code.o cpp_code.o -o test
./test
```

Now, everything depends on `FLAGS`. If we run this with empty `FLAGS`, the compiler will default to 64-bit mode, and the code will run correctly:

```
Exception correctly caught: 1
```

However, if we set `FLAGS` to `-m32`, forcing 32-bit mode, the result will be very different:

```
libc++abi.dylib: terminating with uncaught exception of type int
```

The problem observable here is undefined behavior of C++ exceptions passing C code barrier. It *may* run correctly if the ABI supports is (which is the case under 64-bit OSX), but it doesn't need to (like 32-bit OSX) - and then the code will crash as soon as the exception crosses C/C++ barrier.

I am running these tests on OSX (10.12.2) with Apple LLVM version 8.0.0 (clang-800.0.42.1), but the same behavior is visible under other versions of Clang and also under Linux.

kjunichi:
Hi, @dabroz .

```shell
clang   ${FLAGS} c_code.o cpp_code.o -o test
```

is it a typo?

```shell
clang++  ${FLAGS} c_code.o cpp_code.o -o test
```

is a collect,isn' it?

In this case, you alse can use -fexception option for compiling C code.

```
export FLAGS="-m32"
clang++ ${FLAGS} -c cpp_code.cpp -o cpp_code.o
clang   ${FLAGS} -fexceptions -c c_code.c     -o c_code.o
clang++  ${FLAGS} c_code.o cpp_code.o -o test
./test 
Exception correctly caught: 1
```

(I ran this on macOS.)

I'm not C/C++ expert, please see #2832 .

dabroz:
@kjunichi you're right, linking target binary also needs `clang++`. Fixed.

`-fexceptions` seems to work well on 32-bit Mac targets. However I would recommend to leave compatibility fixes (pointer casts etc.) in place, as being able to compile whole mruby as C++ can also be useful in some scenarios (embedding).

bggd:
Can We add an option like ```enable_c_as_cxx``` or ```force_compile_as_cxx``` or something to mruby's Rake?

matz:
@dabroz could you check if the recent commit (1a1f834) works OK for you?

dabroz:
@matz I'll check this update and see if tests can be updated to test all possible ABIs.

PR #3294, opened 2016-11-25
Put a reference to the range_class in mrb_state

bouk:
Can cause segfaults otherwise

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -191,6 +191,7 @@ typedef struct mrb_state {
   struct RClass *proc_class;
   struct RClass *string_class;
   struct RClass *array_class;
+  struct RClass *range_class;
   struct RClass *hash_class;
 
   struct RClass *float_class;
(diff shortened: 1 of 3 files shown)


include/mruby.h
```diff
@@ -191,6 +191,7 @@ typedef struct mrb_state {
   struct RClass *proc_class;
   struct RClass *string_class;
   struct RClass *array_class;
+  struct RClass *range_class;
```
bouk: @matz does this need to be added to root_scan_phase ?

matz:
I don't want to increase the size of mrb_state if possible. Could you tell us how we can cause SEGV?

bouk:
If you look at the test, you can see that you can do `Range = Array` which will create a `TT_RANGE` with Array class. We could also save the range class as a global variable

matz:
Sorry, I found the error reproducing code in the PR.  I think I found the work around.

PR #3297, opened 2016-11-25
Remove extra edges structure for RRange

bouk:
Fixes initialize issue reported by @charliesome

I'm guessing the reason struct has an extra struct is to make `RVALUE` smaller? If so, might it be a better alternative to implement RRange using RData, so we can have data structure checks everywhere? Currently there's a lot of unchecked access to `r->edges` which can always be null if the range wasn't initialized

### DIFF
--- a/include/mruby/range.h
+++ b/include/mruby/range.h
@@ -14,14 +14,11 @@
  */
 MRB_BEGIN_DECL
 
-typedef struct mrb_range_edges {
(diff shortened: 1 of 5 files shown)


matz:
This pull request is unacceptable because mrb_value*2 + int is bigger than 3 words.
Implementing Ranges using RData may be a good idea.

PR #3309, opened 2016-11-30
Prevent array size calculation overflows.

clayton-shopify:
@matz 

There are a couple other places where an array size calculation can overflow. First, in `ary_modify` where the byte size should be calculated using `size_t` instead of `mrb_int` (much like https://github.com/mruby/mruby/pull/3304). And second, in `ary_concat` where the addition of the two array lengths can overflow.

The following input demonstrates a segfault in `ary_concat`:
```ruby
a = [0]
31.times { a.concat(a) }
```
After this change it exits with `array size too big (ArgumentError)`.

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -118,7 +118,7 @@ ary_modify(mrb_state *mrb, struct RArray *a)
     }
     else {
       mrb_value *ptr, *p;
-      mrb_int len;
+      size_t len;
 
       p = a->ptr;
       len = a->len * sizeof(mrb_value);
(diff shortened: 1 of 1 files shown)


tsahara:
CI fails.  Is compier option `[-Werror=declaration-after-statement]` too strict?

```
CC    src/array.c -> build/debug/src/array.o
/home/travis/build/mruby/mruby/src/array.c: In function ‘ary_concat’:
/home/travis/build/mruby/mruby/src/array.c:250:3: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
cc1: some warnings being treated as errors
```

matz:
I know it's strict, but some compiler that is supported by mruby still complains mixed declaration.

tsahara:
Thanks nobu and matz.  I like traditional syntax better than C99.

PR #3314, opened 2016-12-01
Fix compile error by #3309

nobu:


matz:
Oops, I did exact same fix in ab4ab7c

PR #3319, opened 2016-12-01
Fix segfault when using result of rest assignment

bouk:
Reported by https://hackerone.com/haquaman


bouk:
This would segfault because the stack would get misaligned

PR #3320, opened 2016-12-02
Fix segfault in Range without initialize_copy

clayton-shopify:
@matz

The following code results in a null pointer dereference, since an uninitialized Range object is used:
```ruby
Range.remove_method(:initialize_copy)
(1..2).dup.to_s
```
This patch solves the issue by getting rid of the `mrb_range_edges` struct, so that initialization can be safely skipped.

This issue was reported by https://hackerone.com/charliesome.

### DIFF
--- a/include/mruby/range.h
+++ b/include/mruby/range.h
@@ -14,14 +14,11 @@
  */
 MRB_BEGIN_DECL
 
-typedef struct mrb_range_edges {
-  mrb_value beg;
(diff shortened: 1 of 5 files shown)


matz:
This patch increases the size of RVALUE (object slot) thus increasing memory consumption.
I don't think it's acceptable.  I understand the issue. I will fix it.

PR #3323, opened 2016-12-06
Fix 36fc1f14 not checking in the right location

bouk:
36fc1f14 should've added the exception class check in the mrb_exc_raise function, as the check in mrb_exc_set doesn't work if the value is nil. I also added the test that was originally in https://github.com/mruby/mruby/pull/3292 (which was failing on master)

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -278,8 +278,6 @@ mrb_exc_set(mrb_state *mrb, mrb_value exc)
     mrb->exc = 0;
   }
   else {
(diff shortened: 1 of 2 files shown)


clayton-shopify:
@matz Have you had a chance to look at this?

matz:
The patch seems OK, but I don't know why Appveyor failed.
That's my concern.

clayton-shopify:
@bouk is away, but I'll rebase this and open a new PR to see if it fixes the AppVeyor problem.

matz:
#3394 merged.

Issue #3270, opened 2016-11-23
Floating Point Parser Behavior Is Locale Dependent


fundamental:
Mruby's parse.y uses libc's strtod to parse floating point literals. For some locale configurations (those which use ',' in place of '.' as a decimal separator) this results in incorrectly parsed literals. Cruby appears to work around this with by using their own implementation, ruby_strtod which by default (unless USE_LOCALE is explicitly enabled) ignores locale settings when parsing ruby code.

A similar solution would be preferred here as setlocale cannot be safely used in applications where mruby is embedded within another program (e.g. for the case of plugins which make use of mruby).

matz:
Hmm, difficult problem. The ideal solution must be reimplementing strtod(), but it would increase the size of mruby.

fundamental:
If the resulting binary size is the main concern it should only affect binaries that pull in the mruby-compiler gem, but if the size of the overall codebase is the concern I don't see much of a way to work around that.

dabroz:
`"1.25".to_f` uses `strtod` as well.

dabroz:
Having said that, we could use `strtod_l` (or `_strtod_l`) on systems that are known to have it.

matz:
I didn't know about `strtod_l`. I am sorry that it's not portable.
I found out that the locale insensitive `strtod` is not that big. I will merge it soon.

fundamental:
Thanks for the fix.

PR #3329, opened 2016-12-07
Mark all the built-in classes during GC sweep

bouk:
This fixes an issue where a built-in class can get garbage collected, put onto the freelist, and reused. This can lead to instantiating an object with an MRB_TT that doesn't match its class.


bouk:
I'm unsure where this appveyor issue is coming from

matz:
Appvoyer issue is caused by #3270. I don't know how to fix it yet.

matz:
The patch looks good to me, but I worry a bit that the test may not be portable.

matz:
Ah, but existing tests already uses the same approach.  OK, I'll merge it.

PR #3330, opened 2016-12-07
Support override of + - * / for Fixnum and Float (rebased)

dabroz:
I've rebased #2579 (Support override of + - * / for Fixnum and Float) and added tests. Could we add this to mruby? It would be useful for most numeric gems (like Bignum or float4).

### DIFF
--- a/build_config.rb
+++ b/build_config.rb
@@ -122,6 +122,8 @@ MRuby::Build.new('test') do |conf|
   conf.enable_test
 
   conf.gembox 'default'
+  conf.gem core: 'mruby-inline-struct'
+  conf.gem core: 'mruby-numeric-override'
 end
 
 MRuby::Build.new('bench') do |conf|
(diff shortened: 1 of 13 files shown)


dabroz:
@matz: what do you need to consider this pull request? I am not against making this optional (new flag in `mrbconf.h`) if you think that including this by default is not necessary.

dabroz:
I made some progress on this (configuration and tests), but it seems that I misunderstood the actual mruby limitation. It looks like there always was a possibility to use `Fixnum`/`Float` `+-*/` with custom classes.

The current version has a configuration flag `MRB_ENABLE_NUMERIC_OVERRIDE` (disabled by default). However, there are still limitations coming from the fact that `+` and other operators are not actual methods (you can't use `super` or `alias_method`, for example).

Therefore I am leaving my latest changes here for all of you to decide what the next step should be.

PR #3331, opened 2016-12-07
Promote load_exec to mruby API as mrb_load_exec (fixes #3248)

### LINKED ISSUE #3248: Catch parse errors in C
When reloading Ruby scripts, currently I need to parse them twice to be able to catch parse errors:

```c
mrbc_context* context = mrbc_context_new(mrb);
mrbc_filename(mrb, context, path);
parser_state* parser = mrb_parse_nstring(mrb, data, length, context);
int num_errors = parser->nerr;
mrb_parser_free(parser);
if (num_errors > 0)
{ ... }
mrb_load_nstring_cxt(mrb, data, length, context);
```

all because `mrb_load_nstring_cxt` is defined as

```c
MRB_API mrb_value
mrb_load_nstring_cxt(mrb_state *mrb, const char *s, int len, mrbc_context *c)
{
  return load_exec(mrb, mrb_parse_nstring(mrb, s, len, c), c);
}
```

and `load_exec` in not in the mruby API.

Is there a different way to do it without parsing the file twice, or could we promote `load_exec` to API?

dabroz:
Rationale: you would no longer need to parse Ruby scripts twice to catch parse errors.

Fixes #3248


jmhmccr:

PR #3335, opened 2016-12-08
fix build on vs2013-vs2015

mattn:


mattn:
Hmm, still not. I have no idea. :(

mattn:
I figure out why appveyor fail.

```c
#  define const
```
(in src/string.c)

This is wrong. This break signature for mrb_float_read.

mattn:
@matz Is this required?

```c
#ifndef __STDC__
# ifdef __GNUC__
#  define const __const__
# else
#  define const
# endif
#endif
```

matz:
It was copied from ruby1.8/missing. Probably for compatibility with old compilers.

PR #3336, opened 2016-12-08
Removed unnecessary const macro - const keyword is already a dependency

felixjones:
const is already used at the top of string.c, so we can guarantee that the compiler supports the const keyword at this position of the code.

This macro also does not work for circumstances where \_\_STDC__ is not defined and your compiler has not definition (the default is to disable the const keyword, this does not sound like intentional behaviour).


matz:
You are right.

Issue #2028, opened 2014-04-08
Add Object#freeze


IceDragon200:
Does mruby have Object#freeze by the way, I did a quick search and didn't find anything... (then again, I just skimmed over it)

jemc:
I'd like to see at least `String#freeze` if possible.

Asmod4n:
Also have a need for String#freeze.

Currently writing a "SecureBuffer" class where you can mark a malloced() Region as read only, read and write or deny any access to it.

Of course i want that buffer to be viewable from ruby land so it has a to_str method via mrb_str_new_static().

It would be cool if that String Object wouldn't be modifyable. Or at best, never be copied to ruby land but only be referenceable.

syohex:
@jemc @Asmod4n `String#freeze` was introduced at this commit(1a45447b8b66159a4a7d0348c4a01a175b3778fd)

archSeer:
Doesn't seem like the original issue was fixed here, the fix is limited to strings.

matz:
OK, I reopen the issue.

PR #3340, opened 2016-12-10
Implement Object#freeze

k0kubun:
close https://github.com/mruby/mruby/issues/2028

I want to use `Object#freeze` for Array and Hash to make them read-only. It's convenient to restrict regions in which you can modify a value. It improves maintainability because it makes easier to search where a value is modified.


jmhmccr:

Issue #3356, opened 2016-12-19
Null pointer dereference in mrb_vm_exec


clayton-shopify:
The following input causes a crash:
```ruby
Proc.remove_method(:initialize)

class A
  define_method(:b, Proc.new)
end

A.new.b
```
A null pointer dereference occurs here: https://github.com/mruby/mruby/blob/432570723d027313f0d8dd5d3724fbc53f160e5c/src/vm.c#L1199

This issue was reported by https://hackerone.com/ston3

matz:
#3362 merged.

PR #3362, opened 2016-12-21
Proc shouldn't have `initialize` method

ksss:
CRuby's `Proc` class does't have `initialize` method.
And It raises some issues when it called `Proc.remove_method(:initialize)`

I think, `Proc` class shouldn't have `initialize` method in mruby too.

ref: #3356


jmhmccr:

PR #3369, opened 2017-01-04
Rewrite String#prepend with Ruby

### LINKED ISSUE #3357: Use after free
The following input causes a crash:
```ruby
def method_missing(m)
  "0000000"
end

m0 = w.instance_eval { prepend(m) }
m0.instance_eval { prepend(m0) }
```
Adjusting the length of the string returned by `method_missing` may also change the way the crash occurs; the submitter also sent this PoC, which for me crashes only in mirb:
```ruby
def method_missing(m)
  "111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111"
end

m0 = w.instance_eval { prepend(m) }
m0.instance_eval { prepend(m0) }
```

This issue was reported by https://hackerone.com/ston3

ksss:
`String#prepend` have unexpected behavior.

```rb
s = "abcdefghijkl"
p s.prepend(s)
#=> "abcdefghijkl\000\000\000\360\002\000\000\000\000\000\000\360"
```

I think it will be able to write more simply.

And fix #3357

### DIFF
--- a/mrbgems/mruby-string-ext/mrblib/string.rb
+++ b/mrbgems/mruby-string-ext/mrblib/string.rb
@@ -385,4 +385,18 @@ class String
     end
   end
   alias each_codepoint codepoints
+
+  ##
+  # call-seq:
+  #    str.prepend(other_str)  -> str
+  #
+  # Prepend---Prepend the given string to <i>str</i>.
+  #
+  #    a = "world"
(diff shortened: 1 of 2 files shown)


ksss:
I think, `String#[]=` should raise an error `TypeError` when argument is none `String` object.

PR #3394, opened 2017-01-11
Fix 36fc1f14 not checking in the right location

clayton-shopify:
@matz 

This is a rebase of https://github.com/mruby/mruby/pull/3323 to see if the AppVeyor problem goes away.


clayton-shopify:
Looks like everything is good now.

Issue #3448, opened 2017-02-10
String#[]= can break UTF-8 length counting


dabroz:
At commit c802cd07baf7132c5053defac883f0ee6b7967b7

When mruby is compiled with `MRB_UTF8_STRING`, following code gives weird results:

    a = ' ' * 34
    a[-6..-1] = 'foobór'
    a.length # is 35, should be 34
    a.split('').length # is 34

PR #3449, opened 2017-02-10
String#ljust and String#rjust reimplemented with optimized Ruby

dabroz:
Unfortunately, #3447 breaks `ljust`/`rjust` for UTF-8 strings.

I've tried fixing this in C, but managing UTF-8 strings in C is quite error-prone. So, I've tried to re-implement them again in Ruby, but in a more optimized way.

I've also added a small optimization to `String#*`.

Without `MRB_UTF8_STRING`:

    $ echo '"".ljust(10000000);"".rjust(10000000)' | time bin/mruby
            0.04 real         0.02 user         0.01 sys

With `MRB_UTF8_STRING`:

    $ echo '"".ljust(10000000);"".rjust(10000000)' | time bin/mruby
            0.51 real         0.48 user         0.02 sys

Also, #3448 was discovered while working on this fix.

### DIFF
--- a/mrbgems/mruby-string-ext/mrblib/string.rb
+++ b/mrbgems/mruby-string-ext/mrblib/string.rb
@@ -263,6 +263,44 @@ class String
     self
   end
 
+  ##
+  #  call-seq:
+  #     str.ljust(integer, padstr=' ')   -> new_str
+  #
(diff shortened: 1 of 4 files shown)


r-obert:
very good :+1: 
the pure version is almost as fast as C, and much faster than the original version written in Ruby.

PR #3453, opened 2017-02-13
Separate compilers setup from config setup

### LINKED ISSUE #3418: Build with "enable_test" fails due to broken "include_paths"
Hi, everyone!

I have a gem tree like this:

```ruby
root
 |
 A –
      |
      B
      |
      C
```

Where `A` is using headers from `B` (say, `include <b/functions.h>`).

But when I'm running a build with `conf.enable_test` (i.e. adding `mruby-test` to the build pipeline), I've got the following error:

```
fatal error: 'b/functions.h' file not found
```

As it turned out, adding `mruby-test` initiates additional compiling task (one for each gem) with broken configuration – unpopulated `include_paths`.

Why does this happen?

- `mruby-test` [generates a gem_table](https://github.com/mruby/mruby/blob/master/mrbgems/mruby-test/mrbgem.rake#L36)

- That [involves a call](https://github.com/mruby/mruby/blob/master/tasks/mrbgem_spec.rake#L346)  to `Spec#setup`

- `Spec#setup` defines [compilers rules](https://github.com/mruby/mruby/blob/master/tasks/mrbgem_spec.rake#L95) – thus adds new tasks

- And we haven't [imported `include_paths`](https://github.com/mruby/mruby/blob/master/tasks/mrbgem_spec.rake#L427) to these compilers.

As I understand, we don't need these compiler rules at all, so we should just skip them. Am I right?

palkan:
Fixes #3418.

I've tried to make the fix as small as possible.

Previously we had 2 places where compilers were configured (`Specification#setup` –  config options – and `List#check` – include paths).
That caused a problem when someone tries to use gem table only to build the dependencies (as in #3418) and does not want to define compilers rules.

From the other hand, we already configure compilers in `List#check` method, so, I think, it make sense to define rules there too.

### DIFF
--- a/tasks/mrbgem_spec.rake
+++ b/tasks/mrbgem_spec.rake
@@ -90,7 +90,9 @@ module MRuby
         build.libmruby << @objs
 
         instance_eval(&@build_config_initializer) if @build_config_initializer
+      end
 
+      def setup_compilers
         compilers.each do |compiler|
           compiler.define_rules build_dir, "#{dir}"
           compiler.defines << %Q[MRBGEM_#{funcname.upcase}_VERSION=#{version}]
@@ -418,6 +420,8 @@ module MRuby
         gem_table = generate_gem_table build
 
         @ary = tsort_dependencies gem_table.keys, gem_table, true
+
+        each(&:setup_compilers)
 
         each do |g|
           import_include_paths(g)


matz:
I will merge it to see how it works.

PR #3455, opened 2017-02-15
Update android build script.

crimsonwoods:
This PR updates build script for Android.

In usual case, required LDFLAGS and CFLAGS should be set by default.
If anyone need overwriting these flags, it can be done in `build_config.rb`.
I think, for many mruby users/developers, no requiring many flags to build mruby is good.

NOTE:
- Please use Android NDK `r12b` or later (recommended is latest `r13b`)
- When use Android NDK `r11c`, cannot build mruby by gcc for mips64 (by clang is okay)
- Security flags (e.g. `-Wa,--noexecstack`) is omitted (please set in your `build_config.rb` if needed).

### DIFF
--- a/examples/targets/build_config_android_arm64-v8a.rb
+++ b/examples/targets/build_config_android_arm64-v8a.rb
@@ -1,0 +1,26 @@
+MRuby::Build.new do |conf|
+
+  # Gets set by the VS command prompts.
+  if ENV['VisualStudioVersion'] || ENV['VSINSTALLDIR']
+    toolchain :visualcpp
+  else
+    toolchain :gcc
+  end
+
+  enable_debug
+
+  # include the default GEMs
+  conf.gembox 'default'
+end
+
+# Requires Android NDK r13 or later.
+MRuby::CrossBuild.new('android-arm64-v8a') do |conf|
+  params = { 
+    :arch => 'arm64-v8a', 
+    :platform => 'android-24',
+    :toolchain => :clang,
+  }
+  toolchain :android, params
+
+  conf.gembox 'default'
+end
--- a/examples/targets/build_config_android_armeabi.rb
+++ b/examples/targets/build_config_android_armeabi.rb
@@ -1,0 +1,26 @@
+MRuby::Build.new do |conf|
+
+  # Gets set by the VS command prompts.
+  if ENV['VisualStudioVersion'] || ENV['VSINSTALLDIR']
+    toolchain :visualcpp
+  else
+    toolchain :gcc
+  end
+
+  enable_debug
+
+  # include the default GEMs
+  conf.gembox 'default'
+end
+
+# Requires Android NDK r13 or later.
+MRuby::CrossBuild.new('android-armeabi') do |conf|
+  params = { 
+    :arch => 'armeabi', 
+    :platform => 'android-24',
+    :toolchain => :clang,
+  }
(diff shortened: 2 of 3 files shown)


crimsonwoods:
Hello, @felixjones 

What do you think about this change?

felixjones:
@crimsonwoods I left out adding default flags because I wanted that responsibility given to the user's personal configuration of the script. However, I actually use pretty much all of these flags in my personal build script, so if everyone's going to be using them because there's a clear benefit with zero draw-backs (and are completely necessary with cases like fix-cortex-a8) then I'm happy including them in the base toolchain script.

I am generally skeptical about making decisions on behalf of the end user, you can probably tell this from the TODO consideration at the top concerning selecting a default target architecture, so my original thinking was let the user define their switches.

---

The examples are nice and contribute to the promotion of MRuby being used on Android. You can probably tell I constructed the revised Android toolchain script from pulling apart and piecing together the old one, so I was not comfortable including my personal build scripts as examples. Your examples are far more simpler and clear than my personal scripts - perhaps moving compile flags into the base script has helped simplify things here.

I am hesitant about selecting platform 24 as a target in the examples, part of my goal with the revised base toolchain was to have the Android platform automatically select the latest installed via the ::platform method.
If I were to change anything it would be to comment out the :platform => 'android-24' so it is left there for the user to see it as a potential option in the example, or alternatively rename the example files to "build_config_android_24_armeabi.rb" and "build_config_android_24_arm64-v8a.rb" to describe that they will specifically target platform 24.

felixjones:
May I ask what is the motivation for selecting mips64r6 as opposed to mips64? I thought mips32 and mips64 are aliases for mips32r2 and mips64r6 with Android's clang compiler.

Also, what are your thoughts on having 'armeabi-v7a' with -mfpu neon and -mfloat-abi=hard as additional architectures? I noticed in the original script there was 'armeabi-v7a-hard' - but with me removing the compile switches from the base script that became unnecessary.

crimsonwoods:
### mips64r6

"mips64r6" is required to build mruby (using NDK r13b) for mips64 architecture on my environment.
On my environment, linker outputs below error:

```
libmruby.a(mt19937ar.o): linking mips:isa64 module with previous mips:isa64r6 modules
libmruby.a(mt19937ar.o): linking -mnan=legacy module with previous -mnan=2008 modules
```

So, due to avoid these errors, I changed "-march" option from "mips64" to "mips64r6".

### -mfpu neon

I think that to support neon is not required for mruby.  Because, as far as I know, mruby has no optimized code for neon.
But, maybe, some mrbgems contained optimized code for neon.

### -mfloat-abi=hard

According to 'setup.mk' bundled with NDK, I added "-mfloat-abi=softfp".
But I think that "-mfloat-abi=hard" should be able to select to improve performance.

felixjones:
The neon case I was specifically thinking about potential mrbgems, the more I think about it the more I think the option should be there.

Perhaps it is worth moving these switches to be parameters specific for the armeabi.

````ruby
MRuby::CrossBuild.new('android-armeabi-v7a-neon') do |conf|
  params = { 
    :arch => 'armeabi-v7a', 
    :mfpu => 'neon',
    # :mfloat_abi => 'hard',
  }
  toolchain :android, params

  conf.gembox 'default'
end
````

crimsonwoods:
I agree with you. It looks good to me.

Issue #3429, opened 2017-02-05
Stop running after eval


ksss:
From https://github.com/mruby/mruby/commit/48e0bbbfeea8268b09ad0a6bbc840834cc443fe0

```
$ cat t.rb
p "---"
p eval "1"
p "---"
```

Before (bf4e79cc62af809138bc7db7e54ece67080b5fa8 )

```
$ mruby -v t.rb
mruby 1.2.0 (2015-11-17)
00001 NODE_SCOPE:
00001   NODE_BEGIN:
00001     NODE_FCALL:
00001       NODE_SELF
00001       method='p' (385)
00001       args:
00001         NODE_STR "---" len 3
00002     NODE_FCALL:
00002       NODE_SELF
00002       method='p' (385)
00002       args:
00002         NODE_FCALL:
00002           NODE_SELF
00002           method='eval' (507)
00002           args:
00002             NODE_STR "1" len 1
00003     NODE_FCALL:
00003       NODE_SELF
00003       method='p' (385)
00003       args:
00003         NODE_STR "---" len 3
irep 0x7fc01ac36430 nregs=5 nlocals=1 pools=2 syms=2 reps=0
file: t.rb
    1 000 OP_LOADSELF	R1
    1 001 OP_STRING	R2	L(0)	; "---"
    1 002 OP_SEND	R1	:p	1
    2 003 OP_LOADSELF	R1
    2 004 OP_LOADSELF	R2
    2 005 OP_STRING	R3	L(1)	; "1"
    2 006 OP_SEND	R2	:eval	1
    2 007 OP_SEND	R1	:p	1
    3 008 OP_LOADSELF	R1
    3 009 OP_STRING	R2	L(0)	; "---"
    3 010 OP_SEND	R1	:p	1
    3 011 OP_STOP

"---"
1
"---"
```

After ( 48e0bbbfeea8268b09ad0a6bbc840834cc443fe0 )

```
mruby 1.2.0 (2015-11-17)
00001 NODE_SCOPE:
00001   NODE_BEGIN:
00001     NODE_FCALL:
00001       NODE_SELF
00001       method='p' (385)
00001       args:
00001         NODE_STR "---" len 3
00002     NODE_FCALL:
00002       NODE_SELF
00002       method='p' (385)
00002       args:
00002         NODE_FCALL:
00002           NODE_SELF
00002           method='eval' (507)
00002           args:
00002             NODE_STR "1" len 1
00003     NODE_FCALL:
[...]

Issue #3415, opened 2017-01-25
Crash when eval "Fiber.yield"


i110:
The following code causes a crash with a SEGV.

```ruby
Fiber.new { eval "Fiber.yield" }.resume
```

Is this possible in mruby?

kazuho:
For context, this is a bug that we ran into while looking for a way to load and evaluate a code within a fiber.

In H2O, we would like to evaluate mruby code that defines a Rack handler within a fiber, so that when a blocking operation is triggered within the evaluation `Fiber.yield` could be used to transfer that operation into C code.

Essentially, we would like to allow users to write something like the snippet below, which will be evaluated within a fiber (note: `join` methods calls `Fiber.yield` to delegate network operations to the main fiber).

```ruby
# cache the response from http://upstream, and use it for every request
cached_resp = http_request("http://upstream/").join[2].join
Proc.new do |env|
    [200, [], [cached_resp]]
end
```

matz:
It's not possible, mruby's fiber cannot cross C function boundaries, although causing SEGV is a bug.

kazuho:
@matz Thank you for the clarification. The reasoning sounds totally reasonable that you cannot call `yield` from an `eval`-ed script running in a fiber due to C function boundaries.

OTOH, may I ask if you have any idea that we could solve the issue? As stated in https://github.com/mruby/mruby/issues/3415#issuecomment-275016287 what we want to do is to evaluate user-specified script in a child fiber. We are more than happy to write such solution in C.

i110:
Another question: Are there standard ways to generate `Proc`'s instance from mruby source code?
(sorry for changing the topic from this issue)

We found that we can address the issue described in https://github.com/mruby/mruby/issues/3415#issuecomment-275016287 by passing the compiled code to some preparing method as a proc. Like the following, `prepare_app` method call that compiled proc in a fiber because compiled_proc may call `Fiber.yield`.
```ruby
def prepare_app(compiled_proc)
  Fiber.new do
      compiled_proc.call
  end.resume
end
```
But as far as I see the mruby API, there seem to

i110:
Sorry, I mistakenly closed this issue

matz:
48e0bbb allows you to call `Fiber.yield` from within `eval()`.

PR #3456, opened 2017-02-16
Fixed not properly returning calls of #eval

christopheraue:
Fixes the `eval(string)` issue of #3429. The generated proc is executed by the same code that's used by `Object#instance_eval(&block)`. Calling `eval('Fiber.yield')` (#3415) works.


christopheraue:
Wooops. Doesn't work after all...

PR #3458, opened 2017-02-16
GitLab.com tests for multiple configurations

dabroz:
Previously I've proposed #3258 as a method to test mruby on different build configurations (and found many bugs with rarely used combinations). However that method vastly increased Travis build times which is unfortunate. I was thinking about a way to enforce compatibility without such side-effect. Lately I was experimenting with GitLab CI and found it to be an excellent candidate for such testing.

I've created a mirror of the official mruby repository, which is automatically updated every 24h:
https://gitlab.com/dabroz/mruby

This patch introduces Gitlab testing, using GitLab CI, to test many different configurations:
 - 32/64 bit architecture
 - float/double
 - 16/32/64 bit int size
 - none/NaN/word boxing
 - ASCII/UTF8

using various compilers:
 - gcc-4.7
 - gcc-4.8
 - gcc-4.9
 - gcc-5
 - gcc-6
 - clang-3.5
 - clang-3.6
 - clang-3.7
 - clang-3.8
 - clang-3.9

#3457 is required for this to pass all tests on GitLab.

This has no impact on GitHub/Travis builds/testing/pull requests etc.

What is currently missing in GitLab testing is Windows/OSX platforms, since GitLab.com shared runners only support Docker (Linux) builds. I will try to devise a way to add different platforms to this testing procedure in the future.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,3214 @@
+---
+stages:
+- pretest
+- test
+pretest:
+  stage: pretest
+  image: registry.gitlab.com/dabroz/mruby:gcc47_0.1
+  tags:
+  - linux
+  variables:
+    CC: gcc-4.7
+    CXX: g++-4.7
+    LD: gcc-4.7
+  script: "./minirake all test"
+Test gcc-4.7 32bit:
+  stage: test
+  image: registry.gitlab.com/dabroz/mruby:gcc47_0.7
+  variables:
+    CC: gcc-4.7
+    CXX: g++-4.7
+    LD: gcc-4.7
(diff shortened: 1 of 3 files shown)


matz:
Thank you. But I am worried about heavy workload to Gitlab.com. If they are OK, I'd like to merge the PR.

dabroz:
This is a good point. I've created a support ticket at GitLab.com: https://gitlab.com/gitlab-com/support-forum/issues/1549

dabroz:
@matz I've got a reply in that thread:

> Please proceed, we plan to continue to provide unlimited build minutes for open source projects

So we should be OK.

PR #3459, opened 2017-02-16
Fixed not properly returning calls of #eval

christopheraue:
Fixes issue #3429.


christopheraue:
fixed by 097da2399ae9dc0a8db1761feae7d1921b7a89ad

PR #3460, opened 2017-02-16
Additional options for Android build script.

felixjones:
As discussed in the comments for PR https://github.com/mruby/mruby/pull/3455#issuecomment-280048567

This adds support for the supplied Android build script to configure the hardware floating point for armeabi-v7a.

````ruby
MRuby::CrossBuild.new('android-armeabi-v7a-neon-hard') do |conf|
  params = { 
    :arch => 'armeabi-v7a', 
    :mfpu => 'neon',
    :mfloat_abi => 'hard',
  }
  toolchain :android, params

  conf.gembox 'default'
end
````

This also adds a new example build script demonstrating this usage.

### DIFF
--- a/examples/targets/build_config_android_armeabi_v7a_neon_hard.rb
+++ b/examples/targets/build_config_android_armeabi_v7a_neon_hard.rb
@@ -1,0 +1,28 @@
+MRuby::Build.new do |conf|
+
+  # Gets set by the VS command prompts.
+  if ENV['VisualStudioVersion'] || ENV['VSINSTALLDIR']
+    toolchain :visualcpp
+  else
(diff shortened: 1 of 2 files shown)


felixjones:
I am not 100% happy with the usage of --no-warn-mismatch. This option seems to be the generally accepted thing to use to allow hardware float vector unit operations to be utilised, but I am concerned that it may be hiding potentially fatal compatibility problems.

The current behaviour will only activate --no-warn-mismatch if neither the 'soft' or 'softfp' mfloat-abi switches are used.

matz:
I will merge this PR. If we see any problem, we might revert the patch.

PR #3494, opened 2017-03-09
Avoid setup duplication

ksss:
I encountered build issue.

https://github.com/zzak/mruby-uri/pull/6
https://travis-ci.org/zzak/mruby-uri/jobs/207632178

Strangely, It built twice mruby-onig-regexp.
But, It built once when wrote `conf.gem 'mruby-onig-regexp'` at `build_config.rb`.

Because **mruby-test** called `MRuby::Gem::Specification#setup` to resolve this dependency from here https://github.com/mruby/mruby/blob/63dbed00946afda34178a479cfa38fa78d620a00/mrbgems/mruby-test/mrbgem.rake#L36

But when wrote `conf.gem`, it was skipped.

I think, it doesn't expected behavior.
So, I proposed to avoid setup duplication.

How do you think?

ref https://github.com/mattn/mruby-onig-regexp/pull/53

### DIFF
--- a/tasks/mrbgem_spec.rake
+++ b/tasks/mrbgem_spec.rake
@@ -48,10 +48,12 @@ module MRuby
         @version = "0.0.0"
         @mrblib_dir = "mrblib"
         @objs_dir = "src"
+        @after_setup = false
         MRuby::Gem.current = self
       end
 
       def setup
+        return nil if @after_setup
         MRuby::Gem.current = self
         MRuby::Build::COMMANDS.each do |command|
           instance_variable_set("@#{command}", @build.send(command).clone)
@@ -90,6 +92,7 @@ module MRuby
         build.libmruby << @objs
 
         instance_eval(&@build_config_initializer) if @build_config_initializer
+        @after_setup = true
       end
 
       def setup_compilers


ksss:
@take-cheeze Would you review this?

take-cheeze:
@ksss  It seems like it's building twice because there is two `MRuby::Build`(*host* and *test*) in [build_config.rb]( https://github.com/zzak/mruby-uri/blob/f022496308904e283a536e7afdae4ea644c7b96c/build_config.rb).
As far as I know about mruby-onig-regexp, it should build onig[mo|uruma] per each `MRuby::Build`.

ksss:
@take-cheez Thank you for reviewing.
I didn't make it clear enough.
It built twice in `MRuby::Build.new('test')`
So, `rake test` built three times in total. 

https://travis-ci.org/zzak/mruby-uri/jobs/207632178#L196-L201

take-cheeze:
@ksss 
You were worrying about double `setup` of mrbgem. Got it.

At the time when I wrote the minimum dependencies testing feature, it wasn't a mrbgem.
And I felt it shouldn't be a mrbgem like mruby compiler at that time.

Anyway I think at least dependencies solving shouldn't be done when `setup` method(and the block passed to gem spec) is evaluated.
Since dependencies won't be completed until all `setup` method of each mrbgems is evaluated.
So it should be moved to rake file of mruby build system or dependencies solving should be done in rake task that compiles mruby test code and generates C code.

(About this patch:)
I feel  `setup` method should only be called in [tasks/mrbgems.rake](https://github.com/mruby/mruby/blob/08a1dd2ac89333bd928a3300f712d1a7fb57e8ff/tasks/mrbgems.rake#L4) to reduce call order complexity.
Instead of ignoring it, it should raise some error on double evaluation of it to check build script errors.

ksss:
@take-cheeze Thank you for your kindness.
OK, I understood that this patch is not good.

I'll try another way.
Thanks.

PR #3529, opened 2017-03-21
Hash sub class creates new sub class objects instead of Hash

kou:


matz:
I prefer `self.new` instead of `new`.

kou:
OK. I changed to `self.new`.

Issue #3492, opened 2017-03-08
Null pointer dereference in get_file


clayton-shopify:
The following input demonstrates a crash:
```ruby
def a
  b
ensure
  yield
end

lambda { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { a { return } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } } }.call
```
The crash first appeared after this commit: https://github.com/mruby/mruby/commit/3390020fd0f60920c743a3f46c526d19fee9021c, and building with `-DMRB_DEFAULT_METHOD_MISSING` makes the crash go away, so the issue may be related to `method_missing`.

ASAN report:
```
ASAN:DEADLYSIGNAL
=================================================================
==5342==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000100f5587d bp 0x7fff5ed0a450 sp 0x7fff5ed0a320 T0)
    #0 0x100f5587c in get_file debug.c:12
    #1 0x100f557e0 in mrb_debug_get_filename debug.c:58
    #2 0x100f16d4e in each_backtrace backtrace.c:132
    #3 0x100f15b7a in output_backtrace backtrace.c:182
    #4 0x100f13614 in exc_output_backtrace backtrace.c:198
    #5 0x100f1239c in mrb_print_backtrace backtrace.c:283
    #6 0x100fdd1c2 in mrb_print_error print.c:33
    #7 0x100ef6a81 in main mruby.c:240
    #8 0x7fffb4357254 in start (libdyld.dylib+0x5254)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV debug.c:12 in get_file
==5342==ABORTING
Abort trap: 6
```
This issue was reported by https://hackerone.com/ssarong

matz:
The issue is related to #3438. I am investigating the solution (probably by removing save/restore backtrace).

bovi:
After this change I get the following warning on "Darwin daboAir.local 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64":

````
[...]

PR #3530, opened 2017-03-21
Replaced PRId64 with PRIdPTR for intptr_t snprintf

felixjones:
https://github.com/mruby/mruby/commit/3703aed7ab7c056ef7a58fd8d25b84b59f715dad introduced the warning https://github.com/mruby/mruby/issues/3492#issuecomment-287977542

> mruby/mrbgems/mruby-compiler/core/codegen.c:2202:47:
> warning: format specifies type 'long long' but the argument has type 'intptr_t' (aka 'long') [-Wformat]
> snprintf(buf, sizeof(buf), "$%" PRId64, (intptr_t)tree);

PRIdPTR fixes this  (expands to the correct format for intptr_t)

### DIFF
--- a/mrbgems/mruby-compiler/core/codegen.c
+++ b/mrbgems/mruby-compiler/core/codegen.c
@@ -2199,7 +2199,7 @@ codegen(codegen_scope *s, node *tree, int val)
       char buf[32];
       int sym;
 
-      snprintf(buf, sizeof(buf), "$%" PRId64, (intptr_t)tree);
+      snprintf(buf, sizeof(buf), "$%" PRIdPTR, (intptr_t)tree);
       sym = new_sym(s, mrb_intern_cstr(mrb, buf));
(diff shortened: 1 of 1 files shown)


matz:
I concern that `PRIdPTR` available on many (or all) platforms.
I should have cast to `int` not `intptr_t`, maybe, since the value must be 1 digit integer.

felixjones:
I thought PRIdPTR was a C99 definition as part of inttypes along with PRId64?

Casting to int would be more ideal if 1 digit will only ever be needed, %d can be straight used in this case.

felixjones:
If 1 digit is expected, then can we use %1d and resize char buf[32] to char buf[3] ?

matz:
It can be more than 1 digit. `$123456` is a valid variable name. Embarrassing.
I silenced the warning by 9eb0c11.

PR #3558, opened 2017-03-28
Should raise FloatDomainError

ksss:
I'm concerned about this use case.

```c
mrb_value v = mrb_funcall(mrb, num, "to_i", 0)
mrb_int i = mrb_fixnum(v);

if (i == 0) {
  // ...
```

example: https://github.com/iij/mruby-pack/blob/88a7fedea413568a1ff0410e109ff55a03b63a5f/src/pack.c#L1022

If `num` is `Float::INFINITY`, then `i` is `0`.
But I think, It's not expected value.

This behavior is compatible with CRuby.

```
$ ruby -e 'Float::INFINITY.to_i'
	from -e:1:in `<main>'
-e:1:in `to_i': Infinity (FloatDomainError)
```

### DIFF
--- a/mrbgems/mruby-time/src/time.c
+++ b/mrbgems/mruby-time/src/time.c
@@ -211,16 +211,7 @@ mrb_time_wrap(mrb_state *mrb, struct RClass *tc, struct mrb_time *tm)
   return mrb_obj_value(Data_Wrap_Struct(mrb, tc, &mrb_time_type, tm));
 }
 
-static void
-check_num_exact(mrb_state *mrb, double num)
-{
-  if (isinf(num)) {
-    mrb_raise(mrb, E_FLOATDOMAIN_ERROR, num < 0 ? "-Infinity" : "Infinity");
-  }
-  if (isnan(num)) {
-    mrb_raise(mrb, E_FLOATDOMAIN_ERROR, "NaN");
-  }
-}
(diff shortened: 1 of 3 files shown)


mrbgems/mruby-time/src/time.c
```diff
@@ -211,16 +211,7 @@ mrb_time_wrap(mrb_state *mrb, struct RClass *tc, struct mrb_time *tm)
...
-check_num_exact(mrb_state *mrb, double num)
-{
-  if (isinf(num)) {
-    mrb_raise(mrb, E_FLOATDOMAIN_ERROR, num < 0 ? "-Infinity" : "Infinity");
-  }
-  if (isnan(num)) {
-    mrb_raise(mrb, E_FLOATDOMAIN_ERROR, "NaN");
-  }
-}
+void check_num_exact(mrb_state *mrb, double num);
```
cremno: - functions with external linkage should be prefixed (`mrb_`)

- is `num` of type `double` (declaration) or `mrb_float` (definition)?

matz:
check_num_exact() should have propoer prefix e.g. `mrb_`

ksss:
Thank you for reviewings.
I rebased branch.

mrbgems/mruby-time/src/time.c
ksss: Thank you so much ✨

PR #3570, opened 2017-03-30
Modify class variable definition in singleton class; fix #3539

### LINKED ISSUE #3539: When class method deal with class variable , variable can not hold value.
hello dev, 
I would like to ask dev whether the behavior is correct or not.

in `cv.rb`
```
class Greeding
  @@text = nil
  class << self
    def text=(word)
      @@text = word
    end
  end

  def text
    @@text
  end
end

Greeding.text= "hello world"
puts Greeding.new.text
```
result
```
bash-3.2$ ./mruby/bin/mruby ./cv.rb

bash-3.2$ ruby ./cv.rb
hello world
```

I expect `Greeding#text` return `"good bye"`.
as follow code, it  behaves the same way on MRI and mruby
```
class Greeding
  @@text = nil
    def self.text=(word)
      @@text = word
    end

  def text
    @@text
  end
end

Greeding.text= "hello world"
puts Greeding.new.text
```

## Env 

- Darwin XXXX 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
- mruby commit id 051e40c0493f2de332f5439e3230c9fe6958bf1a
- ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin16]

keizo042:


keizo042:
Thank you!

PR #3583, opened 2017-04-04
[Proposal] Moving class definitions for tasks under lib/ directory?

udzura:
Current status: Classes/Modules used in mruby compilation tasks are defined in the same directory as tasks, with `.rake` extension.

I suppose this is confusing for average Rubyists, who are interested in mruby and its maintenance!


udzura:
One more question: How about `./mrblib/mrblib.rake` and `./src/mruby_core.rake` files?
This is natural when thinking they are `Makefile` alternatives, but strange when this is a Ruby project (I guess tasks are recommended to be in the same `tasks` or `lib/tasks` directory)

PR #3594, opened 2017-04-06
Get constant of parent class even if child class is defined in signle…

### LINKED ISSUE #3575: cannot see a constant defined in the grandparent module if the parent class is singleton class
I'm not sure Ruby's scoping rule allows it but it is not compatible with CRuby.

```rb
module A
  CONST = 123
  class << self
    p 1, CONST
    class B
      p 2, CONST
    end
  end
end
```

```
% ruby -v a.rb
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]
1
123
2
123

% bin/mruby a.rb
1
123
trace:
	[0] a.rb:6
	[1] a.rb:5
	[2] a.rb:3
	[3] a.rb:1
a.rb:6: uninitialized constant #<Class:0x7fd92a802ce0>::B::CONST (NameError)
```

keizo042:
…ton class; fix #3575

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -127,10 +127,19 @@ MRB_API struct RClass*
 mrb_class_outer_module(mrb_state *mrb, struct RClass *c)
 {
   mrb_value outer;
+  struct RClass *cls;
 
   outer = mrb_obj_iv_get(mrb, (struct RObject*)c, mrb_intern_lit(mrb, "__outer__"));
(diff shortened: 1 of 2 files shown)


tsahara:
Thank you for fixing #3575.  But I found a variant of the issue and reported it as #3598.

PR #3604, opened 2017-04-12
Find Class/Module in nested singleton class; fix #3598

### LINKED ISSUE #3598: cannot see a constant defined in the grandparent module from doubly-nested singleton class
```rb
module A
  CONST = 123
  class << self
    p 1, CONST
    class << self
      p 2, CONST
    end
  end
end
```

```
% ruby -v a.rb
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]
1
123
2
123

% bin/mruby a.rb
1
123
trace:
	[0] a.rb:6
	[1] a.rb:5
	[2] a.rb:3
	[3] a.rb:1
a.rb:6: uninitialized constant #<Class:#<Class:A>>::CONST (NameError)
```

keizo042:


src/variable.c
```diff
@@ -953,14 +953,13 @@ mrb_vm_const_get(mrb_state *mrb, mrb_sym sym)
     if (c->iv && iv_get(mrb, c->iv, sym, &v)) {
       return v;
     }
-    if (c->tt == MRB_TT_SCLASS) {
+    for (c2 = c; c2 && c2->tt == MRB_TT_SCLASS;) {
       mrb_value klass;
-      klass = mrb_obj_iv_get(mrb, (struct RObject *)c,
+      klass = mrb_obj_iv_get(mrb, (struct RObject *)c2,
                              mrb_intern_lit(mrb, "__attached__"));
       c2 = mrb_class_ptr(klass);
```
matz: I don't like for-statement like this. Use `while` here.

keizo042:
Thank you for the review.
I updated.

PR #3608, opened 2017-04-13
Fix an off-by-one causing a buffer overflow in mirb.

clayton-shopify:
If MRuby is built without libreadline, then a buffer overflow occurs in mirb when a line 1023 characters or longer is read:
```
$ ruby -e 'puts "0"*1023' | bin/mirb

> Segmentation fault (core dumped)
```
This happens because mirb reads into a 1024-byte buffer, but doesn't leave enough space for the `\n` and `\0` that will be appended to the input. This PR fixes the off-by one.

This issue was reported by https://hackerone.com/geeknik


clayton-shopify:
To reproduce this without removing the libreadline headers, you can just add `&& false` to this line and rebuild: https://github.com/mruby/mruby/blob/045e78c844289694cf36415261644d362ff2fd21/mrbgems/mruby-bin-mirb/mrbgem.rake#L6

PR #3740, opened 2017-07-11
fixed printing failed assertions

christopheraue:
The `msg` outside the block was not modified since it was shadowed by `msg` inside the block.


christopheraue:
Well, how this patch could break AppVeyor is beyond me...

bovi:
@christopheraue I think this is a bug of AppVeyor. I can't even find the build (https://ci.appveyor.com/project/matz/mruby)...

PR #3748, opened 2017-07-19
Specify Ruby files order through magic comments

palkan:
### Disclaimer

I marked this PR as _Experimental_ to emphasize that it's the main purpose is to discuss the problem (or this is not a problem at all?).

### The problem

Currently, Ruby files are loaded and compiled in lexicographical order. Thus we have to use such tricks as using file prefixes just to control this (even in [mruby core](https://github.com/mruby/mruby/tree/master/mrblib)). And that's an [example](https://github.com/palkan/acli/commit/266651c2b185032f7bfd4fa4edc44533b2b19069) from my lib (which inspired me to write this).

So, there is no way (other then naming hacks) to specify dependencies (usually we need it when using modules with shared logic).

### The Idea

I was thinking of how to tell the _builder_ to compile Ruby files in the specified order and came up with the following idea: let's use _magic comments_ to declare dependencies:

```ruby
# require: utils/logging

class Base
  include Utils::Logging
  ...
end
```

Looks like Sprockets 😏 

**Pros**: easy to use/read, backward-compatible.

**Cons**: that differs from Ruby.

### Other ideas

Another option I see is to specify the build order _hints_ somewhere in config:

```ruby
# Load specified files first, then everything else
conf.compile_first = %w(utils/**/*.rb zzz.rb)
```

**Pros**: no language changes.
**Cons**: harder to maintain (maybe, that's OK for mRuby libs, 'cause they're usually not too big).


bovi:
I agree that this is an annoying problem. I actually fall several times over this. Yet I don't like the magic comment. The load specific config file on the other side I like.

christopheraue:
Well, you can already use

```ruby
conf.rbfiles = Dir["#{conf.dir}/mrblib/utils/**/*.rb"] + Dir["#{conf.dir}/mrblib/zzz.rb"]
```
in the gemspec. As long as your files are still placed in the mrblib/ folder this should make sure that the files are loaded in the given order. I had a similar problem with my gem [where I did this](https://github.com/christopheraue/m-ruby-concurrently/blob/d5a10409755ee466e37fc0f357e24ac83cb7055e/mrbgem.rake#L32).

matz:
I prefer magic commit. I care about maintainability.

palkan:
Ok. If we are going with magic comments, how they should work?

Current implementation _ideologically_ works similar to MRI magic comments (at least, `encoding`, as I see [here](https://github.com/ruby/ruby/blob/ruby_2_4/parse.y#L6951)): we read source code line by line, ignore blank lines, parse magic "require" comments, scan until we find a line that do not match the magic comment (see test examples included in PR).

Not sure what to add/remove/improve.

palkan:
Update:
- Rebased/refactored on the top of the recent master changes (moving build utils to `lib/`)
- Added `libtest` task (to test build utils)

palkan:
🆙

Hey everyone!

Are we still interested in this feature? Maybe it's time to re-visit it?

matz:
Closing without merge — the underlying problem (file load order in mrblib) is real, but in practice mruby has lived with it for nine years using a single naming-prefix hack (`mrblib/10error.rb`) and an existing `spec.rbfiles = [...]` override that no in-tree mrbgem currently needs. The need hasn't grown.

If multi-file mrblib ordering issues start surfacing in several places, this proposal — or a variant with a less-confusing keyword than `# require:` (e.g. `# load_after:`, since runtime `require` has different semantics) — is worth revisiting. Thanks for the experiment and the prototype.

PR #3758, opened 2017-07-30
Improved speed of enumeration methods

christopheraue:
Replaced `yield if block_given?` with `block.call if block` because the latter is ~20% faster than the former:

```
$ time build/bench/bin/mruby -e '
  def m
    yield if block_given?
  end

  i = 10_000_000
  while i > 0; m{}; i -= 1 end'

real    0m1.899s
user    0m1.899s
sys     0m0.000s
```

```
$ time build/bench/bin/mruby -e '
  def m(&b)
    b.call if b
  end

  i = 10_000_000
  while i > 0; m{}; i -= 1 end'

real    0m1.511s
user    0m1.510s
sys     0m0.001s

### DIFF
--- a/mrbgems/mruby-array-ext/mrblib/array.rb
+++ b/mrbgems/mruby-array-ext/mrblib/array.rb
@@ -252,7 +252,7 @@ class Array
 
   # for efficiency
   def reverse_each(&block)
-    return to_enum :reverse_each unless block_given?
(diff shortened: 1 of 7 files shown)


matz:
Thank you. You may want to add your name to `AUTHORS` file.

christopheraue:
Thanks for the invitation: #3759

PR #3769, opened 2017-08-09
add method(compact, compact!) and test of Hash to mruby-hash-ext

vvakame:
I was implement `Hash#compact` and `Hash#compact!` to mruby-hash-ext.

I saw this document and code.
* http://ruby-doc.org/core-2.4.1/Hash.html#method-i-compact
* https://github.com/mruby/mruby/blob/f9c3ebd29d7410be209b3f22b9923f0f81f0141d/mrbgems/mruby-array-ext/mrblib/array.rb#L218-L251

and exec below code.
```
$ irb -v
irb 0.9.6(09/06/30)
$ ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
$ irb
irb(main):001:0> {:a => true, :b => nil}.compact!
=> {:a=>true}
irb(main):002:0> {:a => true}.compact!
=> nil
```

CRuby document says.

Array http://ruby-doc.org/core-2.4.1/Array.html#method-i-compact

> compact! → ary or nil
> Removes nil elements from the array.
> 
> Returns nil if no changes were made, otherwise returns the array.

Hash http://ruby-doc.org/core-2.4.1/Hash.html#method-i-compact

> compact! → hsh
> Removes all nil values from the hash. Returns the hash.

Hash document is not pointing to something like `Returns nil if no changes were made, otherwise returns the hash.`...
Is my implementation right?

### DIFF
--- a/mrbgems/mruby-hash-ext/mrblib/hash.rb
+++ b/mrbgems/mruby-hash-ext/mrblib/hash.rb
@@ -115,6 +115,40 @@ class Hash
   alias update merge!
 
   ##
+  # call-seq:
+  #    hsh.compact     -> new_hsh
+  #
+  # Returns a new hash with the nil values/key pairs removed
+  #
+  #    h = { a: 1, b: false, c: nil }
+  #    h.compact     #=> { a: 1, b: false }
+  #    h             #=> { a: 1, b: false, c: nil }
+  #
+  def compact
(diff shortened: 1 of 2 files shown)


matz:
You may want to add your name to the `AUTHORS` file.  Send a pull request.

PR #3776, opened 2017-08-10
Fixed calling missing method through super with 126 args

christopheraue:
The following input raised an error that it shouldn't:

```
$ bin/mruby -e 'class SCls
  def method_missing(m, *)
    puts "calling missing #{m}"
  end
end

class Cls < SCls
  def unknown
    super(
      1,2,3,4,5,6,7,8,9,0,
      11,2,3,4,5,6,7,8,9,0,
      21,2,3,4,5,6,7,8,9,0,
      31,2,3,4,5,6,7,8,9,0,
      41,2,3,4,5,6,7,8,9,0,
      51,2,3,4,5,6,7,8,9,0,
      61,2,3,4,5,6,7,8,9,0,
      71,2,3,4,5,6,7,8,9,0,
      81,2,3,4,5,6,7,8,9,0,
      91,2,3,4,5,6,7,8,9,0,
      101,2,3,4,5,6,7,8,9,0,
      111,2,3,4,5,6,7,8,9,0,
      121,2,3,4,5,6)
  end
end
Cls.new.unknown'
trace:
        [0] -e:25
-e:25: can't convert Fixnum into Proc (TypeError)
```

To blame was:
```
if (n == CALL_MAXARGS-1) {
  [...]
  n++;
}
[...]
if (n == CALL_MAXARGS) {
  bidx = a+2;
}
[...]
blk = regs[bidx];
```
This got the third register instead of the last one with the block and tried to convert it into a proc. The solution was to recreate the control flow of OP_SEND: First getting the block and then use it to create the regs for method missing.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1379,9 +1379,7 @@ RETRY_TRY_BLOCK:
       c = mrb_class(mrb, recv);
       m = mrb_method_search_vm(mrb, &c, mid);
       if (!m) {
-        mrb_value sym = mrb_symbol_value(mid);
         mrb_sym missing = mrb_intern_lit(mrb, "method_missing");
-
         m = mrb_method_search_vm(mrb, &c, missing);
         if (!m) {
           mrb_value args;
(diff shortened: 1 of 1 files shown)


matz:
CI fails. Due to changes in L1382 and L1407 (mid was updated).

christopheraue:
Argh! Sorry for my inability to properly run the tests -.-

Issue #3590, opened 2017-04-04
Null pointer dereference in OP_ENTER


clayton-shopify:
The following input demonstrates a crash:
```ruby
class A
  def foo
  end
end

class B < A
  def foo(*args)
    super(*args, &:b)
  end
end

B.new.foo
```
ASAN report:
```
==68149==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000028 (pc 0x0001007a4d01 bp 0x7fff5f5d0170 sp 0x7fff5f5c8660 T0)
==68149==The signal is caused by a READ memory access.
==68149==Hint: address points to the zero page.
    #0 0x1007a4d00 in mrb_vm_exec vm.c:1573
    #1 0x100792f39 in mrb_vm_run vm.c:824
    #2 0x1007c5849 in mrb_top_run vm.c:2630
    #3 0x100896845 in mrb_load_exec parse.y:5762
    #4 0x100897655 in mrb_load_file_cxt parse.y:5771
    #5 0x10062fe66 in main mruby.c:227
    #6 0x7fffbbbba234 in start (libdyld.dylib:x86_64+0x5234)

==68149==Register values:
rax = 0x0000000000000028  rbx = 0xf2f20000f2f2f200  rcx = 0x0000000000000028  rdx = 0x0000100000000005
rdi = 0x0000100000000000  rsi = 0x0000100000000000  rbp = 0x00007fff5f5d0170  rsp = 0x00007fff5f5c8660
 r8 = 0x0000100000000000   r9 = 0x0000100000000000  r10 = 0x0000000000000080  r11 = 0xffffe1d0a0a4de00
r12 = 0xf2f20000f1f1f1f1  r13 = 0x00001fffebeb9e84  r14 = 0xf2f20000f2f2f2f2  r15 = 0xf2f2f2f2f2040000
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV vm.c:1573 in mrb_vm_exec
==68149==ABORTING
Abort trap: 6
```
This issue was reported by Dinko Galetic & Denis Kasak (https://hackerone.com/dgaletic).

Issue #3551, opened 2017-03-27
Heap buffer overflow in mrb_str_to_dbl


clayton-shopify:
The following input demonstrates a crash:
```ruby
%W[0 0 000000000000000000000000].map(&:to_f) while true
```
ASAN report:
```
==6422==ERROR: AddressSanitizer: SEGV on unknown address 0x603000031e00 (pc 0x000106affe28 bp 0x7fff59207eb0 sp 0x7fff59207c60 T0)
    #0 0x106affe27 in mrb_str_to_dbl (mruby:x86_64+0x100111e27)
    #1 0x106b1b527 in mrb_str_to_f (mruby:x86_64+0x10012d527)
    #2 0x106b4bee9 in mrb_exec_irep (mruby:x86_64+0x10015dee9)
    #3 0x106b4d605 in mrb_f_send (mruby:x86_64+0x10015f605)
    #4 0x106b5d2d0 in mrb_vm_exec (mruby:x86_64+0x10016f2d0)
    #5 0x106b52959 in mrb_vm_run (mruby:x86_64+0x100164959)
    #6 0x106b84c39 in mrb_top_run (mruby:x86_64+0x100196c39)
    #7 0x106c53a25 in mrb_load_exec (mruby:x86_64+0x100265a25)
    #8 0x106c54835 in mrb_load_file_cxt (mruby:x86_64+0x100266835)
    #9 0x1069f03a6 in main mruby.c:227
    #10 0x7fffbbbba234 in start (libdyld.dylib:x86_64+0x5234)
```
This issue was reported by Dinko Galetic and Denis Kasak (https://hackerone.com/dgaletic).

I did a bit of investigation and it appears the crash happens when a garbage collection occurs during an OP_ARRAY:
```
irep 0x60c0000133c0 nregs=4 nlocals=1 pools=2 syms=2 reps=0
file: 215879.rb
    1 000 OP_JMP	007
    1 001 OP_STRING	R1	L(0)	; "0"
    1 002 OP_STRING	R2	L(0)	; "0"
    1 003 OP_STRING	R3	L(1)	; "000000000000000000000000"
    1 004 OP_ARRAY	R1	R1	3    <---------- CRASH HERE
    1 005 OP_LOADSYM	R2	:to_f
    1 006 OP_SENDB	R1	:map	0
    1 007 OP_LOADT	R1
    1 008 OP_JMPIF	R1	001
    1 009 OP_LOADNIL	R1
    1 010 OP_STOP
```
[...]

Issue #3504, opened 2017-03-14
Crash when super is called with splat arguments and a block


clayton-shopify:
The following input demonstrates a crash:
```ruby
class A
  def foo(*args)
  end
end

class B < A
  def foo(*args)
    super(*args, &:bar)
  end

  # It doesn't matter if this method is defined or not
  def bar
  end
end

B.new.foo
```
I haven't tracked down the root cause but I did notice that `regs[1]` is a `MRB_TT_PROC` here, but it should be an array: https://github.com/mruby/mruby/blob/4b1e5d47b326d0ca3272370b8d13b85344f929fd/src/vm.c#L1560

That causes `argc` to be set incorrectly a couple lines later, which results in a crash here: https://github.com/mruby/mruby/blob/4b1e5d47b326d0ca3272370b8d13b85344f929fd/src/vm.c#L1610

ASAN report:
```
==23760==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x62f00000c430 at pc 0x00010c9969eb bp 0x7fff53605560 sp 0x7fff53604d20
READ of size 16 at 0x62f00000c430 thread T0
    #0 0x10c9969ea in __asan_memcpy (libclang_rt.asan_osx_dynamic.dylib+0x419ea)
    #1 0x10c5f5dec in array_copy array.c:71
    #2 0x10c5f5bcf in mrb_ary_new_from_values array.c:80
    #3 0x10c769bae in mrb_vm_exec (mruby+0x100178bae)
    #4 0x10c756f99 in mrb_vm_run (mruby+0x100165f99)
    #5 0x10c789639 in mrb_top_run (mruby+0x100198639)
    #6 0x10c858b45 in mrb_load_exec (mruby+0x100267b45)
    #7 0x10c859955 in mrb_load_file_cxt (mruby+0x100268955)
    #8 0x10c5f35d6 in main mruby.c:227
    #9 0x7fffaf458254 in start (libdyld.dylib+0x5254)

0x62f00000c430 is located 0 bytes to the right of 49200-byte region [0x62f000000400,0x62f00000c430)
allocated by thread T0 here:
    #0 0x10c99ff87 in wrap_realloc (libclang_rt.asan_osx_dynamic.dylib+0x4af87)
    #1 0x10c6ec675 in mrb_default_allocf (mruby+0x1000fb675)
    #2 0x10c66e228 in mrb_realloc_simple gc.c:201
    #3 0x10c66e90e in mrb_realloc gc.c:215
[...]

PR #3779, opened 2017-08-12
Some more vm refactorings

christopheraue:
One additional thing: While going through the code I was wondering if sections:

https://github.com/mruby/mruby/blob/8bf492f12707777c321dcf40494f757947649f62/src/vm.c#L1366-L1369

and 

https://github.com/mruby/mruby/blob/8bf492f12707777c321dcf40494f757947649f62/src/vm.c#L1560-L1563

are needed or could be removed.

In both cases the stack should have room for the block. Actually, the stack was just accessed at the block's offset a few lines earlier. Removing those lines does not break any tests.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1345,7 +1345,10 @@ RETRY_TRY_BLOCK:
       else {
         blk = regs[bidx];
         if (!mrb_nil_p(blk) && mrb_type(blk) != MRB_TT_PROC) {
-          blk = regs[bidx] = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
+          /* store the converted proc not directly in the stack because the stack
+             might have been reallocated during mrb_convert_type(), see #3622 */
+          blk = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
+          regs[bidx] = blk;
         }
       }
       c = mrb_class(mrb, recv);
@@ -1528,7 +1531,10 @@ RETRY_TRY_BLOCK:
       recv = regs[0];
       blk = regs[bidx];
       if (!mrb_nil_p(blk) && mrb_type(blk) != MRB_TT_PROC) {
-        blk = regs[bidx] = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
+        /* store the converted proc not directly in the stack because the stack
+           might have been reallocated during mrb_convert_type(), see #3622 */
+        blk = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
+        regs[bidx] = blk;
       }
       c = ci->target_class->super;
       m = mrb_method_search_vm(mrb, &c, mid);
--- a/src/vm.c
+++ b/src/vm.c
@@ -1339,20 +1339,12 @@ RETRY_TRY_BLOCK:
 
       recv = regs[a];
       if (GET_OPCODE(i) != OP_SENDB) {
-        if (bidx >= ci->nregs) {
-          ci->nregs = bidx+1;
-          stack_extend(mrb, ci->nregs);
-        }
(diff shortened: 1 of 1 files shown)


matz:
The first check was added by 5513fcee2. It seems to be added for no reason (you may remove it).
The second check was added by 93d802987 related to #3551. According to #3551 `bidx` may be bigger than `ci->nregs`, so we have to check before accessing `regs[bidx]`.

matz:
I made the following change:

``` diff
index 9c427772..e1471051 100644
--- a/src/vm.c
+++ b/src/vm.c
@@ -1201,13 +1201,13 @@ RETRY_TRY_BLOCK:
       if (GET_OPCODE(i) != OP_SENDB) {
         SET_NIL_VALUE(regs[bidx]);
       }
+      else if (bidx >= mrb->c->ci->nregs) {
+        stack_extend(mrb, bidx+1);
+        mrb->c->ci->nregs = bidx+1;
+      }
       else {
         mrb_value blk = regs[bidx];
         if (!mrb_nil_p(blk) && mrb_type(blk) != MRB_TT_PROC) {
-          if (bidx >= mrb->c->ci->nregs) {
-            stack_extend(mrb, bidx+1);
-            mrb->c->ci->nregs = bidx+1;
-          }
           regs[bidx] = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
         }
       }
@@ -1383,7 +1383,6 @@ RETRY_TRY_BLOCK:
       mrb_sym mid = ci->mid;
       int a = GETARG_A(i);
       int n = GETARG_C(i);
-      mrb_value blk;
       int bidx;
 
       if (mid == 0 || !ci->target_class) {
@@ -1430,13 +1429,16 @@ RETRY_TRY_BLOCK:
       else {
         bidx = a+n+1;
       }
-      blk = regs[bidx];
-      if (!mrb_nil_p(blk) && mrb_type(blk) != MRB_TT_PROC) {
-        if (bidx >= ci->nregs) {
-          stack_extend(mrb, bidx+1);
-          ci->nregs = bidx+1;
+      if (bidx >= ci->nregs) {
+        stack_extend(mrb, bidx+1);
+        ci->nregs = bidx+1;
+      }
+      else {
+        mrb_value blk = regs[bidx];
+
+        if (!mrb_nil_p(blk) && mrb_type(blk) != MRB_TT_PROC) {
+          regs[bidx] = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
         }
-        regs[bidx] = mrb_convert_type(mrb, blk, MRB_TT_PROC, "Proc", "to_proc");
       }
 
       /* push callinfo */
```

christopheraue:
Alright, I also investigated some more.

Like you said, because of #3504 the `ci->nregs = bidx+1` was introduced in b64f08784cc50861dec80c26fce715982c02c9e9. This led to the follow up error #3551 whose fix introduced the `if (bidx >= ci->nregs)` check in 071164b7999d0932fd60fb3c580a6c374ab1cf4f and the `stack_extend(mrb, ci->nregs)` in 93d802987e722444d0520a8d3f28002841c4f825.

Then, the code causing #3504 reappeared again in #3590. The fix for it moved the code dealing with the block in OP_SUPER from below the `cipush`  to above the `cipush` (d9fb8b69b0621e8cd2c7c43fd8511a83718d7e45). The `if (bidx >= ci->nregs) { ... }` from then on works with the original callinfo and not the pushed one. `ci->nregs` needed to be modified for the pushed one because it is initialized to 0. But for the original ci it is propertly set and a check is not needed.

It looks to me, that getting the block before `cipush` seems to be the proper fix for the issues. Checking out d9fb8b69b0621e8cd2c7c43fd8511a83718d7e45 and removing the `if (bidx >= ci->nregs)` blocks does not crash with the code given in #3504, #3551 and #3590. master does not crash either.

I'm going to add a commit with the checks removed. Just leave it out if I missed something.

PR #3799, opened 2017-08-29
Refactored #assert_raise and #assert_nothing_raised

christopheraue:
`assert_nothing_raised` didn't check the given exceptions anyway.

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -143,52 +143,41 @@ def assert_not_include(collection, obj, msg = nil)
   assert_false(collection.include?(obj), msg, diff)
 end
 
-def assert_raise(*exp)
-  ret = true
-  if $mrbtest_assert
-    $mrbtest_assert_idx += 1
-    msg = exp.last.class == String ? exp.pop : nil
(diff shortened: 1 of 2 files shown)


tsahara:
`assert_raise` API is changed accidentally.  Before this pull request was merged, `assert_raise` checked that the raised exception was _one of_ the arguments of `assert_raise`.  Now we can specify only one exception class.

christopheraue:
Ok, I'm going to fix that. But I'm curious in which cases you use it that way. Because only the first raised exception will be catched in `assert_raise` anyway.

tsahara:
[This test case](https://github.com/iij/mruby-regression-tests/blob/master/test/2800.rb) (a regression test for #2800) raises TypeError on master and 1.3.0 but ArgumentError on 1.2.0 and [our stable version of mruby based on 1.1.0](https://github.com/iij/mruby/tree/stable_1_0).  It should accept both TypeError and ArgumentError to run it on multiple versions of mruby.

Thanks for the fix!

christopheraue:
Ah, ok. Thanks for clearing that up.

PR #3806, opened 2017-09-02
Trying to make the code more readable

flaviommedeiros:
Hi all,

I'm running some C code analysis tool that pointed me to this small code fragment. I'm suggesting this patch to try to make the source code more understandable by removing mixing of ternary if statements with attributions.

Would you guys agree with this type of changes? The tools pointed me to almost 20 cases like this one in the source code of mruby.

### DIFF
--- a/src/numeric.c
+++ b/src/numeric.c
@@ -380,7 +380,11 @@ flo_shift(mrb_state *mrb, mrb_value x, mrb_int width)
 #if defined(_ISOC99_SOURCE)
     val = trunc(val);
 #else
-    val = val > 0 ? floor(val) : ceil(val);
+    if (val > 0){
+        val = floor(val);    
+    } else {
+        val = ceil(val);
+    }
 #endif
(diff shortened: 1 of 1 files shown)


christopheraue:
I'm not a fan of these one-short-line-per-branch if statements. They have too much visual weight for the single operation they mostly only do. In this case, the ternary version also better emphasizes the assignment because it starts with it. But that's just my view...

dkasak:
I must say I also have an intense disliking of these particular changes since this concrete use of the ternary operator was a textbook example of the use of the ternary expression and, in my opinion, very clean. This just replaced it with a more imperative version for no good reason at all, IMO.

To make matters worse, the style was borked in the patch (with the opening brace being juxtaposed to the closing parenthesis, without whitespace).

FWIW, I don't think the result is any clearer at all and I would refrain from such unproductive changes in general since that can potentially open a huge can of worms (not to mention bugs, by accident), but then again, I'm not the one maintaing the code.

PR #3814, opened 2017-09-25
allow to override default printing function

### LINKED ISSUE #3734: Printing mruby errors without stderr
When an error occurs within the mruby VM the simplest method I've seen to identify it is the `mrb_print_error` function which displays the error as well as the backtrace via `mrb_print_backtrace`. Unfortunately these functions print directly to stderr which is not available in all circumstances. It would be advantageous to have these functions print to a string buffer under the hood. With a string based representation it would be possible to display error messages to users even when stderr is not available.

The specific circumstances that I've needed this support was debuging a GUI application ported to  windows and displaying error strings via the `MessageBox()` API. I currently have a rather messy patch in place to convert the `fprintf` calls to `snprintf`/`strncat` calls and it would be great to see this support available upstream.

dabroz:
`mrb_state` now features a `mrb_print_func` callback, that allows to use custom printing function to be used instead of the default fprintf implementation, like `__android_log_print` (Android) or `OutputDebugString` (Win32).

Fixes #3734

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -113,6 +113,12 @@ struct mrb_state;
  */
 typedef void* (*mrb_allocf) (struct mrb_state *mrb, void*, size_t, void *ud);
 
+/**
+ * Function pointer type of custom printing function. It will be used instead of
+ * fprintf(stdout/stderr) if provided.
+ */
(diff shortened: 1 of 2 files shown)


dabroz:
@matz any comment on this?

fourmond:
This would definitely come in useful for me, someday.

PR #3815, opened 2017-09-25
Fix MSVC 14.0 warnings

dabroz:
I fixed all warnings exposed by Appveyor MSVC 14.0, except one related to exception handling. Mostly fixing `int` vs `mrb_int` casts.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -858,7 +858,7 @@ mrb_get_mid(mrb_state *mrb) /* get method symbol */
   return mrb->c->ci->mid;
 }
 
-static inline int
+static inline mrb_int
 mrb_get_argc(mrb_state *mrb) /* get argc */
 {
   return mrb->c->ci->argc;
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1054,7 +1054,7 @@ MRB_API mrb_sym mrb_obj_to_sym(mrb_state *mrb, mrb_value name);
 MRB_API mrb_bool mrb_obj_eq(mrb_state*, mrb_value, mrb_value);
 MRB_API mrb_bool mrb_obj_equal(mrb_state*, mrb_value, mrb_value);
 MRB_API mrb_bool mrb_equal(mrb_state *mrb, mrb_value obj1, mrb_value obj2);
-MRB_API mrb_value mrb_convert_to_integer(mrb_state *mrb, mrb_value val, int base);
+MRB_API mrb_value mrb_convert_to_integer(mrb_state *mrb, mrb_value val, mrb_int base);
 MRB_API mrb_value mrb_Integer(mrb_state *mrb, mrb_value val);
 MRB_API mrb_value mrb_Float(mrb_state *mrb, mrb_value val);
 MRB_API mrb_value mrb_inspect(mrb_state *mrb, mrb_value obj);
--- a/include/mruby/string.h
+++ b/include/mruby/string.h
@@ -409,7 +409,7 @@ MRB_API int mrb_str_cmp(mrb_state *mrb, mrb_value str1, mrb_value str2);
 MRB_API char *mrb_str_to_cstr(mrb_state *mrb, mrb_value str);
 
 mrb_value mrb_str_pool(mrb_state *mrb, mrb_value str);
-mrb_int mrb_str_hash(mrb_state *mrb, mrb_value str);
+uint32_t mrb_str_hash(mrb_state *mrb, mrb_value str);
 mrb_value mrb_str_dump(mrb_state *mrb, mrb_value str);
 
 /*
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -3452,7 +3452,7 @@ backref_error(parser_state *p, node *n)
   c = (int)(intptr_t)n->car;
 
   if (c == NODE_NTH_REF) {
-    yyerror_i(p, "can't set variable $%" MRB_PRId, (mrb_int)(intptr_t)n->cdr);
+    yyerror_i(p, "can't set variable $%" MRB_PRId, (int)(intptr_t)n->cdr);
   }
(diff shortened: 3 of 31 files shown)


include/mruby.h
```diff
@@ -176,6 +176,8 @@ struct mrb_cache_entry {
 };
 #endif
 
+typedef uint32_t khint_t;
+
 struct mrb_jmpbuf;
```
matz: I don't like this disclosure. The scope should be as small as possible.
Can you elaborate?

mrbgems/mruby-compiler/core/parse.y
```diff
@@ -3452,7 +3452,7 @@ backref_error(parser_state *p, node *n)
   c = (int)(intptr_t)n->car;
 
   if (c == NODE_NTH_REF) {
-    yyerror_i(p, "can't set variable $%" MRB_PRId, (mrb_int)(intptr_t)n->cdr);
+    yyerror_i(p, "can't set variable $%" MRB_PRId, (int)(mrb_int)(intptr_t)n->cdr);
```
matz: This triple cast appears weird. I think `(int)(intptr_t)` should work.

mrbgems/mruby-eval/src/eval.c
```diff
@@ -298,7 +298,7 @@ f_eval(mrb_state *mrb, mrb_value self)
 
   mrb_get_args(mrb, "s|ozi", &s, &len, &binding, &file, &line);
 
-  proc = create_proc_from_string(mrb, s, len, binding, file, line);
+  proc = create_proc_from_string(mrb, s, (int)len, binding, file, line);
```
matz: This cast means nothing because `len` is casted to `size_t` in `create_proc_from_string`. I'd rather change the prototype of `create_proc_from_string` to take `mrb_int len`.

include/mruby/string.h
```diff
@@ -409,7 +409,7 @@ MRB_API int mrb_str_cmp(mrb_state *mrb, mrb_value str1, mrb_value str2);
 MRB_API char *mrb_str_to_cstr(mrb_state *mrb, mrb_value str);
 
 mrb_value mrb_str_pool(mrb_state *mrb, mrb_value str);
-mrb_int mrb_str_hash(mrb_state *mrb, mrb_value str);
+khint_t mrb_str_hash(mrb_state *mrb, mrb_value str);
```
matz: Ah, this is the reason you have disclosed `khint_t` to global.

But `mrb_int` is big enough for most of the case (because its modulo is used in the hash functions). If you really care, use `uint32_t` instead of `khint_t`.

src/pool.c
```diff
@@ -30,7 +30,7 @@ struct mrb_pool_page {
   size_t offset;
   size_t len;
   void *last;
-  char page[];
+  char page[1];
```
matz: This makes allocation size bigger by 1 byte.

mrbgems/mruby-sprintf/src/sprintf.c
```diff
@@ -194,7 +194,7 @@ check_name_arg(mrb_state *mrb, int posarg, const char *name, int len)
 
 #define GETNAMEARG(id, name, len) (\
   check_name_arg(mrb, posarg, name, len),\
-  (posarg = -2, mrb_hash_fetch(mrb, get_hash(mrb, &hash, argc, argv), id, mrb_undef_value())))
+  (posarg = -2, mrb_hash_fetch(mrb, get_hash(mrb, &hash, (int)argc, argv), id, mrb_undef_value())))
```
matz: How about making `argc` of `get_hash` to `mrb_int`?

mrbgems/mruby-sprintf/src/sprintf.c
```diff
@@ -999,7 +999,8 @@ mrb_str_format(mrb_state *mrb, mrb_int argc, const mrb_value *argv, mrb_value fm
       case 'A': {
         mrb_value val = GETARG();
         double fval;
-        int i, need = 6;
+        int i;
```
matz: I'd rather replace all `int` to `mrb_int` in the function.

src/pool.c
nobu: Shouldn't `page_alloc()` use `offsetof(struct mrb_pool_page, page)` instead of `sizeof(struct mrb_pool_page)` together?

mattn: This simply have to disable C4100.

```
#pragma warning(disable : 4200) 
```
see https://msdn.microsoft.com/en-us/library/79wf64bc.aspx

>This non-standard extension is often used to interface code with external data structures that have a variable length. If this scenario applies to your code, you can disable the warning:

include/mruby.h
dabroz: Fixed.

mrbgems/mruby-compiler/core/parse.y
dabroz: Fixed.

mrbgems/mruby-eval/src/eval.c
dabroz: Fixed.

include/mruby/string.h
dabroz: Fixed, changed `mrb_str_hash` return type to `uint32_t`.

src/pool.c
dabroz: Fixed, used `#pragma warning` with MSVC.

mrbgems/mruby-sprintf/src/sprintf.c
dabroz: Fixed.

mrbgems/mruby-sprintf/src/sprintf.c
dabroz: Fixed. But because the address of `i` was used in `frexp` call, a new temporary variable has been introduced.

dabroz:
Fixed all mentioned issues.

PR #3836, opened 2017-10-28
Remove one of the lines that did the same thing

ukrainskiysergey:
This line is apparently useless, because a bit lower there is almost the same line.


matz:
I' d rather remove the other line.

PR #3872, opened 2017-12-08
[WIP] Add sanitizer support in gcc like toolchain.

take-cheeze:
Travis build of macos takes too much time to fix...
Is there any fast starting macos CI service?

- [x] Define `enable_sanitizer` method in gcc like toolchain(including clang).
- [x] Fix leak of mruby-socket test caused by `tempnam`.
- [ ] Need to fix travis osx build error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,27 @@
 language: c
 
-sudo: false
-
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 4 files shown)


matz:
We cannot accept this bundle of multi patches. Please resubmit.

PR #3878, opened 2017-12-09
Fix socket closing

take-cheeze:
Seems like it needs to call `closesocket` instead of `close` in Windows.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms737582(v=vs.85).aspx

[`BasicSocket#close`](https://github.com/iij/mruby-socket/blob/ab54185005ec87fe4f5b10df95ad29659884141b/src/socket.c#L721) does support it but in mruby-io it isn't.

### DIFF
--- a/mrbgems/mruby-socket/test/unix.rb
+++ b/mrbgems/mruby-socket/test/unix.rb
@@ -1,5 +1,3 @@
-unless SocketTest.win?
-
 def unixserver_test_block
   path = SocketTest.tmppath
   File.unlink path rescue nil
@@ -126,5 +124,3 @@ assert('UNIXSocket#recvfrom') do
     # a[1][1] would be "" or something
   end
 end
-
-end  # win?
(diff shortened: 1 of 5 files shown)


mrbgems/mruby-socket/test/unix.rb
```diff
@@ -1,5 +1,3 @@
-unless SocketTest.win?
```
matz: Is this change intentional? Does UNIXSocket test run OK for Windows?

take-cheeze: At least many `mruby-io` related stuff is failing too so I don't think that line will do anything good.
(Unless you want to hide the CI test errors in windows.)
You should exclude whole `mruby-io` and `mruby-socket` mrbgem instead of doing those kind of thing.

This PR only makes fix to exceptions inside Windows' dynamic library, that's all.
I've fixed segv error and there is the rest errors.

mattn: I suggest to merge this since CI is failing (crash) on Windows.

crash

take-cheeze:
After some rethinking `BasicSocket#is_socket=` should be renamed to `_is_socket=` since it's for internal use.

matz:
#3878 merged.

PR #3882, opened 2017-12-12
use filename in locale

mattn:
filename should be string in locale


mrbgems/mruby-io/src/io.c
```diff
@@ -557,6 +557,7 @@ mrb_cloexec_open(mrb_state *mrb, const char *pathname, mrb_int flags, mrb_int mo
 {
   mrb_value emsg;
   int fd, retry = FALSE;
+  char* utf8 = mrb_locale_from_utf8(pathname, -1);
```
matz: I think `mrb_locale_from_utf8` function returns a string in the current locale converted *from* `UTF8`.  Isn't the variable name `utf8` inappropreate?

mattn: Ah, sure. will fix.

PR #3896, opened 2017-12-13
fix compilation error on VS2017

mattn:


matz:
I committed similar fix in 5b304c1

PR #3907, opened 2017-12-20
let mruby-math compile with MRB_WITHOUT_FLOAT

dabroz:
This may seem silly as inclusion of the gem does nothing, but it allows to use default gembox with `MRB_WITHOUT_FLOAT`.


ebkalderon:
Would you mind elaborating why this PR was closed? It appears that `mruby-math` still cannot compile with `MRB_WITHOUT_FLOAT` set in the default gembox.

Issue #3917, opened 2017-12-31
safe and portable MRB_METHOD_TABLE_INLINE


miura1729:
MRB_METHOD_INLINE makes mruby effective for speed and memory.  (https://github.com/mruby/mruby/commit/8f2c62407c0659d84126545e19505a851059e750) But MRB_METHOD_INLINE is machine depend and no check system. I propose safe and portable MRB_METHOD_INLINE. (https://github.com/miura1729/mruby/commit/f65157e7810095419096d91079c3fdb245aaab32)
 Key idea of this patch is prepare a static variable to put the address of the function and store variable address in method table.This enable using 2bits as flag because the address of the variable is alined.
To make declaring and initializing static variable easy, I redefine mrb_define_method, mrb_define ... as macro.

matz:
Please make a pull request instead of submitting an issue.

miura1729:
I send pull request. Please check.

PR #3918, opened 2018-01-01
Safe METHOD_INLINE_TABLE

miura1729:
see https://github.com/mruby/mruby/issues/3917


matz:
I addressed the portability issue by 83f2654 that requires `--align-functions=n (n>1)`. It's not portable but it works where you can specify function alignment.

miura1729:
Oh! I didn't know --align-functions=n  . Using this option is better than my idea.

PR #3921, opened 2018-01-04
Should be true for close_on_exec flag

ksss:
`close_on_exec` flag of duplicated object should be true.

And I think, It is not need that close fd when failed.
Because `fd` and `fd2` will close by GC on `fptr_finalize`.

### DIFF
--- a/mrbgems/mruby-io/src/io.c
+++ b/mrbgems/mruby-io/src/io.c
@@ -180,7 +180,7 @@ mrb_io_flags_to_modenum(mrb_state *mrb, int flags)
   return modenum;
 }
 
-void
+static void
 mrb_fd_cloexec(mrb_state *mrb, int fd)
 {
 #if defined(F_GETFD) && defined(F_SETFD) && defined(FD_CLOEXEC)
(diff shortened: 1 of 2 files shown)


mrbgems/mruby-io/src/io.c
```diff
@@ -573,15 +575,8 @@ mrb_io_initialize_copy(mrb_state *mrb, mrb_value copy)
...
-  if (failed) {
-    mrb_sys_fail(mrb, 0);
-  }
-  fptr_copy->fd2 = mrb_dup(mrb, fptr_orig->fd2, &failed);
-  if (failed) {
-    close(fptr_copy->fd);
-    mrb_sys_fail(mrb, 0);
-  }
+  fptr_copy->fd = mrb_cloexec_dup(mrb, fptr_orig->fd);
+  fptr_copy->fd2 = mrb_cloexec_dup(mrb, fptr_orig->fd2);
```
matz: The old code prevents fd leak.

ksss: Isn't GC enough to close fd?

matz: I think it's good to prevent resource leaks as early as possible.

ksss: OK, I fixed it.

PR #3923, opened 2018-01-08
Shouldn't use tempnam

ksss:
This warning was displayed when building.

```
/Users/ksss/src/github.com/ksss/mruby/mrbgems/mruby-socket/test/sockettest.c:13:15: warning: 'tempnam' is
      deprecated: This function is provided for compatibility reasons only. Due to security concerns
      inherent in the design of tempnam(3), it is highly recommended that you use mkstemp(3) instead.
      [-Wdeprecated-declarations]
  char *tmp = tempnam(NULL, "mruby-socket");
              ^
/usr/include/stdio.h:395:1: note: 'tempnam' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security conc...
^
/usr/include/sys/cdefs.h:180:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((deprecated(_msg)))
```

### DIFF
--- a/mrbgems/mruby-socket/test/sockettest.c
+++ b/mrbgems/mruby-socket/test/sockettest.c
@@ -2,18 +2,49 @@
 #include <stdlib.h>
 
 #include "mruby.h"
+#include "mruby/error.h"
 
-#ifdef _WIN32
-  #define tempnam _tempnam
+#if defined(_WIN32) || defined(_WIN64)
+
+#define close _close
+#define unlink _unlink
+
+#ifdef _MSC_VER
+static int
+mkstemp(char *p)
+{
+  int fd;
+  char* fname = _mktemp(p);
+  if (fname == NULL)
+    return -1;
+  fd = open(fname, O_RDWR | O_CREAT | O_EXCL, _S_IREAD | _S_IWRITE);
+  if (fd >= 0)
+    return fd;
+  return -1;
+}
+#endif
+
+#else
+
+#include <unistd.h>
+
 #endif
 
 mrb_value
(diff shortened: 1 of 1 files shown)


matz:
This change does not make the test secure. 

We used `tempnam` to avoid potential file name crash. `tempnam` is not encouraged because it is possible to guess scrambled file name before opening it. Your fix does not scramble at all, that means the attacker does not needs to guess.

We may need to fix the code to remove `tempnam`, but not this way.

ksss:
OK, I fixed to use `mkstemp` like mruby-io/test

matz:
Security-wise, it's not ideal (as bad as `mruby-io` test), but acceptable.

mrbgems/mruby-socket/test/sockettest.c
```diff
@@ -2,18 +2,49 @@
 #include <stdlib.h>
 
 #include "mruby.h"
+#include "mruby/error.h"
+
+#if defined(_WIN32) || defined(_WIN64)
+
+#define close _close
+#define unlink _unlink
```
mattn: please define these only for _MSC_VER.

mrbgems/mruby-socket/test/sockettest.c
```diff
@@ -2,18 +2,49 @@
...
+
+#ifdef _MSC_VER
```
mattn: So please move up this before `#define close _close`.

mattn:
I'll send fix in later.

matz:
I fixed it in f7ff2d0e

ksss:
Thank you for fixing.

PR #3936, opened 2018-01-28
mruby-time: support time_t is uint

ken-mu:
# Background

Some embedded platforms support `time_t` as `uint32`. Following problems are occurred if these platforms use current `Time` :
* `Time.gm`, `Time.local`, `Time.new`: ArgumentError is raised if more than 2038/1/19 3:14:8 is input (In case of `time_t` is `uint32_t`)

# Changes

* Add new macro `MRB_TIME_T_UINT`
* uint specific value is returned if `MRB_TIME_T_UINT` is defined

# Test

Define macro `MRB_TIME_T_UINT` on the platform `time_t` is uint like following:
```
conf.cc.defines = %w(MRB_TIME_T_UINT)
```

test code is following (test code is not pushed because following is not portable):
```
assert('Time.at lower than 1970 (for time_t is uint)') do
  assert_raise(ArgumentError){ Time.at(-1) }
end

assert('Time.at 2106/2/7 6:28:15 (for time_t is uint32_t)') do
  t = Time.at(4294967295)
  assert_equal t.year, 2106
  assert_equal t.mon, 2
  assert_equal t.mday, 7
  assert_equal t.hour, 6
  assert_equal t.min, 28
  assert_equal t.sec, 15
end

assert('Time.gm 2106/2/7 6:28:15 (for time_t is uint32_t)') do
  t = Time.new(2106, 2, 7, 6, 28, 15)
  assert_equal t.year, 2106
  assert_equal t.mon, 2
  assert_equal t.mday, 7
  assert_equal t.hour, 6
  assert_equal t.min, 28
  assert_equal t.sec, 15
end

assert('Time.new 2106/2/7 6:28:15 (for time_t is uint32_t)') do
  t = Time.gm(2106, 2, 7, 6, 28, 15)
  assert_equal t.year, 2106
  assert_equal t.mon, 2
  assert_equal t.mday, 7
  assert_equal t.hour, 6
  assert_equal t.min, 28
  assert_equal t.sec, 15
end

assert('Time.at higher than 2106/2/7 6:28:15 (for time_t is uint32_t)') do
  assert_raise(ArgumentError){ Time.at(4294967296) }
end

assert('Time.new higher than 2106/2/7 6:28:15 (for time_t is uint32_t)') do
  t = Time.new(2106, 2, 7, 6, 28, 16)
  assert_equal t.year, 1970
  assert_equal t.mon, 1
  assert_equal t.mday, 1
  assert_equal t.hour, 0
  assert_equal t.min, 0
  assert_equal t.sec, 0
end
```


matz:
`mktime` is supposed to return `time_t` and `-1` for errors. How does it behave on platforms with unsigned `time_t`?

ken-mu:
I had not understand why range error is raised when time is -1.
I think C language mktime need to return -1 if calender time cannot be represented even if `time_t` is `uint`, so I removed `ifdef` from above error handling.

ken-mu:
On our platform with unsigned `time_t`, no error is returned.
e.g. If 2000/13/1 00:00:00 is input with `mktime`, returned 978307200 (2001/1/1 00:00:00).
I think above behavior is not satisfied C language specification, 
so now we are contacting the manufacturer of the platform.

ken-mu:
Above test method was not correct.
I had misunderstood about the meaning of "cannot be represented".

I tested again and confirmed returning -1 if `mktime` cannot convert from `tm` to `time_t`.

PR #3945, opened 2018-02-10
Fix core dependencies resolving in dependencies gems.

take-cheeze:


matz:
Can you explain about your "core dependencies resolving in dependencies gems" issue bit more?

take-cheeze:
`:core` dependencies resolving before `:mgem` deps only works on `build_config.rb`.
With this PR `:core` deps will be resolved in `mrbgem.rake` too.

matz:
OK, sounds reasonable.

PR #3956, opened 2018-02-23
Native support of mruby-cli in mruby build system.

take-cheeze:
Just for a proof of concept I've rewritten `mruby-bin-strip` only with ruby script for `mruby-cli` feature.
For `mrb_irep` section modifications, `MRubyVM` (subset of `RubyVM` in CRuby) is implemented too.

Since [mruby-cli](https://github.com/hone/mruby-cli) isn't catching up with recent mruby releases, we should support this with mrbgem build system.
~~ This is not complete because I need to add minimal dependency `mrb_open` for better testings too.  Symbol like `mrb_open_cli` must be generated. ~~
Now implemented: https://github.com/mruby/mruby/pull/3956/commits/34a3939936ef17aee5e1a06276fa72688823d2de

See https://github.com/mruby/mruby/pull/1676 for implementation detail.


include/mruby/class.h
```diff
@@ -89,6 +89,11 @@ void mrb_gc_mark_mt(mrb_state*, struct RClass*);
 size_t mrb_gc_mark_mt_size(mrb_state*, struct RClass*);
 void mrb_gc_free_mt(mrb_state*, struct RClass*);
 
+extern mrb_int
```
take-cheeze: Ah, sorry these corresponds to those in CRuby:
https://github.com/ruby/ruby/blob/421a73f51a39ff6829f9715b423a928782ec8ec3/vm.c#L339-L341
And it should be `mrb_state` specific.

matz:
The proposed change sounds reasonable. It is OK to change the build system and writing command line tools in mruby.

But I don't want to add members to `mrb_state` unless they are absolutely necessary.

include/mruby/class.h
take-cheeze: Maybe this should be optional like VM hooks.

matz:
Thank you for this PoC PR. Closing it 8 years on.

Per the 2018 conversation, you noted you'd split this into 3 follow-up PRs once the direction was approved:
1. Build system changes → opened as #3976 (also being closed)
2. `mruby-mrubyvm` gem (`RubyVM` class for mruby)
3. Switch `mruby-strip` to a mruby script

Of those, only the build-system split materialized; the other two were never split out. None of the three found follow-up activity in the 8 years since.

Master has gained alternative paths that cover similar use cases:

- **`mruby-bin-mrb`**: a compiler-free runtime executor. Lets users distribute CLI tools as `.mrb` bytecode files alongside this small executor — covers the "CLI tool written in Ruby" case without needing build-system support for embedded bytecode.
- **Shell-script-style CLI gems** like `mruby-bin-config`: when a CLI tool is mostly thin wrappers, a generated shell/batch script (`mruby-config`) avoids needing C or embedded bytecode at all.

Closing in favor of those paths plus the lack of a champion. Thanks again for the original exploration.

PR #3957, opened 2018-02-23
Pass same mode parameter to `IO.sysopen` and `IO.open` in `IO.read`.

take-cheeze:
Since `IO.read filename, mode: 'rb'` won't work in windows .


tsahara:
LGTM.  Can I merge it? @matz

PR #3963, opened 2018-03-04
Give me block by mrb_get_args() (forced block arguments)

dearblue:
I added feature to ``mrb_get_args()`` function for forced block argument.

By giving two consecutive ``&``, exception is raised if block is not given.

```c
/* raise exception if no block given */
mrb_get_args(mrb, "&&", &block);
```

I tried to labor saving in the processing that required block. How is it?


dearblue:
``!`` is used by other specifiers with nil.

I thought that consistency could be maintained if it was ``&&``, and I did so.

I will fix it.

matz:
There's non-nil usage of `!` e.g. `*!`. Reverse meaning of `&!` from `s!` or `a!` might be a drawback.

dearblue:
Oops.

I just read in mruby.h document only for ``mrb_get_args()``.

I understand by your comment about ``!``.

matz:
Oops, I should have updated mruby.h document too.

PR #3976, opened 2018-03-22
Add support of CLI in mrbgem build system.

take-cheeze:
See: #3956
`@mrb_open_cli` is added to check whether to generate `mrb_open_cli` function or to use `mrb_open`.


matz:
Thank you for this PR — the build-system half of the split you proposed in #3956. Closing it 8 years on.

The build-system change was approved in concept by Matz in 2018 ("OK to change the build system and writing command line tools in mruby"), but the PR has sat without follow-up activity since 2019, and the master codebase has moved substantially:

- The `Rakefile` / `lib/mruby/gem.rb` API the diff references has evolved (the `tsort_dependencies`, `generate_gem_table`, gem-table iteration patterns are different now). A re-base from this branch would be substantial rework.
- Master has adopted lighter-weight alternatives for the "Ruby-script CLI tool" use case:
  - `mruby-bin-mrb` (compiler-free runtime executor) covers the `.mrb`-bytecode-distribution case without embedded-bytecode build infrastructure.
  - Shell-script generators (`mruby-bin-config`) cover the thin-wrapper case.
- No current `mrbgems/mruby-bin-*` gem has migrated its tool to `.rb`, indicating no demand from existing tools.

The infrastructure addition is small (~120 lines) but maintaining it for a use case that has no current consumers and no champion is hard to justify after 8 years. If a future need emerges (e.g. a contributor wants to ship a substantial Ruby-script CLI tool), it would be cleaner to design fresh against today's `Rakefile`/gem.rb than to revive this branch.

Closing as obsoleted by 8 years of evolution. Thanks again for the original work.

PR #3979, opened 2018-03-22
Add comment of opcode value.

take-cheeze:
For debugging VM on debuggers.


matz:
I plan to overhaul opcode structure. So the code above will be removed anyway. I will improve here, by adding enum name for opcode so that we can cast to get opcode names e.g. `p (enum mrb_opcode)0 #=> OP_NOP`.

PR #3983, opened 2018-03-25
Fix test gem script.

take-cheeze:
To fix mruby-cfunc test build and others.


matz:
Both commits are covered by b37b31208 and 633599a2c respectively.

PR #3984, opened 2018-03-25
Catch up with latest mruby-socket.

take-cheeze:
https://github.com/iij/mruby-socket/commit/36d016afd663d34607c843d03371bedb71efa34e


tsahara:
This should be reverted.  The change was [incorporated into mruby/mruby first](https://github.com/mruby/mruby/pull/3891) then [I have backported it to iij/mruby-socket](https://github.com/iij/mruby-socket/commit/36d016afd663d34607c843d03371bedb71efa34e).  There is no reason to set `hints.ai_protocol` twice here.

take-cheeze:
Ah sorry will revert this. That's my mistake of subtree merging.

PR #4003, opened 2018-04-17
Fallback to float when compiled binary with 64bit compiler.

### LINKED ISSUE #3997: In `MRB_INT32` enabled build, test fails.
Seems fallback to float isn't working because of `MRB_WITHOUT_FLOAT`.

This assertion get exception raise:
https://github.com/mruby/mruby/blob/1905091634a6a2925c911484434448e568330626/test/t/integer.rb#L161

From: https://github.com/mruby/mruby/blob/1905091634a6a2925c911484434448e568330626/src/string.c#L2170

take-cheeze:
closes #3997.
In compiler built with `MRB_INT64`, it treats integer as `Fixnum` that doesn't fit to `int32_t`.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -126,9 +126,10 @@ read_irep_record_1(mrb_state *mrb, const uint8_t *bin, size_t *len, uint8_t flag
       }
       src += pool_data_len;
       switch (tt) { /* pool data */
-      case IREP_TT_FIXNUM:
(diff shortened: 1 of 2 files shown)


matz:
I think we need to update the mrbc format not to depend on the `mrb_int` size of the platform in the future.

PR #4011, opened 2018-04-26
Fix wrong the number of marked objects of MRB_TT_ENV in gray mark phase

kou:
If MRB_TT_ENV has stack shared flag or bidx flag, flags is too large
than the real stack size.


kou:
We should have done this in 93f5f225772c398be6e409da3d3ef0f07ffbe1cf.
We may have more similar missing changes.

kou:
Thanks for confirming it.

PR #4012, opened 2018-04-27
Fix heap use after free on mruby-aws-sigv4.

take-cheeze:
https://103.128.221.202.static.iijgio.jp/jenkins/view/0-master-mrbgem/job/master-mruby-aws-sigv4/37/consoleFull
BTW is modifying `self` value of env OK?


matz:
It's OK for this case.

PR #4013, opened 2018-04-27
Fix Enumerable#hash on non 32-bit integer conf.

kimushu:
Enumerable#hash is implemented on enum.rb with hard-coded range (0x7fffffff).
This does not work on the environments with small integer (31-bit or less).

This patch redefines Enumerable#hash on C side.

### DIFF
--- a/mrblib/enum.rb
+++ b/mrblib/enum.rb
@@ -333,16 +333,4 @@ module Enumerable
   #
   # ISO 15.3.2.2.20
   alias to_a entries
-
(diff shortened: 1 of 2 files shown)


matz:
I merged and updated the patch. I fixed it not to use `mrb_funcall` which cause recursive calls to VM.

matz:
I didn't run tests for `MRB_INT16` environment. If I missed anything, let me know.

PR #4017, opened 2018-05-02
Add `-d` option for `mruby` and `mirb`.

mimaki:


matz:
The CI build generated a few warnings and errors. Can you check?
Some of them might be caused by previous patches.

PR #4022, opened 2018-05-07
Add `-r` option for `mruby` and `mirb`. (mrbgem is not supported)

mimaki:


matz:
First of all, why do we need `-r` for `mirb` and `mruby`? I can imagine `mirb` can load its initialize file before the session start, but have no idea for `mruby`.
Second, test suites failed. We need to pass tests before merging it.

mimaki:
In the current **mruby** interpreter, **require** is not supported.
If the library and application source files are split, I need to merge the source code.

When trying various libraries ,it is convenient to have functions that can be used without merging into source code.
**-r** is a kind of **require** of Ruby.

Sorry for errors, I fixed them.

matz:
OK, I understand.

tsahara:
:+1:

PR #3991, opened 2018-04-04
Fix possible heap use after free in `mrb_exec_irep` and stack expanding.

take-cheeze:
Since `mrb_exec_irep` doesn't overwrite `ci->env` it may causes use after free error with `MRB_GC_STRESS`, address sanitizer, [iij/mruby-require](http://github.com/iij/mruby-require).
To fix this `REnv` of `RProc` associated to callinfo needs to be adjusted too.

```
require loop check : .                                                                                                                                                                                                                                                                  
require nest : =================================================================                                                                                                                                                                                                        
==14524==ERROR: AddressSanitizer: heap-use-after-free on address 0x61d000047948 at pc 0x000000490a16 bp 0x7fff491a7440 sp 0x7fff491a7430                                                                                                                                                
READ of size 4 at 0x61d000047948 thread T0                                                                                                                                                                                                                                              
    #0 0x490a15 in gc_mark_children /home/takeshi/dev/mruby/src/gc.c:666                                                                                                                                                                                                                
    #1 0x492bbc in gc_gray_mark /home/takeshi/dev/mruby/src/gc.c:914                                                                                                                                                                                                                    
    #2 0x49359e in incremental_marking_phase /home/takeshi/dev/mruby/src/gc.c:1009                                                                                                                                                                                                      
    #3 0x4948af in incremental_gc /home/takeshi/dev/mruby/src/gc.c:1122                                                                                                                                                                                                                 
    #4 0x4949c7 in incremental_gc_until /home/takeshi/dev/mruby/src/gc.c:1147                                                                                                                                                                                                           
    #5 0x495986 in mrb_full_gc /home/takeshi/dev/mruby/src/gc.c:1247                                                                                                                                                                                                                    
    #6 0x48ed47 in mrb_obj_alloc /home/takeshi/dev/mruby/src/gc.c:513                                                                                                                                                                                                                   
    #7 0x4b423b in str_new /home/takeshi/dev/mruby/src/string.c:61                                                                                                                                                                                                                      
    #8 0x4b71e6 in byte_subseq /home/takeshi/dev/mruby/src/string.c:408                                                                                                                                                                                                                 
    #9 0x4b75ee in str_substr /home/takeshi/dev/mruby/src/string.c:453                                                                                                                                                                                                                  
    #10 0x4be49c in mrb_str_aref /home/takeshi/dev/mruby/src/string.c:1053                                                                                                                                                                                                              
[...]

PR #4033, opened 2018-06-03
fix env->stack misadjusting: fix #4029

### LINKED ISSUE #4029: local variables gets broken
I'm totally not sure what's the real cause of this problem, but anyway something seems to be broken with proc and cyclic hash.

```ruby
class Foo
  def doit
    hash = { "foo" => nil }
    hash["foo"] = hash
  end
end

foo = Foo.new
1.times {
  puts foo.doit
  p foo
}
```

I expected `p foo` outputs like `#<Foo:0x7f936cbfe2b0>`, but the results are different, and even worse, different everytime.

```
$ build/host/bin/mruby repro.rb
{...}
"foo"
$ build/host/bin/mruby repro.rb
{...}
#<Proc:0x7ffc57834f60@/Users/i110/mruby/mrblib/enum.rb::61>
$ build/host/bin/mruby repro.rb
{...}
#<Proc:0x7f8093053190@/Users/i110/mruby/mrblib/hash.rb::199>
$ build/host/bin/mruby repro.rb
{...}
[]
$ build/host/bin/mruby repro.rb
{...}
[["foo", {...}]]
$ build/host/bin/mruby repro.rb
{...}
#<Proc:0x7f96b5053190@/Users/i110/mruby/mrblib/hash.rb::199>
```

What's happening?

i110:
Currently `envadjust` checks `oldbase <= env->stack < oldbase + newsize`, but it's wrong - we should use oldbase + **oldsize** as the upper bound.
After https://github.com/mruby/mruby/pull/3991, in one `envadjust` call, an `env `can be checked 2 times (when `ci[0]->env` and `ci[1]->proc->e.env` are identical). If newbase is allocated immediately after oldbase, `env->stack` pointer accidentally gets adjusted 2 times and refers out of range of the new stack area.

To reproduce, consume large amount of stack and then refer some lexical variable (`x`) in a closure.
```ruby
def doit(depth)
  return if depth == 0
  v0, v1, v2, v3, v4, v5, v6, v7, v8, v9 = (1..10).map { 'BROKEN' }
  doit(depth - 1)
end

x = 'OK'
proc {
  doit(100)
  puts x
}.call
```

$ seq 10 | xargs  -n 1 build/host/bin/mruby repro.rb
```
OK
OK
OK
OK
OK
OK
BROKEN
OK
BROKEN
OK
```

As far as I checked, OSX seems to often reallocate a new buffer immediately after old buffer, but in Linux it's not so often.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -141,7 +141,7 @@ stack_init(mrb_state *mrb)
 }
 
 static inline void
-envadjust(mrb_state *mrb, mrb_value *oldbase, mrb_value *newbase, size_t size)
+envadjust(mrb_state *mrb, mrb_value *oldbase, mrb_value *newbase, size_t oldsize)
 {
   mrb_callinfo *ci = mrb->c->cibase;
 
@@ -151,7 +151,7 @@ envadjust(mrb_state *mrb, mrb_value *oldbase, mrb_value *newbase, size_t size)
     mrb_value *st;
 
     if (e && MRB_ENV_STACK_SHARED_P(e) &&
-        (st = e->stack) && oldbase <= st && st < oldbase+size) {
+        (st = e->stack) && oldbase <= st && st < oldbase+oldsize) {
       ptrdiff_t off = e->stack - oldbase;
 
       e->stack = newbase + off;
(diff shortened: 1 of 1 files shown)


matz:
Thank you!

PR #4034, opened 2018-06-04
Let inspect recursion do the right thing

i110:
There are some problems around `inspect` of recursive objects.

- Recursive hash only generates "{...}"
`{ "foo" => <hash itself> }` generates `{...}`, but should be `{"foo"=>{...}}`

 - Recursive struct has the same problem as hash

- Recursive array generates very long string
`[ true, <array itself> ]` generates `[true, [true, [true, .... [..]]]]] ... ]`. We want `[true, [...]]` instead

These behaviors are also bad from point of view of performance due to unnecessary tree traversing (especially recursive array consumes much more time and memory). Furthermore, we want to distinguish real SystemStackError from recursive inspection, but currently that error is hidden.


matz:
Thank you for the patch. The reason behind the current behavior is memory consumption. Recursion record can be very big for object trees. I am not sure whether we should choose the right behavior at the cost of memory.

i110:
Thank you for the explanation. I'm not sure too which is good because I have no idea of memory restricted environments. But at least, I think Array#inspect should be fixed so that it behave as same as Hash's one (i.e. returns `[...]`) for consistency and performance.

For what it's worth, I just took a simple profile.

```ruby
WIDTH, HEIGHT, PR = ARGV[0].to_i, ARGV[1].to_i, ARGV[2]
puts "# of hashes: %d" % [(WIDTH ** (HEIGHT + 1) - 1) / (HEIGHT - 1)]

def gen(depth)
  hash = {}
  return hash if depth == 0
  WIDTH.times {|i| hash[i] = gen(depth - 1) }
  hash
end

h = gen(HEIGHT)
recur = {}
str = PR ? h._inspect(recur) : h.inspect
puts "recur.size: #{recur.size}"

GC.start
$stdin.gets
```
This script construct Hash Tree using specified witdh and depth, and before it exits I manually hit the following command to see RSS (because ObjectSpace doens't have `memsize_of`..)
```
ps aux | grep "[m]ruby"
```

The result is as follows:

| layout (WxH) | 6x6 | 7x6 | 8x6 | 6x7 | 7x7 | 6x8 |
|--------------|-----|-----|-----|-----|-----|-----|
| # of hashes | 55987 | 137257 | 299593 | 335923 | 960800 | 2015539 |
| recursion record size | 9331 | 22876 | 49932 | 55987 | 160133.1667 | 335923 |
| master RSS (KB) | 30892 | 64960 | 109628 | 175980 | 396540 | 791620 |
| PR RSS (KB) | 31628 | 65788 | 114344 | 181164 | 403016 | 809756 |
| ratio | 102.38% | 101.27% | 104.30% | 102.95% | 101.63% | 102.29% |

So roughly 1-4% momentary overhead are there I think. umm.

Issue #4021, opened 2018-05-07
Broken lexical variables in eval


i110:
At latest master (https://github.com/mruby/mruby/commit/72581696d35e8da79a3a9d606294602c83abdf9f), the following code says nothing (i.e. proc returns nil) but of cause I expected `FOO`. 

```ruby
puts eval(<<'EOS').call
foo = 'FOO'
proc { foo }
EOS
```

What's more interesting is,

```ruby
def do_eval(code)
  eval(code)
end

puts do_eval(<<'EOS').call
foo = 'FOO'
proc { foo }
EOS
```
says the following, eval-ed code itself!
```
foo = 'FOO'
proc { foo }
```

i110:
After some investigations, I found https://github.com/mruby/mruby/pull/4012 is the cause of this issue. By setting `ci->env` is set to proc's env which encloses the eval-caller's context, inner closure refers that env too (https://github.com/mruby/mruby/blob/master/src/proc.c#L69). But that `env->stack` should point `proc` caller's stack which is placed at the next of actual one.

@take-cheeze Could you explain the detail of that PR? I cannot see https://103.128.221.202.static.iijgio.jp/jenkins/view/0-master-mrbgem/job/master-mruby-aws-sigv4/58/console now.

take-cheeze:
@i110 The behavior you expect seems more correct.
Since https://github.com/mruby/mruby/pull/3991 is merged #4012 can be reverted.
I'll check aws-sigv4 after reverted.

i110:
@take-cheeze Thank you. I filed https://github.com/mruby/mruby/pull/4039. BTW just for reference, could you give, me brief explanations of why you added these code, when you have a time?

take-cheeze:
@i110 there was some memory access error and seems there is another happening.
will investigate later.

PR #4039, opened 2018-06-06
fix broken lexical variables

i110:
As discussed in https://github.com/mruby/mruby/issues/4021, the commit 26e436e24797f0c3228bc9900615afe7d2e29ddf caused problems around `ci->env`. If I understand correctly, there're no reasons to set it in there.


take-cheeze:
@i110 Could you add the test code too?

i110:
okay, I'm searching where is the best place to add the tests under test/t/ ..

take-cheeze:
@i110 you should place it to `mruby-eval` mrbgem if you use `eval`

i110:
@take-cheeze I added the test at https://github.com/mruby/mruby/pull/4039/commits/1dbeda5c9cfe60b0dbdf94a8fea7e12a721ff162

PR #4051, opened 2018-06-20
Make minirake parallel.

take-cheeze:
`sh` won't block until specific number of child processes with this change.
Needs #4050 to be merged.

Adds `-j` / `--jobs` option like `make` command.


matz:
Unfortunately, CI fail.

take-cheeze:
@matz I've rebased with #4050 . Seems like appveyor fails due to resource locking so keeping it with single jobs.

PR #4056, opened 2018-06-21
Fix job waiter of minirake.

take-cheeze:
Seems job limiting wasn't working.


matz:
This patch makes `minirake -j` 2+ times slower. Actually, minirake seems not to work in parallel at all after this patch.

PR #4066, opened 2018-07-04
Added ~/Android/Sdk/ndk-bundle as default NDK home

duangsuse:
This PR adds _more recent_ default AndroidStudio Linux Android NDK to default NDK search path

Since only one line of code added, I didn't create a new branch

No modification to core/mrbgems, so no test case needed

:beer:


matz:
I don't use Android SDK, so I cannot decide if this is sufficient. If anyone sees the problem, I will revert this.

PR #4068, opened 2018-07-09
add mrbc option `--remove-lv`

yurie:
* refactor: move `irep_remove_lv` from `mruby-bin-strip` gem to src/dump and rename to `mrb_irep_remove_lv`
* add: mrbc option `--remove-lv` to remove LVAR section

### DIFF
--- a/include/mruby/dump.h
+++ b/include/mruby/dump.h
@@ -31,6 +31,7 @@ MRB_API mrb_value mrb_load_irep_file(mrb_state*,FILE*);
 MRB_API mrb_value mrb_load_irep_file_cxt(mrb_state*, FILE*, mrbc_context*);
 #endif
 MRB_API mrb_irep *mrb_read_irep(mrb_state*, const uint8_t*);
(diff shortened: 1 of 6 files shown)


matz:
The fundamental idea seems OK but is `src/dump.c` is the best place for `mrb_irep_remove_lv`?
It seems unnecessary for most of the cases. Maybe in `codegen.c`?

yurie:
Updated.
Now declaration of the function is in `dump.h`. Should I move to `irep.h` or other file?

matz:
`irep.h` seems better. Besides that, should we provide an option to remove all debug information altogether, not just lv?

mimaki:
I agree to remove debug information.

2018-07-10 16:30 GMT+09:00 matz <notifications@github.com>:

yurie:
moved to `irep.h`.

Summary of output debug information of `mrbc` with this patch:

 * (no option): `IREP`, `LVAR` sections
 * `-g` option: `IREP`, `DBG`, `LVAR` sections
 * `--remove-lv` option: `IREP` sections

PR #4076, opened 2018-08-04
Add built-in symbols feature (EXPERIMENTAL!)

dearblue:
\# Most English text is machine translation by Google translation. The Japanese text which becomes the original text is also written.

## Summary

Add built-in symbols embedded in object code at build time.
<br>ビルド時にオブジェクトコードへと埋め込まれる組み込みシンボルを追加します。

## Description

  - Allocate strings as built-in symbols and arrays of symbol structures (`symbol_name`) as `static const` at build time.
    <br>組み込みシンボルとしての文字列、及びシンボルの構造体 (`symbol_name`) の配列を `static const` としてビルド時に確保します。

  - Collect the character strings as symbols. At this time, small strings matching the end of the large string are merged.
    <br>シンボルとなる文字列をまとめます。このとき大きい文字列の末端に一致する小さい文字列は統合されます。

    e.g.:

      - "Object" is included in "BasicObject"
      - "RUBY_VERSION" is included in "MRUBY_VERSION"
      - "cmp" and "p" are included in "__ary_cmp"

  - If you create multiple `mrb_state`, you can share embedded symbols.
    <br>複数の `mrb_state` を作成した場合、組み込みシンボルを共有することが出来ます。

  - The symbols other than the built-in symbols (I call it runtime symbols at will) are placed in the `struct symbol_name` allocated in the heap as before.
    <br>組み込みシンボル以外のシンボル (僕は勝手に実行時シンボルと呼んでいます) は、これまで通りヒープに確保された `struct symbol_name` に置かれます。

  - Although the `.text` + `.rodata` section increases by about 10 KB, measurement by `valgrind mruby - e ''` decreased malloc from 181.6 KB to 147.8 KB.
    In mrbtest, malloc has decreased by about 1.8 MB.
    <br>`.text` + `.rodata` セクションが約 10 KB 増えますが、`valgrind mruby -e ''` による計測では malloc が 181.6 KB から 147.8 KB へと減少しました。
    mrbtest においては、malloc が約 1.8 MB 減少しました。

### Compatibility

  - The value of `mrb_sym` reaches the most significant bit.
    <br>`mrb_sym` の値が最上位ビットに到達します。

  - There are no changes to functions, macros, or structures that lose incompatibility.
    <br>互換性を失う関数やマクロ、構造体の変更はありません。

### Appended configurations

  - `MRB_ENABLE_BUILTIN_SYMBOLS`
[...]

PR #4078, opened 2018-08-06
put symbol definitions into ROM with gperf

yurie:
This is a patch to put symbol definitions into ROM, as I talked in RubyKaigi 2018.

This patch generates perfect hash function with gperf instead of khash.
With the patch, pre-defined symbols are in ROM area, not in khash struct in RAM, so mruby consume less RAM memory in execution time for embedded environments.

### Usage

Define macro `USE_PRESYM` in build_config.rb.

```ruby
  conf.cc.defines = %w(USE_PRESYM)
```

Without `USE_PRESYM`, nothing changed.

### struct and functions

The patch add a struct and two functions in include/mruby/presym.h.

* `struct name2presym` for name and symbol number pairs.
* `const struct name2presym *presym_find(const char *, unsigned int);` returns name2presym from name and its length.
* `const char *presym_sym2name(uint32_t sym);` returns pointer to symbol name from symbol number.

These functions are defined in src/presym.c, that is generaged by gperf.

### Rake tasks

We add Rake tasks to update src/presym.c.

* `rake presym:build` scan `src/*.c` and `mrblib/*.rb`, and update src/presym.c

`presym:build` execute 3 sub tasks. This task geneartes src/presym.json as imidiate list of symbols.
You can edit presym.json to add other symbols you may use.

* `rake presym:gen_json` find symbols in `src/*.c` and `mrblib/*.rb`, and generate JSON file.
* `rake presym:gen_key` load JSON file and generate presym.key for gperf.
* `rake presym:gen_c` generate presym.c from presym.key with gperf.

(Note: we will make another PR to put method table into ROM.)


matz:
@yurie @takahashim It is very interesting we have another pull request to address a similar problem.
What do you think about #4076?

dearblue:
Since I can neglect the work of #4076 any more, I would like to be unified into this #4078!
It is better that the khash table is integrated as well.

   - For decrease the amount of RAM
   - For the amount of increase in ROM
   - For speed reduction

I would like a comparison it.

matz:
FYI, `minirake` does not support `namespace`. We need to either remove `namespace` from the `presym` build, or hack `minirake` to support `namespace`.

matz:
Static (ROM-able) symbols are implemented in a different way in `mruby3` branch.

Issue #4108, opened 2018-09-06
Signed integer overflow and information leak in mrb_str_format


clayton-shopify:
https://hackerone.com/d4nny reported the following:

-----

Introduction
============

This is a Windows 32bit specific bug in mruby. The root cause is present in the code, but glibc's implementation on snprintf() avoids the bug in Linux build. Although I believe the integer overflow that causes this has to be fixed.

Proof of concept
================
```
print "%1.2147483647f"%3.14e45
```

Please look at the attached screenshot to see the info leak in action.

Root cause
===========

```
        fmt_setup(fbuf, sizeof(fbuf), *p, flags, width, prec);
        need = 0;
        if (*p != 'e' && *p != 'E') {
          i = INT_MIN;
          frexp(fval, &frexp_result);
          i = (mrb_int)frexp_result;
          if (i > 0)
            need = BIT_DIGITS(i);
        }
        need += (flags&FPREC) ? prec : 6;                 //integer overflow
        if ((flags&FWIDTH) && need < width)
          need = width;
        need += 20;
        if (need <= 0) {
          mrb_raise(mrb, E_ARGUMENT_ERROR,
                    (width > prec ? "width too big" : "prec too big"));
        }

        CHECK(need);
        n = snprintf(&buf[blen], need, fbuf, fval);
        if (n < 0) {
          mrb_raise(mrb, E_RUNTIME_ERROR, "formatting error");
        }
        blen += n;
```
The main bug is integer overflow in line 1059. Here `need` is the number of bytes needed to represent the floating point number in string format. This is being added with the `precision` which is in attackers control. On a 32bit build, we can pass `2147483647` as precision, which will do a signed overflow and became negative value. There is no integer overflow check here - which is the vulnerability.

if we set the width as `1`,  the condition on line 1060 will satisfy and `need` will become `1`.

Now at 1069, the snprintf() is called with need as `21` and format as `%1.2147483647f`. This is the place where visual C library and libc's implementations change - and that will become the condition whether we can see info leak.

In libc, the `.2147483647` precision value is considered too much and snprintf returns -`1`. So libc based builds a re not affected.

in visual C library, this is a valid precision and it prints 21 characters into the buffer (because that is the length we specified - `need`). Now the vulnerability happens when the return value `n`is added to the `blen` at line number 1073.

[...]

PR #4121, opened 2018-09-14
Prevent signed integer overflow.

clayton-shopify:
https://github.com/mruby/mruby/issues/4108 was addressed in 625976d7931006bae7e960e561ecdaccb0cb6a28, but signed integer overflows can still occur. I'm concerned that compilers might decide to optimize away the overflow checks, so I've rewritten the code in a way that should avoid signed integer overflows.

I also noticed that the overflow check in `GETNUM` was incorrect, so I fixed that as well.


matz:
Thank you!

PR #4125, opened 2018-09-22
Suffix support

ukrainskiysergey:
Some time ago I ported the suffix support from cruby to mruby.

In my implementation, it just replaces 5i with Kernel.Complex(0, 5), 5r with Kernel.Rational(5, 0) and 5ri with Kernel.Complex(0, Kernel.Rational(5, 0)).

For this reason, this implementation:
1. Does not affect the execution time.
2. Does not require any changes in bytecode, boxing or anything else.
3. Allows users to make their own implementation of Complex and Rational classes.

This pull request contains suffix support itself ("include/mrbconf.h" and "mrbgems/mruby-compiler/core/parse.y") and simple implementation of Complex and Rational classes for testing purpose (should be removed).

I do not think that this pull request should be merged into the main code base. But maybe someone will need this code, so I decided to share it.

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -58,6 +58,9 @@
 # endif
 #endif
 
+#define MRB_COMPLEX_NUMBERS
+#define MRB_RATIONAL_NUMBERS
+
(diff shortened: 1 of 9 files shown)


matz:
merged.

PR #4127, opened 2018-09-24
Fix comparisons in str_squeeze.

clayton-shopify:
Dinko Galetic & Denis Kasak noticed that strings beginning with `\xff` can cause a crash in `str_squeeze` because `s[0]` is treated as a signed value, causing it to be equal to `lastch`, which is -1. After this occurs, `i` is decremented and `s[-1]` is accessed.

The following inputs demonstrate crashes:
```ruby
"\xff00000000000000000000000".squeeze("\xff")
```
```ruby
"\xff00000000000000000000000".squeeze!("\xff")
```
```ruby
"\xff00000000000000000000000".squeeze
```
```ruby
"\xff00000000000000000000000".squeeze!
```
Changing `s` to an `unsigned char *` fixes the problem.

### DIFF
--- a/mrbgems/mruby-string-ext/src/string.c
+++ b/mrbgems/mruby-string-ext/src/string.c
@@ -543,7 +543,7 @@ str_squeeze(mrb_state *mrb, mrb_value str, mrb_value v_pat)
 {
   struct tr_pattern *pat = NULL;
   mrb_int i;
-  char *s;
+  unsigned char *s;
   mrb_int len;
   mrb_bool flag_changed = FALSE;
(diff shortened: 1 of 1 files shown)


matz:
The issue is addressed by 9e3cbaa along with #4130. Reverted.

dgaletic:
`str_squeeze` mishandles the `ÿ` character after the revert. It gets deleted if it's the first character in a string, e.g. `p "ÿ".squeeze` prints an empty string.

matz:
Ah, it doesn't work for ISO-8859-1 strings. Will be fixed soon.

Issue #4150, opened 2018-10-29
Embedding symbol index seems to be unnecessary


take-cheeze:
https://github.com/mruby/mruby/blob/eebb56713e27cde27a7b899da8636a0634704863/include/mruby/ops.h#L73-L83

matz:
They are used when forwarding to `OP_SEND`. I didn't want to call `mrb_intern` each time forwarding.

take-cheeze:
Is it possible to reserve symbols used in language features in the symbol table like:

| ID | String |
|---|---|
| 1 | `+` |
| 2 | `-` |
| 3 | `*` |
| ... | ... |
| 10 | `method_missing` |
| ... | ... |
| XXX | user defined symbols |
| ... | ... |

So that we can always access the reserved symbol with constant.

matz:
Ah, they will not be required when embedded symbols are available. OK, remove them from the `OP` code.

matz:
We can use `mrb_intern` for now.

take-cheeze:
Moving to #4151

PR #4151, opened 2018-10-29
Reduce instruction size

take-cheeze:
Implementation of #4150 .


take-cheeze:
Implementation should be done.

matz:
`mrb_reserved_symbols` in `symbol.h` seems like premature optimization. They will be overridden when embedded symbols come. Plain `mrb_intern` should be enough for the time being.

take-cheeze:
Even if embedded symbols come, I think these reserved symbols is needed for symbols that is used in language feature.
Also they could be replaced with macro when its optimization doesn't match.

And code generator could be optimized with it too.
(I didn't modified it this time.)

mrbgems/mruby-compiler/core/codegen.c
```diff
@@ -982,7 +983,9 @@ gen_call(codegen_scope *s, node *tree, mrb_sym name, int sp, int val, int safe)
     gen_move(s, cursp(), recv, 1);
     skip = genjmp2(s, OP_JMPNIL, cursp(), 0, val);
   }
-  idx = new_sym(s, sym);
+  if (!mrb_symbol_constsym_send_p(sym)) {
```
take-cheeze: Need to check `&& n == 1`.

PR #4180, opened 2018-12-06
Support lock file for git.

take-cheeze:
Inspired by https://udzura.hatenablog.jp/entry/2018/12/05/143801
Creates lock file to `"#{MRUBY_CONFIG}.lock`.
When pulled or `checksum_hash` is set should be updated with it.
Lock file looks like:
```yaml
---
builds:
  host:
    https://github.com/mattn/mruby-onig-regexp.git:
      url: https://github.com/mattn/mruby-onig-regexp.git
      branch: master
      commit: 354c0df670fe6822528d5a508fb7722233f4b920
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,12 +8,19 @@ MRUBY_BUILD_HOST_IS_OPENBSD = RUBY_PLATFORM.include?('openbsd')
 $LOAD_PATH << File.join(MRUBY_ROOT, "lib")
 
 # load build systems
+require 'yaml'
 require "mruby-core-ext"
 require "mruby/build"
 require "mruby/gem"
 
 # load configuration file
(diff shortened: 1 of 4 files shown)


take-cheeze:
Forgot to note that local repository won't be shared between different `MRuby::Build` with this too.

take-cheeze:
I want to use `-C` instead of `--git-dir` and `--work-tree` but seems like it's a new option: https://github.com/tensorflow/tensorflow/pull/4580

udzura:
I think this feature seems to be required by many many mrubyists!

take-cheeze:
TODO: Save version of mruby itself too

matz:
OK

Issue #4186, opened 2018-12-12
suspect legacy code of removed Hash's khash implementation and a question


jasl:
In mruby 2.0 release note, it said "Hash implementation replaced from khash",
but in current master branch, [include/mruby/khash.h](https://github.com/mruby/mruby/blob/master/include/mruby/khash.h) still there, and included in [include/mruby/hash.h#L11](https://github.com/mruby/mruby/blob/master/include/mruby/hash.h#L11) but it seems no longer in use (I used keyword `kh_ht` to search the repo), does it mean it can be cleaned?

And I have a question:

I'm trying to to upgrade [Shopify ESS](https://github.com/Shopify/ess) to mruby 2.0  but blocked by the change of Hash.

(Disclamation: I'm not staff of Shopify, just a fanboy, and not know too much in C),

In previous mruby, `kh_ht` relates functions are in `hash.h` so [ESS can use them to fetch keys and values and the size of whole Hash object](https://github.com/Shopify/ess/blob/master/ext/enterprise_script_service/data.cpp#L152-L173), but in new implementation, the Hash's low-level APIs are hidden in `hash.c`, is there any equivalent way to do the job in mruby 2.0?

jasl:
And mruby 2.0 don't have an API to get the Hash's size currently

jasl:
Looks `mrb_hash_foreach` and #4188 can solve my problem

matz:
Just merged #4188. Hash size can be obtained by `mrb_hash_size()` now.

PR #4187, opened 2018-12-13
Expose htable header

dsh0416:
As #4186 mentioned, the breaking changes of the removal `khash` structure have made some libraries not working. [Shopify ESS](https://github.com/Shopify/ess) uses the `khash` feature to serialize/deserialize mruby binaries by `msgpack`. The removal of the `khash` makes it impossible to work.

I here suggest that we could expose the `htable` header from the `hash.c` to `htable.h`, which makes it possible to call it from the C program directly. An example usage is shown below.

```c
case MRB_TT_HASH: {
  auto *kh = RHASH_TBL(ruby_value);
  segment *seg;
  mrb_int i;

  if (kh) {
    auto size = size_t{};
    if (size > UINT32_MAX) {
      throw fatal_error(status_code::structure_too_deep);
    }
    packer.pack_map((uint32_t) size);

    seg = kh->rootseg;
    while (seg) {
      for (i=0; i<seg->size; i++) {
        if (!seg->next && i >= kh->last_len) {
          return;
        }
        if (mrb_undef_p(seg->e[i].key)) continue;

        emit_ruby_as_msgpack_rec(engine, seg->e[i].key, packer, depth + 1);
        emit_ruby_as_msgpack_rec(engine, seg->e[i].val, packer, depth + 1);
      }
      seg = seg->next;
    }
  } else {
    packer.pack_map(0);
  }
  return;
}
```

### DIFF
--- a/include/mruby/hash.h
+++ b/include/mruby/hash.h
@@ -9,6 +9,7 @@
 
 #include "common.h"
 #include <mruby/khash.h>
+#include <mruby/htable.h>
 
 /**
  * Hash class
--- a/include/mruby/htable.h
+++ b/include/mruby/htable.h
@@ -1,0 +1,27 @@
+#include "common.h"
+
+struct segkv {
+  mrb_value key;
+  mrb_value val;
+};
+
+typedef struct segment {
+  uint16_t size;
+  struct segment *next;
+  struct segkv e[];
+} segment;
+
+typedef struct segindex {
+  size_t size;
+  size_t capa;
+  struct segkv *table[];
+} segindex;
+
+/* Instance variable table structure */
+typedef struct htable {
+  segment *rootseg;
+  segment *lastseg;
+  mrb_int size;
+  uint16_t last_len;
+  segindex *index;
+} htable;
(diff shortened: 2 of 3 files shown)


take-cheeze:
The implementation of the associative array should be hidden so 👎  for this.
(And the names should be prefixed with `mrb_` when it's going to be exposed.)
Use https://github.com/mruby/mruby/blob/265171a28c2805ae49d0ad70d0f35220b4cb4d4b/include/mruby/hash.h#L214-L215 instead.

dsh0416:
@take-cheeze 
I'm not quite sure about that, because none of the exposed structs are start with `mrb_`. The common used names are things like `RType`, and the previous implementation exposes all methods start with `khash_`. I agree with you that we could do this with the `mrb_hash_foreach` function, but I'm not sure if this could cover all the situations comparing to thre previous exposed `khash` APIs.

On the otherhand, it doesn't make any sense that we expose macros like 
```c
#define RHASH_TBL(h)          (RHASH(h)->ht)
```
inside the definition of `hash.h`, but we can do nothing about that table.

jasl:
And Hash C API don't have get its size which can simple get from `htable->size`

dsh0416:
I believe that #4188 may be a better solution, so that I would close this pr here.

PR #4202, opened 2018-12-25
Simplify full-core.gembox

shuujii:
Dependencies order is auto detected.

### DIFF
--- a/mrbgems/full-core.gembox
+++ b/mrbgems/full-core.gembox
@@ -1,9 +1,6 @@
 MRuby::GemBox.new do |conf|
-  conf.gem :core => "mruby-sprintf"
-  conf.gem :core => "mruby-print"
-
   Dir.glob("#{root}/mrbgems/mruby-*/mrbgem.rake") do |x|
     g = File.basename File.dirname x
-    conf.gem :core => g unless g =~ /^mruby-(print|sprintf|bin-debugger|test)$/
+    conf.gem :core => g unless g =~ /^mruby-(?:bin-debugger|test)$/
   end
 end


dearblue:
`mruby-print` and` mruby-sprintf` do not described their mutual dependencies.
Also, the order of filenames returned by the `Dir.glob` method is undefined.
Therefore, depending on the order, the `Kernel # printf` method will be unavailable (` mruby-sprintf` must be before `mruby-print`).

https://github.com/mruby/mruby/blob/45e8cb47ea0dab9a0cd22d2d78c728799f6a848c/mrbgems/mruby-print/mrblib/print.rb#L51-L54

As a means of solving this, I proposing it in the form of "weak dependency" (#4129).

shuujii:
Sorry for breaking. How about a this solution? (https://github.com/shuujii/mruby/commit/778ea2a77a93accac59981b481a00633a22e63e3)

matz:
I don't like dynamic dispatchin shuujii@778ea2a

shuujii:
This PR can be re-merged because the dependency issue of `mruby-print` and `mruby-sprintf` was solved via 1d8456fe.

dearblue:
@shuujii, How about making a new pull request?
You can reapply it by doing `git revert 37d795dec7b495d418dc2f9020cf980c158ba9ed`.

shuujii:
Thank you for the comment. As I wrote in the comment above, I think that it can be merged as it is now, so I don't think it is necessary to make a new pull request.

matz:
OK. I will reapply.

PR #4225, opened 2019-01-13
Improve compatibility to CRuby for `Float#to_s`

shuujii:
### Bfore:

  ~~~ruby
  p Float::INFINITY.to_s  #=> "inf"
  p 50.0.to_s             #=> "50"
  p 1e20.to_s             #=> "1e+20"
  ~~~

### After / CRuby:

  ~~~ruby
  p Float::INFINITY.to_s  #=> "Infinity"
  p 50.0.to_s             #=> "50.0"
  p 1e20.to_s             #=> "1.0e+20"
   ~~~

### DIFF
--- a/src/numeric.c
+++ b/src/numeric.c
@@ -10,6 +10,7 @@
 #endif
 #include <limits.h>
 #include <stdlib.h>
+#include <string.h>
 
 #include <mruby.h>
 #include <mruby/array.h>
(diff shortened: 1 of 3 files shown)


take-cheeze:
Not a revert request but we should be careful that mruby promotes overflowed integer since mruby doesn't have `BigInt`.
Related issues cause by this PR:
https://github.com/mattn/mruby-json/pull/40
https://github.com/pepabo/mruby-msd/pull/13
https://github.com/mattn/mruby-json/pull/42

matz:
OK, I will revert part of this PR.

matz:
So I removed explicit `.0` addition at the end from `Float#to_s`. @take-cheeze could you check it solves mrbgems problems?

take-cheeze:
LGTM now

matz:
Thank you.

shuujii:
Is it okay in the case of exponential notation? (e.g. `3.0e+45`)

matz:
I think so. The only reason I tried to remove the trailing `.0` is integer compatibility, so exponential notation is not included.

PR #4243, opened 2019-01-24
Remove definition of `Kernel#sprintf` in `mruby-print`

shuujii:
`Kernel#sprintf` is defined in `mruby-sprintf`

### DIFF
--- a/mrbgems/mruby-print/mrblib/print.rb
+++ b/mrbgems/mruby-print/mrblib/print.rb
@@ -52,9 +52,6 @@ module Kernel
     def printf(*args)
       raise NotImplementedError.new('printf not available')
     end
-    def sprintf(*args)
-      raise NotImplementedError.new('sprintf not available')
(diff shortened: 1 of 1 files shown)


takahashim:
> Kernel#sprintf is defined in mruby-sprintf

`mruby-sprintf` is just a mrbgem.  In some situations (such as bare metal environment with less memory, and only uses numeric values except debug logging to serial output), we may want to use `mruby-print` but not want to use `mruby-sprintf`.  So `Kernel#sprintf` might not be  defined, even if you can use `puts` and/or `p` in `mruby-print`.

matz:
Yes, but we have no big reason to define `sprintf` in `mruby-print` gem.

shuujii:
`sprintf` in `mruby-print` merely raise `NotImplementedError`. Therefore there is no essential difference if the implementation isn't provided, I think.

Issue #4239, opened 2019-01-20
About corrected patches for mruby core initialization


dearblue:
If out of memory occurs with `mrb_open()`, the process will be crashed by `SIGABRT` or `SIGSEGV`.

This behavior is undesirable for me as I try to generate multiple `mrb_state`.

[This patch set](https://github.com/mruby/mruby/compare/mruby:b83e4df...dearblue:6797ebf) corrects problems that occurred by simulating out of memory.
(branch is in [fix-mrb_open-with-nomem](https://github.com/dearblue/mruby/tree/fix-mrb_open-with-nomem))

However, it is divided into 18 steps commit, and some can be made independent.

Should it be better to divide the pull requests for each, or may they be a single pull request at once?

Thanks.

## Potential issue

If `mrb_XXX_gem_final()` is called while initialization processing is not completed, problems may arise there.

matz:
No, you don't have to split the PR. But longer (or bigger) PR might take time to review.

dearblue:
Ok. Thank you.
We will `git rebase` in this form and make a pull request.

PR #4250, opened 2019-01-27
Fix problems with `mrb_open()` if occurs out of memory

dearblue:
(This comment is a copied from #4239)

If out of memory occurs with `mrb_open()`, the process will be crashed by `SIGABRT` or `SIGSEGV`.

This behavior is undesirable for me as I try to generate multiple `mrb_state`.

This patch set corrects problems that occurred by simulating out of memory.

Thanks.

## Potential issue

If `mrb_XXX_gem_final()` is called while initialization processing is not completed, problems may arise there.


matz:
merged.

PR #4258, opened 2019-02-02
Extend only when necessary in `lib/mruby-core-ext.rb`

shuujii:


matz:
Maybe we should drop 1.8 support soon. But thank you anyway.

shuujii:
> Maybe we should drop 1.8 support soon.

That's good news for me 😄

PR #4263, opened 2019-02-06
Remove definition of `MRuby::Build#exefile` in `mruby-bin-mruby-config`

shuujii:
It's defined in `lib/mruby/build.rb`.


matz:
appveyor reports an error. maybe we should update `lib/mruby/build.rb` as well?

shuujii:
Sorry, I will fix it and recreate pull request.

PR #4265, opened 2019-02-07
Fix `Enumerator#(initialize|inspect)` for `nil`/`false`

shuujii:


matz:
`Enumerator.new(obj,method,args..)` form has been deprecated. So I will remove this altogether. But thank you anyway.

PR #4271, opened 2019-02-10
`bin=` should be an instance method in `MRuby::Gem::Specification`

shuujii:
`Array#flatten` is unneeded because we should use `bins=` for passing an
array, I think.

### DIFF
--- a/lib/mruby/gem.rb
+++ b/lib/mruby/gem.rb
@@ -119,8 +119,8 @@ module MRuby
         @conflicts << {:gem => name, :requirements => req.empty? ? nil : req}
       end
 
-      def self.bin=(bin)
-        @bins = [bin].flatten
+      def bin=(bin)
+        @bins = [bin]
       end
 
       def build_dir


matz:
If we remove `flatten`, the assignment should be `@bin = bin` I think.

shuujii:
Thank you for your comment. I interpreted your comment as follows:

- `bin=` should accept both one name (`String`) and multipule names (`Array`).
- If we don't use `flatten`, the assignment must be `@bins = bin` etc for
  an array.

But `bin=` accept both (one name and multiple names) and `bins=` (just an
attr writer) accept multiple names only is unnatural for me.

I think it should be one of the following:

1. `bin=` accept one name only and `bins=` accept multiple names only (this PR).
2. `bins=` accept both and we remove `bin=`.

matz:
Hmm, it seems no one uses `bin=` and uses `bin << path` instead. So I am going to remove the method.

PR #4279, opened 2019-02-16
Fix destroyed "inline packed symbols" on 32 bit mode with `MRB_WORD_BOXING`

dearblue:
When building with `MRB_WORD_BOXING` in 32 bit mode, symbols are destroyed depending on the character string length.

```
$ build/host32-word/bin/mruby -e 'p [:zzzzzz, :ZZZZZ]'
[:zzzzb, :ZZZd]
```

In 32-bit mode with `MRB_WORD_BOXING`, only the value of` mrb_sym` is guaranteed 24 bits, there are only 4 spaces for lower case letters and 3 letters for numeral alphabet mixture.

For this reason, I restricted the length of the character string that can be inline packed symbols with this combination of configurations.

The "inline packed symbols" is nice idea :thumbsup:

### DIFF
--- a/src/symbol.c
+++ b/src/symbol.c
@@ -34,13 +34,18 @@ static const char pack_table[] = "_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRS
 static mrb_sym
 sym_inline_pack(const char *name, uint16_t len)
 {
+  enum {
+    lower_length_max  = (MRB_SYMBOL_BITSIZE - 2) / 5,
+    mix_length_max    = (MRB_SYMBOL_BITSIZE - 2) / 6
+  };
+
   char c;
   const char *p;
   int i;
   mrb_sym sym = 0;
   int lower = 1;
 
-  if (len > 6) return 0;        /* too long */
+  if (len > lower_length_max) return 0; /* too long */
   for (i=0; i<len; i++) {
     uint32_t bits;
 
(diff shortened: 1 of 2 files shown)


src/symbol.c
```diff
@@ -34,13 +34,18 @@ static const char pack_table[] = "_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRS
 static mrb_sym
 sym_inline_pack(const char *name, uint16_t len)
 {
+  enum {
```
matz: I'd use `const int` here rather than `enum`. I don't want to use `enum` for a non-enumerative purpose.

matz:
I don't see any reference to `MRB_WORD_BOXING` in the patch. Does this work OK for the case?

src/symbol.c
dearblue: OK. I fixed it.

dearblue:
`MRB_SYMBOL_BITSIZE` that I previously entered in #4077 is defined only when there is` MRB_WORD_BOXING`.

This patch includes commits that have been modified as always defined.

This should result in `MRB_SYMBOL_BITSIZE` =` 24` for 32 bit mode + `MRB_WORD_BOXING`, otherwise` 32`.

  - `build_config.rb`

    ```ruby
    MRuby::Build.new do
      toolchain :clang
      cc.defines << %w(MRB_INT32 MRB_WORD_BOXING)
      cc.flags << "-m32"
      linker.flags << "-m32"
      gem core: "mruby-bin-mruby"
      gem core: "mruby-print"
    end
    ````

  - exec test code

    ```
    $ ./build/host/bin/mruby -e 'p [:zzzzzz, :ZZZZZ]'
    [:zzzzzz, :ZZZZZ]
    ```

PR #4288, opened 2019-02-20
`assert_true`/`assert_false` should pass when actual is only `true`/`false`

shuujii:
For the following reasons:

- Previous behavior is confusable because it's different from test/unit rubygem's `assert_true`
- Tests may pass unintentionally in an inappropriate way; ref #4285 #4287

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -78,7 +78,7 @@ end
 def assert_true(ret, msg = nil, diff = nil)
   if $mrbtest_assert
     $mrbtest_assert_idx += 1
-    unless ret
+    unless ret == true
       msg ||= "Expected #{ret.inspect} to be true"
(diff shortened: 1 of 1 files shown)


matz:
This is a bit controversial. In Ruby, there's no `boolean`. Thus `assert_true` can accept any truthy value in theory. But since it can detect more errors (like #4285), I apply it anyway. Maybe reverted in the future.

shuujii:
`assert` is provided as a method that accept truthy value in test/unit and
minitest rubygem, but mruby's `assert` is different behavior. Therefor we
should add other method (e.g. `assert_truthy`) for this purpose if
necessary, I think.

matz:
Probably. I will revisit the issue when it (`assert_truthy`) is really needed.

PR #4296, opened 2019-02-24
Refine mrbgem name in assertion failure/skip message for core test

shuujii:
### Before

    Skip: GC in rescue => backtrace isn't available (mrbgems: mruby-test)

### After

    Skip: GC in rescue => backtrace isn't available (core)

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -22,7 +22,7 @@ def assertion_string(err, str, iso=nil, e=nil, bt=nil)
   msg += " [#{iso}]" if iso && iso != ''
   msg += " => #{e.cause}" if e && e.respond_to?(:cause)
   msg += " => #{e.message}" if e && !e.respond_to?(:cause)
-  msg += " (mrbgems: #{GEMNAME})" if Object.const_defined?(:GEMNAME)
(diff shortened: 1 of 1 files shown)


test/assert.rb
```diff
@@ -22,7 +22,7 @@ def assertion_string(err, str, iso=nil, e=nil, bt=nil)
   msg += " [#{iso}]" if iso && iso != ''
   msg += " => #{e.cause}" if e && e.respond_to?(:cause)
   msg += " => #{e.message}" if e && !e.respond_to?(:cause)
-  msg += " (mrbgems: #{GEMNAME})" if Object.const_defined?(:GEMNAME)
+  msg += " (#{GEMNAME == 'mruby-test' ? 'core' : "mrbgems: #{GEMNAME}"})"
```
dearblue: `GEMNAME` does not seem to be defined in `bintest`.
Therefore, if an assertion fails, an undefined exception will be raised.

shuujii: Ah, I've forgotten `test/assert.rb` is run on CRuby. I'll fix the issue.

Thank you!

PR #4299, opened 2019-02-26
Remove unneeded `=>` in test skip/error messages

shuujii:
### Before

    Skip: File.expand_path (with ENV) =>  (mrbgems: mruby-io)

### After

    Skip: File.expand_path (with ENV) (mrbgems: mruby-io)

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -20,8 +20,10 @@ end
 def assertion_string(err, str, iso=nil, e=nil, bt=nil)
   msg = "#{err}#{str}"
   msg += " [#{iso}]" if iso && iso != ''
-  msg += " => #{e.cause}" if e && e.respond_to?(:cause)
-  msg += " => #{e.message}" if e && !e.respond_to?(:cause)
+  if e
+    m = e.respond_to?(:cause) ? e.cause : e.message
+    msg += " => #{m}" if m && !m.empty?
+  end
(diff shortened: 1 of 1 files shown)


test/assert.rb
```diff
@@ -20,8 +20,10 @@ def t_print(*args)
 def assertion_string(err, str, iso=nil, e=nil, bt=nil)
   msg = "#{err}#{str}"
   msg += " [#{iso}]" if iso && iso != ''
-  msg += " => #{e.cause}" if e && e.respond_to?(:cause)
-  msg += " => #{e.message}" if e && !e.respond_to?(:cause)
+  if e
+    m = e.respond_to?(:cause) ? e.cause : e.message
+    msg += " => #{m}" if m && !m.empty?
```
matz: I have no objection to removing unneeded `=>` but this patch does not seem to remove it. Am I missing something?

shuujii: My expression might not be good.

It is my intention to output `=>` only when it involves exception message.

### With message

    Skip: GC in rescue => backtrace isn't available (core)

### Without message

    Skip: Method#source_location (mrbgems: mruby-method)

An intention when a mrbgem name is outputted in 280bad6 is the same, I think.

matz: Got it. Thank you for your explanation.

PR #4301, opened 2019-02-27
Remove unnecessary backticks; ref #2858

shuujii:


matz:
This quoting style (`something') is not uncommon in the mruby (and CRuby) source (e.g. mruby-method/src/method.c#L330) . Should we change them uniformly?

shuujii:
I don't have strong opinion about quoting style (I'd prefer \`something'
because it is same as CRuby).

Quoting style was changed \`something' to 'something' by #2858 but there is
only one place that isn't unified in `parse.y`. Therefore I created this PR.

PR #4306, opened 2019-03-01
Use assertion methods in 'Array (Shared Array Corruption)' test

shuujii:


matz:
This particular (incomplete) test is to ensure #359 was fixed. The choice is either:
* remove this test altogether
* make sure clearing shared array does not corrupt the other.
My vote is the former.

shuujii:
My intention is the later. I'd prefer the later because I think we don't
easily notice an issue if tests don't exist, but I don't oppose either.

matz:
I am not positive about adding regression tests to the standard test suites. Mostly because it increases the number of assertion and time to execute the tests. Maybe we should add regression test suites.

PR #4311, opened 2019-03-03
Simplify `MRubyTestSkip` in `test/assert.rb`

shuujii:


test/assert.rb
```diff
@@ -19,11 +19,8 @@ def t_print(*args)
 # Create the assertion in a readable way
 def assertion_string(err, str, iso=nil, e=nil, bt=nil)
   msg = "#{err}#{str}"
-  msg += " [#{iso}]" if iso && iso != ''
-  if e
-    m = e.respond_to?(:cause) ? e.cause : e.message
-    msg += " => #{m}" if m && !m.empty?
-  end
+  msg += " [#{iso}]" if iso && !iso.empty?
+  msg += " => #{e}" if e && !e.to_s.empty?
```
matz: If I remember correctly `e.to_s` does not use `cause` at all. Is this intentional?

shuujii: Yes, this is intentional.

- Previous `cause` is not `Exception#cause` but `MRubyTestSkip#cause`
  (mruby don't have `Exception#cause`).
- I removed `MRubyTestSkip#cause` and merely used `Exception#to_s` (same as
  `Exception#message`).
- I want to pass `String` as `e` for future improvement.

matz: OK, I understand. Thank you for  your explanation.

PR #4317, opened 2019-03-07
Set `GEMNAME` on bintest

shuujii:


shuujii:
Sorry, it seems the previous PR mixed. I'll recreate PR.

PR #4324, opened 2019-03-12
Update `AUTHORS` [ci skip]

shuujii:


shuujii:
Thank you!

BTW there is the word 'Original' at first line in `AUTHORS`, so should separate original members and others?

matz:
The word "original" does not mean anything. Probably I should remove it.

PR #4334, opened 2019-03-20
Remove redundant content in assertion failure message and diff

shuujii:
Based on test-unit RubyGem.

#### Example of before this patch:

    - Assertion[1] Failed: Expected 1 to be nil
       Expected: nil
         Actual: 1
    - Assertion[2] Failed: Expected [1, 3] to include 2
       Collection: [1, 3]
           Object: 2

#### Example of after this patch:

    - Assertion[1]
       <1> was expected to be nil
    - Assertion[2]
       <[1, 3]> was expected to include
       <2>


matz:
I admit the message has redundant content, but the new `<something>` format looks ugly to my eyes.

shuujii:
I updated assertion failure message and diff based on minitest RubyGem. How about this?

matz:
Looks much better. Thank you!

PR #4341, opened 2019-03-25
Remove implementation of `Symbol#===`

shuujii:
For reducing program size.

### DIFF
--- a/src/symbol.c
+++ b/src/symbol.c
@@ -306,26 +306,6 @@ mrb_init_symtbl(mrb_state *mrb)
  *
  */
 
-
-/* 15.2.11.3.1  */
-/*
- *  call-seq:
- *     sym == obj   -> true or false
- *
- *  Equality---If <i>sym</i> and <i>obj</i> are exactly the same
- *  symbol, returns <code>true</code>.
- */
-
-static mrb_value
-sym_equal(mrb_state *mrb, mrb_value sym1)
-{
-  mrb_value sym2;
-
-  mrb_get_args(mrb, "o", &sym2);
-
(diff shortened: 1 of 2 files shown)


take-cheeze:
This shouldn't be merged.
`Symbol#===` should be moved to `mruby-symbol-ext` gem.
Though this may change behavior of `case ～ when ～ end` and will broke some code.

take-cheeze:
See: https://wandbox.org/permlink/haMjLZxsE4drNBxV

shuujii:
Sorry, I don't understand what is the problem. Example above seems to produce the same result before and after this change.

matz:
@take-cheeze probably you have misunderstood the behavior of `undef`. The `undef` operator makes the specified method uncallable (thus an error in wandbox), whereas this patch removes the definition from `Symbol` class (and makes the overridden `Kernel#===` appeared).

take-cheeze:
Ah, sorry seems to fallback to `Kernel#===` .

take-cheeze:
And CRuby uses `rb_obj_equal` too.
https://github.com/ruby/ruby/blob/46968fab0a0cec4a3e6e37f30b13d20998ba5e7a/string.c#L10493
Specializing like `mrb_type(rhs) == MRB_TT_SYMBOL && mrb_symbol(self) == mrb_symbol(rhs)` seems to be better implementation.

take-cheeze:
Though it still makes incompatibility with CRuby: https://wandbox.org/permlink/cCS7wIV2sfqUjoyv

matz:
This is the undefined behavior in ISO, so the difference is OK, I think (and no one removes the important `===` from `Kernel` module in realistic applications).

PR #4352, opened 2019-04-01
Fix modifiable class name

shuujii:
#### Fix the following example:

    ```ruby 
    Object.const_set :A, Module.new{const_set :B, Class.new}
    ab = A::B.to_s
    p ab         #=> "A::B" # Good
    ab[0] = "x"
    p A::B.to_s  #=> "x::B" # Bad
    ```


shuujii:
Sorry, I've updated because fixes incorrect.

PR #4356, opened 2019-04-03
Add `assert_match` and `assert_not_match`

shuujii:


matz:
I am against this change. The difference in the behavior between mruby (`fnmatch` like matching) and other ruby (regular expression matching) is not acceptable.
Besides that, `str =~ str` cause `TypeError` in CRuby anyway.

shuujii:
Is this proposal acceptable if the behavior (`fnmatch` like matching) is
same between mruby and other ruby?

matz:
Yes. I think it's OK that `assert_match` does not use regular expression match like `CRuby`.

matz:
I didn't make myself clear. The `test/driver.rb` is for mruby, so that you don't have to support `CRuby` here.
I should have noticed that when you added `CRuby` support for `t_print`. No branch based on `RUBY_ENGINE` is required.

shuujii:
I will list my thoughts below:

- `test/assert.rb` is running on both CRuby (for **bintest**) and mruby (for **mrbtest**). `mrbgems/mruby-test/driver.c` is only used on mruby.

- Therefore `t_print` is originally supported CRuby (it rescued `__t_printstr__`) and branch is needed. I don't have strong opinion about branching method, though using `RUBY_ENGINE` is lighter than rescue, `respond_to?` etc, I think.

- In `assert_match`, I can probably remove branch if I reimplement `_str_match?` by Ruby code. But I'm reluctant to do it because regular expression (or `File.fnmatch`) is available in CRuby.

shuujii:
I've updated so that the behavior on mrbtest (mruby) and bintest (Ruby) is the same.

Please review it.

PR #4386, opened 2019-04-16
Fix compilation on gcc 4.9.x

tomas:
Older gcc's don't permit these variables to be declared within `for` loops.

### DIFF
--- a/mrbgems/mruby-test/driver.c
+++ b/mrbgems/mruby-test/driver.c
@@ -50,10 +50,10 @@ static mrb_value
 t_print(mrb_state *mrb, mrb_value self)
 {
   mrb_value *argv;
-  mrb_int argc;
+  mrb_int argc, i;
 
   mrb_get_args(mrb, "*!", &argv, &argc);
-  for (mrb_int i = 0; i < argc; ++i) {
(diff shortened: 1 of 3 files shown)


shuujii:
GCC 4.9 seems to permit `for` loop initial declaration in C99 mode (see [this example](https://wandbox.org/permlink/fZsJdVD0CW9El5E1)).

Does occur an error if you specify `-std=gnu99` compiler option in your environment?

tomas:
@shuujii I guess that would work too, but it would require updating `build_config.rb`, right?

shuujii:
Probably yes if you have customized compiler flags in `build_config.rb`.

If you have used default `build_config.rb`, `tasks/toolchains/gcc.rake` is loaded:

https://github.com/mruby/mruby/blob/2f6baf60e957711a85dfdd05725fa946d274b09d/build_config.rb#L1-L9

and `-std=gnu99` should be set:

https://github.com/mruby/mruby/blob/2f6baf60e957711a85dfdd05725fa946d274b09d/tasks/toolchains/gcc.rake#L1-L4

However you need to explicitly include `-std=gnu99` if you have used `CFLAGS` environment variable (maybe we should fix so that it is automatically set even in this case).

PR #4391, opened 2019-04-19
Change modifier to `MRB_INLINE` from `static inline`

dearblue:
Changed some functions defined under the `include/` directory from `static inline` to `MRB_INLINE`.
The purpose is to make it easy to understand as part of the exposed C API.

However, I may have changed the extra functions to `MRB_INLINE`.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -868,7 +868,7 @@ typedef const char *mrb_args_format;
  */
 MRB_API mrb_int mrb_get_args(mrb_state *mrb, mrb_args_format format, ...);
 
-static inline mrb_sym
+MRB_INLINE mrb_sym
 mrb_get_mid(mrb_state *mrb) /* get method symbol */
 {
   return mrb->c->ci->mid;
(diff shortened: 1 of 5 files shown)


shuujii:
I didn't know the proper use of `MRB_INLINE` and `static inline`. I think this should be documented or changed to more meaningful name (`MRB_INLINE_API`, `MRB_API_INLINE` etc).

matz:
`MRB_INLINE` is an abstraction of `static inline` and `static __inline` (MSVC). So mruby should always use `MRB_INLINE` instead of `static inline`.

shuujii:
It seems that `MRB_INLINE` and `static inline` are the same because:

https://github.com/mruby/mruby/blob/e9f3902ab5544de31fc91459e1846650aa228142/include/mruby/value.h#L75-L78

So these are distinguished in this PR, I think.

dearblue:
> So mruby should always use `MRB_INLINE` instead of `static inline`.

I did not know this, and I was convincing that `MRB_INLINE` and `static inline` are distinguished.

Therefore, this PR was submitted from false recognition.

matz:
Ah, I forgot about `inline` definition in `mruby/value.h`. I will reorganize the defines.

PR #3977, opened 2018-03-22
Add API `mrb_read_irep_flags` to read irep to dynamic memory.

take-cheeze:
Related to: #3956.
Since in some cases we can't release memory allocated for symbols properly.

PR #4403, opened 2019-04-24
Read irep from buffers

dearblue:
This patch appends `_buf` to some functions that read irep, and adds a function to check the buffer size as an argument.
Also, confirmation of the binary version is done by the internal `read_binary_header()` function which is called indirectly.

  - `MRB_API mrb_irep *mrb_read_irep_buf(mrb_state *mrb, const void *buf, size_t bufsize)`
  - `MRB_API mrb_value mrb_load_irep_buf(mrb_state *mrb, const void *buf, size_t bufsize)`
  - `MRB_API mrb_value mrb_load_irep_buf_cxt(mrb_state *mrb, const void *buf, size_t bufsize, mrbc_context *c)`

If you give `bufsize = -1`, the buffer size is virtually not checked.
The `read_binary_header ()` function has been changed to return `MRB_DUMP_READ_FAULT` if the size of irep itself is larger than the buffer size.

----

The `mrb_read_irep ()` function, etc. assumes that literal data is given.

For example, given irep data received by inter-process communication etc., it will occur problems if there is a conflict between the buffer size and `struct rite_binary_header.binary_size`.

So I need to check the size and version of the binary data before passing it to the `mrb_read_irep()` function.

I wanted a function to take care of this and made it a pull request.

----

This PR competes with #3977.

### DIFF
--- a/include/mruby/dump.h
+++ b/include/mruby/dump.h
@@ -31,6 +31,7 @@ MRB_API mrb_value mrb_load_irep_file(mrb_state*,FILE*);
 MRB_API mrb_value mrb_load_irep_file_cxt(mrb_state*, FILE*, mrbc_context*);
 #endif
 MRB_API mrb_irep *mrb_read_irep(mrb_state*, const uint8_t*);
+MRB_API mrb_irep *mrb_read_irep_buf(mrb_state*, const void*, size_t);
 
 /* dump/load error code
  *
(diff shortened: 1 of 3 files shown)


matz:
I am still wondering whether function names are sufficient. `mrb_read_irep_buf` or `mrb_read_irep_from_buf`?

dearblue:
I thought `mrb_read_irep_from_buf` was clear but somewhat redundant, so I made it `mrb_read_irep_buf`.
If it is not desirable, I will change it.

Asmod4n:
I had a similar issue a while ago when reading a irep over the network, only had to pass it a string as a mrb_value and call mrb_str_modify on it to work.

But I guess this would be a better solution :)

PR #4427, opened 2019-05-09
Use `mrb_obj_is_kind_of` instead of `Kernel#is_a?` in `mruby-io`

shuujii:


shuujii:
Sorry, I made a mistake, so I will close once.

PR #4440, opened 2019-05-17
Add support for CC="gcc --option ..." again

kou:
If $rake_root_fiber is used, sh runs command in another Fiber. If
command is ran in another Fiber, "rescue RuntimEerror" can't rescue
exception for system(...) failure.

How to reproduce:

    $ CC="gcc -std=gnu99" ./minirake
    (in /home/kou/work/ruby/mruby.kou)
    CC    mrbgems/mruby-compiler/core/codegen.c -> build/test/mrbgems/mruby-compiler/core/codegen.o
    sh: 1: gcc -std=gnu99: not found
    rake aborted!
    Command Failed: ["gcc -std=gnu99" -g -std=gnu99 ...]

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -30,10 +30,13 @@ module MRuby
     def _run(options, params={})
       return sh command + ' ' + ( options % params ) if NotFoundCommands.key? @command
       begin
+        fiber, $rake_root_fiber = $rake_root_fiber, nil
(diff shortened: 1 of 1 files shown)


shuujii:
Does this change prevent commands from running in parallel?

matz:
@shuujii you are right. I will fix the issue in a different way.

kou:
Oh, sorry.

PR #4446, opened 2019-05-17
(Proof of Concept) mruby tuning profiles [ci skip]

dearblue:
Not only mruby, it is one of the difficult things to adjust the performance vs. memory efficiency of the software.
If the approximate target device is divided and the adjustment value for it is prepared by default, it is a good indicator to do fine adjustment.

This PR divides into four profiles.

***Caution: There is no basis for the definitions in the patch.***

- `MRB_CONSTRAINED_BASELINE_PROFILE` - for microprocessors. Reduce memory consumption.
- `MRB_BASELINE_PROFILE` - Default value of mruby.
- `MRB_MAIN_PROFILE` - For desktop computers. Assume that a huge amount of RAM, 10 MiB or more, is on board.
- `MRB_HIGH_PROFILE` - for servers. Assume that mruby VM has a long life.

As you can see the profile name has been ~~stolen~~ imitated from H.264.

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -159,4 +159,70 @@
 # define TRUE 1
 #endif
 
+/*
+** mruby tuning profiles
+**/
+
+/* A profile for micro controllers */
+#if defined(MRB_CONSTRAINED_BASELINE_PROFILE)
+# ifndef KHASH_DEFAULT_SIZE
+#  define KHASH_DEFAULT_SIZE 16
+# endif
+
+# ifndef MRB_STR_BUF_MIN_SIZE
+#  define MRB_STR_BUF_MIN_SIZE 32
+# endif
+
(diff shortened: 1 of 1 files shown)


matz:
Very interesting. I will merge it to experiment.

dearblue:
Oh...
Because I using `~~stolen~~` in the intention of joke, is my commit message OK...?
And I hope will be removed it anytime you no longer need or unable mature by next mruby release...

dearblue:
I forgot to mention how to apply the profile.
#I think it is not necessary to explain to wise people anew, though.

Just add the profile name to `conf.cc.defines` in `build_config.rb`.

----

As a concept, I came up with a `conf.profile` like `conf.toolchain` and `conf.gembox` now.

PR #4449, opened 2019-05-18
Move `tap` to `mruby-kernel-ext` from `mruby-object-ext`

shuujii:


matz:
I think this is the wrong direction. `mruby-kernel-ext` should contain function like methods, and `mruby-object-ext` should contain methods shared by all objects. Thank you for pointing out the inconsistency, anyway.

PR #4451, opened 2019-05-18
Implementation of method reference operator

ukrainskiysergey:
Experimental implementation of experimental Method Reference Operator (based on https://github.com/ruby/ruby/commit/67c574736912003c377218153f9d3b9c0c96a17b).
Not sure if it should be merged into mruby, but I ported it just in case.


matz:
The method reference operator was retracted from Ruby 2.7 (decided in today's dev meeting).
It may be resurrected in the future. I close this issue for the time being.

PR #4455, opened 2019-05-19
Remove "LINE" section reader

dearblue:
Because it is not currently output by `mrbc`.


matz:
OK. Now is the time to remove this.

PR #4456, opened 2019-05-19
Remove byteorder processing from irep load/dump

dearblue:
The `-E` and `-e` switches of the `mrbc` command are remains.
Because for backward compatibility.

To be on the safe side, I confirmed that `.mrb` files generated by `mrbc` are the same on little-endian machine (FreeBSD 12.0 AMD64) and big-endian machine (NetBSD 8.0 SPARC64 on QEmu).

### DIFF
--- a/include/mruby/dump.h
+++ b/include/mruby/dump.h
@@ -17,10 +17,6 @@
 MRB_BEGIN_DECL
 
 #define DUMP_DEBUG_INFO 1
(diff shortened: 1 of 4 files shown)


matz:
We will keep this (for a while). Because we have a vague plan to store endian native integers/floats in the `mrb` files. If we abandon the idea, we will apply this patch.

PR #4468, opened 2019-05-25
Rename `Kernel#__method__` to `Kernel#__callee__`

shuujii:
Because the current behavior of `__method__` is equivalent to `__callee__`.

### Example

  ```ruby
  # example.rb
  def src
    __send__(ARGV[0])
  end
  alias dst src
  %w[src dst].each {|n| puts "call #{n} => #{__send__(n).inspect}"}
  ```

#### Ruby:

    $ ruby example.rb __method__
    call src => :src
    call dst => :src

    $ ruby example.rb __callee__
    call src => :src
    call dst => :dst

#### mruby:

    $ bin/mruby example.rb __method__
    call src => :src
    call dst => :dst

### DIFF
--- a/mrbgems/mruby-enumerator/mrblib/enumerator.rb
+++ b/mrbgems/mruby-enumerator/mrblib/enumerator.rb
@@ -596,7 +596,7 @@ module Kernel
   #       def repeat(n)
   #         raise ArgumentError, "#{n} is negative!" if n < 0
   #         unless block_given?
-  #           return to_enum(__method__, n) do # __method__ is :repeat here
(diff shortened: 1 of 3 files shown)


matz:
I don't like this approach. The right way is either fixing the `__method__` behavior or providing both `__method__` and `__callee_` (with correct behavior).

shuujii:
I just checked a little, but I think it is difficult to correctly implement `__method__` without increasing memory usage of core because original method name needs to be saved. Is there any good way?

Personally, I don't think it is desirable to increase memory usage of core just for `__method__`, and I think `__method__` doesn't use it much if `__callee__` exists.

matz:
It's finally merged.

PR #4483, opened 2019-06-01
Prohibit `MRB_INT64` on 32-bit architecture

shuujii:
I think `mrb_int` size should be less than or equal to pointer size. In the
case of `MRB_INT64` on 32-bit architecture, `struct RString` etc sizes
exceed 6 words at least for now.

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -63,6 +63,10 @@
 # endif
 #endif
 
+#if defined(MRB_32BIT) && defined(MRB_INT64)
+#error MRB_INT64 cannot be used on 32-bit architecture
+#endif
+
 #define MRB_COMPLEX_NUMBERS
 #define MRB_RATIONAL_NUMBERS
 
--- a/include/mruby/boxing_word.h
+++ b/include/mruby/boxing_word.h
@@ -11,10 +11,6 @@
 # error MRB_INT16 is too small for MRB_WORD_BOXING.
 #endif
 
-#if defined(MRB_INT64) && !defined(MRB_64BIT)
-#error MRB_INT64 cannot be used with MRB_WORD_BOXING in 32-bit mode.
-#endif
-
 #ifndef MRB_WITHOUT_FLOAT
 struct RFloat {
   MRB_OBJECT_HEADER;


dearblue:
I will oppose.

  - In 32-bit mode kernel space, 32-bit integers are insufficient (in personal terms, I'm glad if I can handle unsigned integer with more than 64 bits...).

    If you define `MRB_WITHOUT_FLOAT`, the restriction is more strong.

  - `gembox "full-core"` succeeds even if `mrb_int` type included in `struct RArray` or `struct RString` is replaced with `intptr_t` (currently `mruby-complex` and `mruby-rational` requires modification).

    I do not think that there is a problem because it can not be longer than 32 bits length in 32-bit mode (it would be a bug if there were any).

shuujii:
Thank you for your comment.

Then, as you say, `mrb_int` size will be allowed to exceed pointer size and type for sizes such as string length it may be better to limit to pointer size (introduce a new type, e.g. `mrb_size`, which is an alias for `intptr_t` or `uintptr_t`?).

Or, it may be possible to limit `mrb_int` size to pointer size and introduce a type for `Fixnum` that be allowed to exceed pointer size (e.g. `mrb_fixnum`).

dearblue:
There is no objection to the introduction of `mrb_size` and a change of type.

I think that it is better to leave the definition of `mrb_int` as it is from the viewpoint of the compatibility so far.

----

I checked what kind of transition the type of the `len` field of `struct RString` has followed:

  - Initially it was `size_t`:
    https://github.com/mruby/mruby/blame/2fc1b8af2db70e2e008cecb9fd2d8600034f95f9/include/mruby/string.h#L29
  - It became `long`:
    https://github.com/mruby/mruby/commit/2970b6aadbb0d5afef683854e0fe36b05024fc36#diff-d1ee00f2602251e1778ccf96047654c7L29
  - It became `int`:
    https://github.com/mruby/mruby/commit/f4cf8ea423d8f9df1d341adba63d1416779aae8a#diff-d1ee00f2602251e1778ccf96047654c7L29
  - It has now become the current `mrb_int`:
    https://github.com/mruby/mruby/commit/7e910bd4c2623775c00eb0fa052efc9e62d7b115#diff-d1ee00f2602251e1778ccf96047654c7L32
  - Release mruby-1.0.0.

----

Other object structures that exceed 6 words in 32-bit mode are `struct RBreak` defined in `include/mruby/error.h`.

https://github.com/mruby/mruby/blob/366617dbc1cbdc96292295f9d31b519579e6968f/include/mruby/error.h#L35-L39

If we define `MRB_USE_FLOAT`, it will be 6 words length.

  - print object size:
  
    ```
    % cc -m32 -Iinclude -o sizeof sizeof.c && ./sizeof
    size of struct RBreak: 28
    % cc -m32 -Iinclude -DMRB_USE_FLOAT -o sizeof sizeof.c && ./sizeof
    size of struct RBreak: 24
    ```

  - `cat sizeof.c`

    ```c
    #include <mruby.h>
    #include <mruby/error.h>
    #include <stdio.h>

    int
    main(int argc, char *argv[])
    {
      printf("size of struct RBreak: %zd\n", sizeof(struct RBreak));
      return 0;
    }
    ```

shuujii:
Thank you for your investigation.

I think another fix is necessary for `RBreak`, so it might be better to create another ticket.

shuujii:
I close this PR because it has been resolved via #4662 and #4835.

PR #4487, opened 2019-06-03
Move inilialized flag to struct member from bit flag in `struct RRange`

shuujii:


include/mruby/range.h
```diff
@@ -24,6 +24,7 @@ struct RRange {
   mrb_value beg;
   mrb_value end;
   mrb_bool excl;
+  mrb_bool initialized;
```
matz: Adding a member enhances the size of `struct RRange` that gradually increases the size of `RVALUE`.
Why do you want to make `initialized` a member?

shuujii: I don't think the size will change because there is padding in `struct RRange`.

### Example:

```c
/* example.c - this file is a polyglot written in C and Ruby :) */

#if 0
require 'tempfile'
out = Tempfile.new.path
[32,64].product(%w[NO NAN WORD]) do |m, b|
  next if m == 32 && b == "NAN"
  opt = "-O -I./include -m#{m} -DMRB_#{b}_BOXING -o #{out}"
  system "#{ENV['CC'] || 'cc'} #{opt} #{__FILE__} && #{out}"
end
__END__
#endif

#include <stdio.h>
#include <mruby.h>
#include <mruby/range.h>

int
main(int argc, char **argv)
{
  printf(
    "CPU: %lu-bit, Boxing: %4s, RRange size: %lu byte (%lu word)\n",
    sizeof(void*) * CHAR_BIT,
#ifdef MRB_WORD_BOXING
    "word",
#elif MRB_NAN_BOXING
    "nan",
#else
    "no",
#endif
    sizeof(struct RRange),
    sizeof(struct RRange) / sizeof(void*)
  );
  return 0;
}
```

#### Before this patch:

```
$ CC=cc ruby example.c
CPU: 32-bit, Boxing:   no, RRange size: 20 byte (5 word)
CPU: 32-bit, Boxing: word, RRange size: 24 byte (6 word)
CPU: 64-bit, Boxing:   no, RRange size: 40 byte (5 word)
CPU: 64-bit, Boxing:  nan, RRange size: 48 byte (6 word)
CPU: 64-bit, Boxing: word, RRange size: 48 byte (6 word)
```

#### After this patch:

```
$ CC=cc ruby example.c
CPU: 32-bit, Boxing:   no, RRange size: 20 byte (5 word)
CPU: 32-bit, Boxing: word, RRange size: 24 byte (6 word)
CPU: 64-bit, Boxing:   no, RRange size: 40 byte (5 word)
CPU: 64-bit, Boxing:  nan, RRange size: 48 byte (6 word)
CPU: 64-bit, Boxing: word, RRange size: 48 byte (6 word)
```

The reasons for the change are:

- Program size is slightly smaller.
- Simple and natural.
- It may be slightly faster.

shuujii: Sorry, I fixed wrong code comment style.

shuujii: I don't remember why I used the `flags` member originally, but I probably misunderstood size is increased or there was no guarantee that additional members would be initialized to zero.

shuujii:
BTW, I just realized, I misunderstood that `mrb_value` size is 1 word with NaN boxing. If NaN boxing is enabled, `mrb_value` size is always 8 byte (same as `double` size), so 2 word on 32-bit CPU.

Therefore, the condition to embed edges was wrong originally, so `struct RRange` size exceeded 6 word with NaN boxing on 32-bit CPU.

I will create a separate PR about this.

matz:
I close this pull-request.

PR #4520, opened 2019-06-21
Fix segfault when removing ivars on immediates (fixes #4519)

### LINKED ISSUE #4519: Null pointer dereference in mrb_check_frozen
The following input demonstrates a crash:
```ruby
0.remove_instance_variable '@a'
```
It appears the issue was introduced in 030dd6655e47dad4af4b48a74646712c00684698.

ASAN report:
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==33115==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x00010b5630c2 bp 0x7ffee48512f0 sp 0x7ffee48512c0 T0)
==33115==The signal is caused by a READ memory access.
==33115==Hint: address points to the zero page.
    #0 0x10b5630c1 in mrb_check_frozen (mruby:x86_64+0x1001c30c1)
    #1 0x10b567d3e in mrb_iv_remove (mruby:x86_64+0x1001c7d3e)
    #2 0x10b3c1ded in mrb_obj_remove_instance_variable (mruby:x86_64+0x100021ded)
    #3 0x10b442014 in mrb_vm_exec (mruby:x86_64+0x1000a2014)
    #4 0x10b43055a in mrb_vm_run (mruby:x86_64+0x10009055a)
    #5 0x10b48fc4c in mrb_top_run (mruby:x86_64+0x1000efc4c)
    #6 0x10b6daa5e in mrb_load_exec (mruby:x86_64+0x10033aa5e)
    #7 0x10b6db919 in mrb_load_file_cxt (mruby:x86_64+0x10033b919)
    #8 0x10b3a36b7 in main (mruby:x86_64+0x1000036b7)
    #9 0x7fff750cc3d4 in start (libdyld.dylib:x86_64+0x163d4)

==33115==Register values:
rax = 0x0000000000000000  rbx = 0x00007ffee4851360  rcx = 0x0000000000000000  rdx = 0x000061a000000080
rdi = 0x0000100000000000  rsi = 0x0000000000000000  rbp = 0x00007ffee48512f0  rsp = 0x00007ffee48512c0
 r8 = 0x00007ffee4851340   r9 = 0x00001fffdc90a268  r10 = 0x0000100000000000  r11 = 0xffffffffffffffa0
r12 = 0x00007ffee485cee0  r13 = 0x00007ffee485cf00  r14 = 0x00001fffdc90a2a0  r15 = 0x00007ffee485cec0
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (mruby:x86_64+0x1001c30c1) in mrb_check_frozen
==33115==ABORTING
Abort trap: 6
```
This issue was reported by Dinko Galetic & Denis Kasak (https://hackerone.com/dgaletic).

dkasak:
The problem is that `mrb_iv_frozen` treats the `mrb_value` as a pointer,
but it can also be an immediate (`nil`, a fixnum, etc). This leads to
a segfault since the immediate's value is not a valid pointer.

Looking at `mrb_obj_frozen` and comparing with CRuby behaviour,
immediates should always be treated as frozen. This commit checks
whether the value involved is an immediate and raises `FrozenError` if
so.

### DIFF
--- a/src/variable.c
+++ b/src/variable.c
@@ -521,7 +521,9 @@ mrb_obj_iv_inspect(mrb_state *mrb, struct RObject *obj)
 MRB_API mrb_value
 mrb_iv_remove(mrb_state *mrb, mrb_value obj, mrb_sym sym)
 {
-  mrb_check_frozen(mrb, mrb_obj_ptr(obj));
+  if (mrb_immediate_p(obj) || MRB_FROZEN_P(mrb_obj_ptr(obj)))
+    mrb_frozen_error(mrb, obj);
+
   if (obj_iv_p(obj)) {
     iv_tbl *t = mrb_obj_ptr(obj)->iv;
     mrb_value val;
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1133,7 +1133,7 @@ MRB_API mrb_noreturn void mrb_exc_raise(mrb_state *mrb, mrb_value exc);
 MRB_API mrb_noreturn void mrb_raise(mrb_state *mrb, struct RClass *c, const char *msg);
 MRB_API mrb_noreturn void mrb_raisef(mrb_state *mrb, struct RClass *c, const char *fmt, ...);
 MRB_API mrb_noreturn void mrb_name_error(mrb_state *mrb, mrb_sym id, const char *fmt, ...);
-MRB_API mrb_noreturn void mrb_frozen_error(mrb_state *mrb, void *frozen_obj);
+MRB_API mrb_noreturn void mrb_frozen_error(mrb_state *mrb, mrb_value frozen_obj);
 MRB_API void mrb_warn(mrb_state *mrb, const char *fmt, ...);
 MRB_API mrb_noreturn void mrb_bug(mrb_state *mrb, const char *fmt, ...);
 MRB_API void mrb_print_backtrace(mrb_state *mrb);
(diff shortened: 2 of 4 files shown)


src/variable.c
```diff
@@ -521,11 +521,15 @@ mrb_obj_iv_inspect(mrb_state *mrb, struct RObject *obj)
 MRB_API mrb_value
 mrb_iv_remove(mrb_state *mrb, mrb_value obj, mrb_sym sym)
 {
-  mrb_check_frozen(mrb, mrb_obj_ptr(obj));
-  if (obj_iv_p(obj)) {
+  if (mrb_immediate_p(obj)) {
+    mrb_raisef(mrb, E_FROZEN_ERROR, "can't modify frozen %S",
+               mrb_obj_value(mrb_class(mrb, obj)));
+  } else if (obj_iv_p(obj)) {
```
shuujii: I'm the one who made the related change. If you call `mrb_raisef()` directly, the binary size will increase, so I think you should use `mrb_frozen_error()`.

example: https://github.com/shuujii/mruby/commit/f422d1434b

The signature of `mrb_frozen_error()` has been changed because it was originally inappropriate.

dkasak: Your patch looks better, with the changes to the signature of `mrb_frozen_error`. What about the tests, though? I think I'll incorporate your changes into my PR.

shuujii: Your tests looks good to me (however, `.freeze` may not be necessary).

dkasak: Oops, yes, that was left there by accident from something I had earlier.

dkasak:
I updated the PR with the suggested changes.

Issue #4511, opened 2019-06-17
Infinite loop in String#each_line when passed an empty string separator


lopopolo:
On Ruby 2.6.3, String#each_line behaves like this with an empty string separator:

```
[2.6.3] > "aaa\n\nbbbbb\n\n\n\n\ncccc".each_line('').map(&:itself)
=> ["aaa\n\n", "bbbbb\n\n", "cccc"]
[2.6.3] > "aaa\nbbbbbcccc".each_line('').map(&:itself)
=> ["aaa\nbbbbbcccc"]
```

The string is split on successive runs of newlines.

As of mruby@c078758, the two above code snippets result in an infinite loop.

lopopolo:
Relevant code:

https://github.com/mruby/mruby/blob/c078758644443fdac6f91867e00abb152f670599/mrblib/string.rb#L14-L27

The while loop never advances the offset since it is zero length.

lopopolo:
This implementation of String#each_line does not have the bug, passes ruby/spec, and does not use `Regexp` (it does depend on `String::try_convert` though):

```ruby
  def each_line(separator = $/, getline_args = nil) # rubocop:disable Style/SpecialGlobalVars
    return to_enum(:each_line, separator, getline_args) unless block_given?

    if separator.nil?
      yield self
      return self
    end
    raise TypeError if separator.is_a?(Symbol)
    raise TypeError if (separator = String.try_convert(separator)).nil?

    if separator.empty?
      append = "\n\n" if separator.empty?
      matched = ''
      part = ''
      dup.split('') do |c|
        if c == "\n"
          matched << c
        elsif matched.length > 1
          yield self.class.new("#{part}#{matched}")
          matched = ''
          part = c
        else
          part << matched unless matched.empty?

          matched = ''
          part << c
        end
      end

      yield self.class.new("#{part}#{matched}") unless part.empty?
      return self
    end
    append = separator
    matched = ''
    part = ''
    dup.split('') do |c|
      if c == separator[matched.length]
        matched << c
      else
        part << c
      end
      next unless matched == separator

      yield self.class.new("#{part}#{append}")
      matched = ''
      part = ''
    end
    yield self.class.new(part) unless part.empty?

    self
  end
```

dearblue:
@lopopolo, I was interested in your patch and tried it.

There were two problems besides `String.try_convert`, but we can run `mrbtest` to success by fixing it.

  - The `String#<<` method is not in the mruby core. Replace it with `+=`.
[...]

PR #4527, opened 2019-06-22
Add paragraph mode to String#each_line in mrblib

lopopolo:
#4511 demonstrated an infinite loop in `String#each_line` when  …
given an empty string separator. In MRI, an empty separator places
String#each_line in paragraph mode, where the String is separated on
successive runs of newlines. In paragraph mode, the String
`"abc\n\n\ndef\nxyz"` is split into `["abc\n\n\n", "def\nxyz"]`.

This commit makes the String#each_line implementation as close to
ruby/spec compliant as possible given the limitations of mruby core.
With this patch, the following specs fail for `String#each_line`:

- uses `$/` as the separator when none is given (can be fixed by
  aliasing and redefining the method to use $/ as the default value
  of separator in mruby-io)
- when no block is given returned Enumerator size should return nil
  (`Enumerator#size` is not supported on mruby)
- tries to convert the separator to a string using to_str (`String#to_str`
  is not implemented on mruby)

This patch has similar memory consumption compared to the prior
implementation and is takes 4x the time the prior implementation takes to
execute:

```console
/usr/bin/time -l ./bin/mruby -e '("aaa\n\nbbbbb\n\n\n\n\ncccc" * 100000).each_line("\n") { }';
```

cc @dearblue 

Fixes GH-4511.

### DIFF
--- a/mrblib/string.rb
+++ b/mrblib/string.rb
@@ -11,18 +11,63 @@ class String
   # and pass the respective line.
   #
   # ISO 15.2.10.5.15
-  def each_line(rs = "\n", &block)
-    return to_enum(:each_line, rs, &block) unless block
-    return block.call(self) if rs.nil?
-    rs.__to_str
-    offset = 0
-    rs_len = rs.length
-    this = dup
-    while pos = this.index(rs, offset)
-      block.call(this[offset, pos + rs_len - offset])
-      offset = pos + rs_len
+  def each_line(separator = "\n", getline_args = nil)
+    return to_enum(:each_line, separator, getline_args) unless block_given?
+
+    if separator.nil?
+      yield self
+      return self
     end
-    block.call(this[offset, this.size - offset]) if this.size > offset
+    raise TypeError unless separator.is_a?(String)
+
+    start = 0
+    pointer = 0
+    string = dup
+
+    if separator.empty?
+      matched_newlines = 0
+      while pointer < string.length
+        c = string[pointer]
+        if c == "\n"
+          matched_newlines += 1
+        elsif matched_newlines > 1 && self.class == String
+          yield string[start...pointer]
+          matched_newlines = 0
+          start = pointer
+        elsif matched_newlines > 1
+          yield self.class.new(string[start...pointer])
+          matched_newlines = 0
+          start = pointer
+        else
+          matched_newlines = 0
+        end
+        pointer += 1
+      end
+    else
+      matched_length = 0
+      separator_length = separator.length
+      while pointer < string.length
+        c = string[pointer]
+        pointer += 1
+        matched_length += 1 if c == separator[matched_length]
+        next unless matched_length == separator_length
+
+        if self.class == String
+          yield string[start...pointer]
+        else
+          yield self.class.new(string[start...pointer])
+        end
+        matched_length = 0
+        start = pointer
+      end
+    end
+    return self if start == string.length
+
+    if self.class == String
+      yield string[start..-1]
+    else
+      yield self.class.new(string[start..-1])
+    end
     self
   end
 


shuujii:
The second argument seems to be unnecessary.

lopopolo:
@shuujii the second argument is specified by CRuby. I've included it for compatibility.

lopopolo:
My latest commit improves performance of the base case to be within a factor of 2 of the existing implementation.

shuujii:
Because the second argument isn't used, it is better to remove it or implement correctly, I think.

BTW, I think `unless block` is faster than `unless block_given?` (ref. ef3bbd2e).

dearblue:
@lopopolo, there are two things I noticed in order to optimize the speed.

  - Pass position and length instead of passing range to get substrings.  
    Avoid range objects being generated each time in the loop.
  - Assign a method call whose value does not change in a loop to a variable outside the loop.  
    Because `==` is also a method, branch by putting a boolean value into the variable.

lopopolo:
> Because the second argument isn't used, it is better to remove it or implement correctly, I think.

I've removed this in the latest patch.

lopopolo:
The latest implementation is faster than the existing implementation while remaining as memory efficient. To test performance I ran this command:

```sh
/usr/bin/time -l ./bin/mruby -e '1000.times { ("aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000).each_line("\n") { } }';
```

## Existing mrblib implementation

```
mruby@master
       17.91 real        17.84 user         0.06 sys
  54132736  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
     45916  page reclaims
         0  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
         0  voluntary context switches
      1472  involuntary context switches
```

## This PR at `8ea45b8`

```
mruby@string-each-line-paragraph-mode-8ea45b86
       17.91 real        17.84 user         0.05 sys
  54140928  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
     45915  page reclaims
         0  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
         0  voluntary context switches
      1343  involuntary context switches
```

## This PR at `6b92acf3`

```
mruby@string-each-line-paragraph-mode-6b92acf3
       17.40 real        17.34 user         0.05 sys
  54132736  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
     45914  page reclaims
         0  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
         0  voluntary context switches
      1219  involuntary context switches
```

lopopolo:
@shuujii and @dearblue thank you for helping me iterate on this PR. I think I've addressed all of the feedback which resulted in an implementation that is faster than the existing one while still being mostly spec-compliant. 🎉

dearblue:
It's a great :+1:

I realized I could replace it with a common loop.

example:

```ruby
if separator.empty?
  separator = "\n\n"
  paragraph_mode = true
end

while (pointer = string.index(separator, start))
  pointer += sep_len
  if paragraph_mode
    pointer += 1 while string[pointer] == "\n"
  end
  if should_yield_subclass_instances
    ...snip...
```

I'm worried about whether it fits your spec and the speed down, but I think the mruby bytecode size will be smaller.

Becomes like the dispensing of small information, I'm sorry.

lopopolo:
@dearblue thanks for the optimization! That does end up cutting down the code size quite a bit and it does still pass ruby-spec. 🎉 

I think this patch is ready for final review.

Issue #4522, opened 2019-06-22
`String#each_line` is slow with `MRB_UTF8_STRING`


dearblue:
I tried with the latest c53b7cedccf7f5260dc8b4f88c5f93ea550bc5df.

```
% /usr/bin/time ruby26 -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
        0.18 real         0.16 user         0.01 sys
% /usr/bin/time ./mruby@master -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
        0.05 real         0.05 user         0.00 sys
% /usr/bin/time ./mruby@master-utf8 -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
      126.55 real       126.27 user         0.00 sys
```

I noticed when I was previewing #4511.

dearblue:
Comparison performance #4569 vs. master (5420895bcb14bdef8fd35d1abab09f63a003baed) and ruby-2.6.3

```console
% /usr/bin/time ./mruby@master -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
       87.85 real        87.82 user         0.00 sys
% /usr/bin/time ./mruby@improve-index -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
       58.54 real        58.52 user         0.00 sys
% /usr/bin/time ruby26 -e 'str = "aaa\n\nbbbbb\n\n\n\n\ncccc" * 10000; str.each_line("\n") { }'
        0.18 real         0.17 user         0.01 sys
```

In order to be even faster, it is likely that `String#each_line` will need to be reimplemented to use an external iterator.

PR #4529, opened 2019-06-23
[WIP] Optimize utf8 string length and byte offset functions

lopopolo:
Replace existing utf8 strlen and byte offset functions with the UTF-8
length algorithm developed by George Pollard and discussed at

    https://www.daemonology.net/blog/2008-06-05-faster-utf8-strlen.html

The algorithm assumes that the String is valid UTF-8, which I believe is
enforced by `utf8_strlen`. The algorithm uses an ASCII fastpath and skips
characters based on their UTF-8 continuation bytes. This approach can be
more than twice as fast as the existing implementation.

The core algorithm is adapted for the following functions:

- utf8len
- mrb_utf8_len
- chars2bytes
- bytes2chars

This patch allows removing the utf8len_codepage static array.

Benchmarks (default build with MRB_UTF8_STRING defined)

`("aaa\n\nbbbbb\n\n\n\n\ncccc" * 100000000).reverse!`

master: 20.99s
patched: 7.36s

`("aaa\n\nbbbbb\n\n\n\n\ncccc" * 100000000)[1000000 - 1]`

master: 18.89s
patched: 8.14s

`("aaa\n\nbbbbb\n\n\n\n\ncccc" * 100000).each_line("\n") { }`

master: [does not complete in 4 minutes wall clock time]
patched: 0.02s

**NOTE**: I did not write the implementation of the algorithm, I only
adapted it to fit within the mruby APIs.

Fixes GH-4522.

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -215,83 +215,224 @@ mrb_gc_free_str(mrb_state *mrb, struct RString *str)
 }
 
 #ifdef MRB_UTF8_STRING
+static inline mrb_int
+utf8_sequence_len(const unsigned char lead)
+{
+  if (lead < 0x80)
+    return 1;
+  else if ((lead >> 5) == 0x6)
+    return 2;
+  else if ((lead >> 4) == 0xe)
+    return 3;
+  else if ((lead >> 3) == 0x1e)
+    return 4;
+  else
+    return 0;
+}
+
 static mrb_int
 utf8len(const char* p, const char* e)
 {
-  const signed char* str = (const signed char*) p;
-  mrb_int i = 0;
-  mrb_int iBefore = 0;
-  mrb_int count = 0;
-
-  while (p + i < e) {
-ascii:
-    i++;
-  }
-
-  count += i - iBefore;
-
-  while (str[i]) {
-    if (str[i] > 0) {
-      iBefore = i;
-      goto ascii;
-    } else {
-      switch (0xF0 & str[i]) {
-      case 0xE0:
-        i += 3;
-        break;
-      case 0xF0:
-        i += 4;
-        break;
-      default:
-        i += 2;
+  mrb_int length = 0;
+  mrb_int sequence_length = 0;
+  const unsigned char* pointer = (unsigned char*) p;
+  const unsigned char* end = (unsigned char*) e;
+  sequence_length = utf8_sequence_len(*pointer);
+  switch (sequence_length) {
+    case 0:
+      // invalid lead, treating as binary data
+      // fall through
+    case 1:
+      // ASCII or binary data
+      length++;
+      pointer++;
+      break;
+    case 2:
+      if (++pointer == end) {
+        // not enough room
+        length++;
         break;
       }
-    }
-
-    return i;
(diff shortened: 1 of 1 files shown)


lopopolo:
cc @dearblue

lopopolo:
I've generated a flamegraph with dtrace that compares master and this PR for the following benchmark:

```
./bin/mruby -e '100000.times { ("aaa\n\nbbbbb\n\n\n\n\ncccc" * 100000).each_line("\n") { } }'
```

## master

Screen Shot 2019-06-23 at 11 08 42 PM

## PR 4529

Screen Shot 2019-06-23 at 11 09 01 PM

shuujii:
It is interesting. However, it doesn't seem to work well in some cases:

#### Before patched with `MRB_UTF8_STRING` (same as Ruby)

    $ bin/mruby -e '["\xe3\x81", "a\x00z"].each{|s| p s.size}'
    2
    3

#### After patched with `MRB_UTF8_STRING`

    $ bin/mruby -e '["\xe3\x81", "a\x00z"].each{|s| p s.size}'
    1
    1

dearblue:
@lopopolo

As a premise, I am imagined that policy of mruby is to handle binary data and strings without distinction.
Therefore, I think that even if there are mixed characters that are invalid as UTF-8, they need to be handled without problems.
I also want to hear other people's opinions about this.

----

Performing malicious tests this patch.

  - The character count of broken UTF-8 strings is different (This is similar to what @shuujii showed):

    ```console
    % ./build/host/bin/mruby -e 'p ("\xf1" * 5).size'
    2
    % ruby26 -e 'str = "\xf1" * 5; p str.encoding, str.size'
    #<Encoding:UTF-8>
    5
    ```

  - Crash when reversing a broken string as UTF-8:

    ```console
    % ./build/host/bin/mruby -e 'p ("\xf1" * 5).reverse'
    zsh: segmentation fault (core dumped)  ./mruby@test-utf8 -e 'p ("\xf1" * 5).reverse'
    ```

  - If missing trailing characters, it will make an out-of-range reference:

    ```console
    % ./build/host/bin/mruby -e 'p "\xf1"[0]'
    "\xf1\x00\x00\x00"
    ```

dearblue:
`mrbtest` does not stop...

```console
% CFLAGS='-O0 -Wall -DMRB_UTF8_STRING' ./minirake -j10 clean `pwd`/build/test/bin/mrbtest
...snip...
% ./build/test/bin/mrbtest

.F..............................................................................F..F...X?..................F..F....F.......F.F.FFFF....F...^C
```

lopopolo:
Oh no, it appears I was running the tests without `MRB_UTF8_STRING` defined. Thanks for the tip @dearblue. This patch is not ready for review. I'll mark it as WIP.

lopopolo:
I've pushed my latest experiment. Not being able to store the encoding on the `RString` means we must process each character in the `char *` to determine the length. The existing approach with the lookup table is faster than what I've been able to come up with.

It looks like CRuby is able to fast-path UTF-8 strlen by storing whether the string has a "broken" encoding or not.

I'm going to abandon this PR. Thanks for taking a look.

PR #4531, opened 2019-06-24
Lazily cache UTF-8 character length on RString

lopopolo:
Store the UTF-8 length as computed by `utf8_strlen` on `RString` when `MRB_UTF8_STRING` is defined. This significantly (3200x) speeds up `String` operations at the cost of one `mrb_int` of memory per `RString`.

The UTF-8 `char_len` is computed on first call to `utf8_strlen` and is cached until `mrb_str_modify` is called, when it is zeroed out. `utf8_strlen` gets a new fastpath for when the character length is cached and retains the fast path for ASCII strings which is used as a fallback.

## Benchmark

```sh
/usr/bin/time -l ./bin/mruby -e 'def str(len); ("💪" * len); end; s = str(100000); puts s.length; 1000000.times { s.length }'
```

## This PR @ `ca692e6`

Completes in 0.30 seconds.

```console
$ /usr/bin/time -l ./bin/mruby -e 'def str(len); ("💪" * len); end; s = str(100000); puts s.length; 1000000.times { s.length }'
100000
        0.30 real         0.30 user         0.00 sys
   2306048  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
       572  page reclaims
         0  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
         0  voluntary context switches
        48  involuntary context switches
```

### master

Completes in 963 seconds. The caching implementation is 3210x faster than master.

```console
$ /usr/bin/time -l ./bin/mruby -e 'def str(len); ("💪" * len); end; s = str(100000); puts s.length; 1000000.times { s.length }'
100000
      963.41 real       961.82 user         0.74 sys
   2252800  maximum resident set size
         0  average shared memory size
         0  average unshared data size
         0  average unshared stack size
       559  page reclaims
         0  page faults
         0  swaps
         0  block input operations
         0  block output operations
         0  messages sent
         0  messages received
         0  signals received
         0  voluntary context switches
    139399  involuntary context switches
```

### DIFF
--- a/include/mruby/string.h
+++ b/include/mruby/string.h
@@ -23,6 +23,9 @@ struct RString {
   union {
     struct {
       mrb_int len;
+#ifdef MRB_UTF8_STRING
+      mrb_int char_len;
+#endif
       union {
         mrb_int capa;
         struct mrb_shared_string *shared;
--- a/src/string.c
+++ b/src/string.c
@@ -258,14 +258,20 @@ mrb_utf8_len(const char *str, mrb_int byte_len)
 static mrb_int
 utf8_strlen(mrb_value str)
 {
+  mrb_int char_len = RSTRING(str)->as.heap.char_len;
   mrb_int byte_len = RSTRING_LEN(str);
 
-  if (RSTRING(str)->flags & MRB_STR_NO_UTF) {
+  if (char_len > 0) {
+    return char_len;
+  }
+  else if (RSTRING(str)->flags & MRB_STR_NO_UTF) {
+    RSTRING(str)->as.heap.char_len = byte_len;
     return byte_len;
   }
   else {
     mrb_int utf8_len = mrb_utf8_len(RSTRING_PTR(str), byte_len);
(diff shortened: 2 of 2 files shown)


lopopolo:
Related to GH-4522 and GH-4529

cc @dearblue

shuujii:
I think this patch will not be accepted as it is because the size of `struct RString` will increase (see: #3297).

However, it may be possible because the size does not increase in a limited environment such as 64-bit system + MRB_INT32 + boxing.

shuujii:
> 64-bit system + MRB_INT32 + boxing

Sorry, boxing has nothing to do with it.

lopopolo:
The tests do not currently pass when invoked like this:

```
CFLAGS="-Wall -DMRB_UTF8_STRING" rake test
```

`str_replace` does not work as expected. `mruby-io` seems to make assumptions about the layout of the `RString` struct. I think I bit off more than I can chew. Abandoning this PR.

PR #4320, opened 2019-03-09
Warn if assertion is missing inside `assert` (like #4309 etc)

shuujii:
Currently many warnings appear (草生える in progress).

~~~
>>> Test host <<<
mrbtest - Embeddable Ruby Test

................W.............................................................
..............................................W...............................
.............................................WW.WWWWW..WWWWWWWWWWWWWWWWWWWWWWW
WWWWWW.W.W..............?....W..........W...................WW................
............WWWWWWWWWWWWWWWWWWWWWWWWWWW.......................................
...........................................WWWWWWWWWWWWWWW.....WWWWWWW........
.....................................WWWWWWW..................................
..........................?...................................................
..............................................................................
..............................................................................
..............................................................................
..............................................................................
........................................................WW....................
..............................................................................
.......W................................................................W.....
.........
Warn: Fiber with splat in the block argument list => no assertion (mrbgems: mruby-fiber)
Warn: Array#bsearch_index => no assertion (mrbgems: mruby-array-ext)
Warn: Time.new [15.2.3.3.3] => no assertion (mrbgems: mruby-time)
Warn: Time [15.2.19] => no assertion (mrbgems: mruby-time)
Warn: Time.gm [15.2.19.6.2] => no assertion (mrbgems: mruby-time)
Warn: Time.local [15.2.19.6.3] => no assertion (mrbgems: mruby-time)
Warn: Time.mktime [15.2.19.6.4] => no assertion (mrbgems: mruby-time)
Warn: Time.now [15.2.19.6.5] => no assertion (mrbgems: mruby-time)
Warn: Time.utc [15.2.19.6.6] => no assertion (mrbgems: mruby-time)
Warn: Time#<=> [15.2.19.7.3] => no assertion (mrbgems: mruby-time)
Warn: Time#asctime [15.2.19.7.4] => no assertion (mrbgems: mruby-time)
Warn: Time#ctime [15.2.19.7.5] => no assertion (mrbgems: mruby-time)
Warn: Time#day [15.2.19.7.6] => no assertion (mrbgems: mruby-time)
Warn: Time#dst? [15.2.19.7.7] => no assertion (mrbgems: mruby-time)
Warn: Time#getgm [15.2.19.7.8] => no assertion (mrbgems: mruby-time)
Warn: Time#getlocal [15.2.19.7.9] => no assertion (mrbgems: mruby-time)
Warn: Time#getutc [15.2.19.7.10] => no assertion (mrbgems: mruby-time)
[...]

PR #4540, opened 2019-06-29
Nested `assert` for mrbtest

dearblue:
When nesting `assert` used in test, it is indented and displayed.  
Assertion numbers are concatenated by `"-"` at this time.

The purpose is to match the apparent numbers when failing with `assert_mruby` which is defined by `mrbgems/mruby-bin-mruby/bintest/mruby.rb` for example.

Child assertions "skip" and "info" are reported as parent assertions "info" and `$ok_test += 1`.
The child assertions "ko" and "crash" are reported as the parent assertion "ko" and `$ko_test += 1`.

When child assertions are mixed, "ko" takes precedence.

If you use `"."` As a concatenation character, I think it is easy to confuse it with the ISO Ruby section number when it is displayed.  
So we chose `"-"` as the concatenation character.

Note that this patch conflicts with #4320.

### Incompatibility

- `$mrbtest_assert_idx` takes `nil` or an integer array object.  
  So far it was `nil` or an integer.
- `$asserts` points to the top of the internal stack in the `assert`.
- `$mrbtest_assert` points to the top of the internal stack in `assert`.

### DIFF
--- a/mrbgems/mruby-bin-mruby/bintest/mruby.rb
+++ b/mrbgems/mruby-bin-mruby/bintest/mruby.rb
@@ -3,9 +3,11 @@ require 'open3'
 
 def assert_mruby(exp_out, exp_err, exp_success, args)
   out, err, stat = Open3.capture3(cmd("mruby"), *args)
-  assert_operator(exp_out, :===, out, "standard output")
-  assert_operator(exp_err, :===, err, "standard error")
-  assert_equal(exp_success, stat.success?, "exit success?")
+  assert do
+    assert_operator(exp_out, :===, out, "standard output")
+    assert_operator(exp_err, :===, err, "standard error")
+    assert_equal(exp_success, stat.success?, "exit success?")
+  end
 end
 
 assert('regression for #1564') do
--- a/mrbgems/mruby-complex/test/complex.rb
+++ b/mrbgems/mruby-complex/test/complex.rb
@@ -1,6 +1,8 @@
 def assert_complex(real, exp)
-  assert_float real.real,      exp.real
-  assert_float real.imaginary, exp.imaginary
+  assert do
+    assert_float real.real,      exp.real
+    assert_float real.imaginary, exp.imaginary
+  end
 end
 
 assert 'Complex' do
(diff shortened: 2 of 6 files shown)


dearblue:
Demonstration:

```ruby
# example test code
assert "test KO #A" do
  assert_false "bad assertion #1"

  assert "nested test KO #B" do
    assert_false "bad assertion #2"

    assert "nested test KO #C" do
      assert_equal "bad assertion #3", "dummy"
    end
  end

  assert do
    assert do
      assert_equal "bad assertion #4", "dummy"
    end

    assert do
      assert "Error" do
        (0/0).to_i
      end
    end

    assert do
      assert "Jump" do
        skip "skip!"
      end
    end
  end

  assert_false "bad assertion #5"
end
```

```console
% ./build/test/bin/mrbtest | ruby -e 's=$stdin.read; nil while s.gsub!(/^(?:.{80})+\K(?=.)/, "\n"); puts s'

................................................................................
........?.......................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
......................................................................FFFXF?.FF.
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
.............
Skip: File.expand_path (with ENV) (mrbgems: mruby-io)
Fail: test KO #A (core)
 - Assertion[1]
    Expected "bad assertion #1" to be false.
 - Assertion[2]
    Fail: nested test KO #B (core)
     - Assertion[2-1]
        Expected "bad assertion #2" to be false.
     - Assertion[2-2]
        Fail: nested test KO #C (core)
         - Assertion[2-2-1]
            Expected: "bad assertion #3"
              Actual: "dummy"
 - Assertion[3]
    Fail: Assertion failed (core)
     - Assertion[3-1]
        Fail: Assertion failed (core)
         - Assertion[3-1-1]
            Expected: "bad assertion #4"
              Actual: "dummy"
     - Assertion[3-2]
        Fail: Assertion failed (core)
         - Assertion[3-2-1]
            FloatDomainError: Error => NaN (core)
     - Assertion[3-3]
        Info: Assertion failed (core)
         - Assertion[3-3-1]
            Skip: Jump => skip! (core)
 - Assertion[4]
    Expected "bad assertion #5" to be false.
Total: 1133
   OK: 1124
   KO: 6
Crash: 1
 Skip: 2
 Time: 0.97 seconds
```

PR #4542, opened 2019-06-29
`Module#extend_object(obj)` should return `obj`

shuujii:


src/class.c
```diff
@@ -1184,7 +1184,7 @@ mrb_mod_extend_object(mrb_state *mrb, mrb_value mod)
   mrb_check_type(mrb, mod, MRB_TT_MODULE);
   mrb_get_args(mrb, "o", &obj);
   mrb_include_module(mrb, mrb_class_ptr(mrb_singleton_class(mrb, obj)), mrb_class_ptr(mod));
-  return mod;
+  return obj;
```
matz: I agree with this change for compatibility with CRuby.

test/t/module.rb
```diff
@@ -275,6 +275,16 @@ def self.const_missing(sym)
   assert_equal 42, Test4ConstMissing.const_get(:ConstDoesntExist)
 end
 
+assert('Module#extend_object', '15.2.2.4.25') do
+  cls = Class.new
+  mod = Module.new { def foo; end }
+  a = cls.new
+  b = cls.new
+  assert_same a, mod.extend_object(a)
```
matz: But the return value from `append_features`, `extend_object` is implementation-defined. So I don't think we need to test the return value here.

shuujii: The specifications of `Module#extend_object` in ISO (final draft) is as follows:

> Let S be the singleton class of object. Invoke the method append features (see
5 15.2.2.4.10) on the receiver with S as the only argument, and return the resulting value.

I think this means returning the object of the argument. Is my interpretation wrong or has it been changed in the final version?

matz: It reads "`extend_object` returns the return value from `append_feature`" and the return value from `append_feature` is implementation defined.

shuujii: Thank you for your elaboration.

Then current mruby conform to ISO, but CRuby does not conform to ISO because it does not return the return value from `append_features`.

Should mruby conform to ISO or CRuby?

matz: ISO here, I think. If the return value is meaningful, it should be compatible with CRuby. But this case is not.

shuujii: Then, there is no need to change it, so I will close this PR.

PR #4550, opened 2019-07-04
Rename `MRB_STR_NO_UTF` to `MRB_STR_NO_MULTI_BYTE`

shuujii:
The previous name may be a little confusing because ASCII only string is
also valid UTF-8 string.


matz:
I'd rather rename it to `MRB_STR_ASCII`, which is clearer and shorter.

PR #4560, opened 2019-07-09
Specialize Enumerable#max and Enumerable#min for Range

lopopolo:
This patch prevents a hang for pathalogical (large) Ranges when
computing max and min.

Range inherits its implementation of max and min from Enumerable.
Enumerable implements max and min by calling each. For Range objects,
this is unnecessary since we know the max and the min by the end and
begin attributes.

It is also very slow. This code hangs unnecessarily:

    (0..2**32).max
    # ... hang
    (0..2**32).min
    # ... hang

This patch overrides max and min after including enumerable to yield
based on the begin and end methods.


shuujii:
This patch is incorrect.

#### Before this patch:

```shell
bin/mruby -e 'p (1...3).max'  #=> 2
```

#### After this patch:

```shell
bin/mruby -e 'p (1...3).max'  #=> 3
```

I strongly recommend writing tests based on CRuby's tests, Ruby Spec, etc.

lopopolo:
Oops I had this fixed in an earlier version of the patch but lost it. I'll add some tests.

shuujii:
I think it is better to refer to CRuby etc for the implementation too.

lopopolo:
Thanks for your patience @shuujii. I verified the behavior of the tests in CRuby by running them in an IRB shell on Ruby 2.6.0.

shuujii:
This patch is still incorrect.

#### Before this patch:

```shell
bin/mruby -e 'p ("a"..."f").max'  #=> "e"
```

#### After this patch:

```shell
bin/mruby -e 'p ("a"..."f").max'  #=> undefined method '-' (NoMethodError)
```

This example is in [Ruby Spec](https://github.com/ruby/spec/blob/0c0a2faab55e37209618a29c3778fa8c77f499e7/core/range/max_spec.rb#L9).

lopopolo:
😞 Thank you @shuujii. I added more tests from Ruby Spec, cleaned up the tests, and got the implementation passing.

lopopolo:
This did require a change to `Range#each` to handle `Float` in the same special case as `Fixnum`.

shuujii:
It doesn't seem to be enough yet:

- When block is given, special path cannot be used because block must be called per comparison (this test is also in Ruby Spec).
- If special path cannot be used, it should be delegated to `super`.
- `Float` is not defined with `MRB_WITHOUT_FLOAT`.

IMO, this extension might be better placed in `mruby-range-ext` gem to avoid increasing core size.

lopopolo:
Ok I think I've got it this time.

1. Moved the fast path implementation to mruby-range-ext
2. Delegate to super if block given
3. delegate to super if endpoints are not Fixnum or Float
4. Import all applicable specs from Ruby Spec
5. Correctly implement the fast path

Thank you so much @shuujii for helping me though this. I'm hoping this will be the first PR I land in mruby 😄

matz:
A beautiful process of improvement. Impressed.

shuujii:
@lopopolo Congrats on your first PR merge in mruby!

More two comments:

- It may not be good to treat `Float` (non `Integer`) the same as CRuby because mruby treats integers and float numbers more uniformly than CRuby (e.g. mruby has `Float#succ`). However I'm not sure how to fix it concretely right now.
- The tests added to `mruby-string-ext` are not tests for `Range#(max|min)` but tests for `Enumerable#(max|min)` because `mruby-string-ext` doesn't depend on `mruby-range-ext`. I think it is undesirable `mruby-string-ext` depends on `mruby-range-ext` or tests for `Range#(max|mix)` exist in `mruby-string-ext`. Therefor, I think it is better to define and use String-like (has `#succ`) class in `mruby-range-ext`.

lopopolo:
@shuujii I've followed up on the bad dependency on mruby-range-ext in GH-4564. 😄

PR #4561, opened 2019-07-09
Fix the order of "expected" and "actual" in `test/t/range.rb`

shuujii:


lopopolo:
Thank you for fixing this broken window. I copy and pasted from this example in GH-4560.

PR #4564, opened 2019-07-09
Implement String#succ in mruby-range-ext for specialized Range#max and Range#min

lopopolo:
Follow up to GH-4560.

cc @shuujii


shuujii:
First of all, `Range#each` treats `String` edges as special and tries `String#upto` first, so I apologize for the confusion.

However, my intention is not to implement `String#succ` (or `String#upto`) itself in `mruby-range-ext`. I think the purpose of testing `String` edges in `Range#(max|mix)` is to verify that they work when fast path is not used, so my intention is to define a class that can do that in a test script.

#### Example:

```ruby
# mrbgems/mruby-range-ext/test/range.rb

class A < Array
  def succ
    dup << size+1
  end
end

assert('Range#max (not fast path without block)') do
  assert_equal [1,2,3], (A[1]..A[1,2,3]).max
  assert_equal [1,2], (A[1]...A[1,2,3]).max
end
```

lopopolo:
@shuujii I'm having trouble finding time to pick this back up. Should I close the PR?

matz:
You don't have to close. I will take care.

PR #4570, opened 2019-07-11
Support `.d` file for some compilers

shuujii:
Support `.d` file for some compilers

`.d` file styles differ between GCC and Clang include in Android NDK r20, 
so dependent files could not be parsed correctly.

#### GCC 8.2.0

```text
/build/host/src/array.o: \
 /src/array.c \
 /include/mruby/common.h \
 /include/mruby/value.h \
 /include/mruby/common.h \
 /src/value_array.h
```

#### Clang 8.0.7 (Android)

```text
/build/host/src/array.o: \
  /src/array.c \
  /include/mruby/common.h \
  /include/mruby/value.h \
  /src/value_array.h

/include/mruby/common.h:

/include/mruby/value.h:

/src/value_array.h:
```

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -127,13 +127,49 @@ module MRuby
     end
 
     private
+
+    #
+    # === Examples of +.d+ file style:
+    #
+    # ==== GCC 8.2.0
+    #
+    #   /build/host/src/array.o: \
+    #    /src/array.c \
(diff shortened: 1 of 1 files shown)


shuujii:
Sorry, this is not due to compiler differences. It is because `-MP` flag is specified in `tasks/toolchains/android.rake`.

shuujii:
I will re-create PR.

PR #4597, opened 2019-07-26
Drop dependency from `mruby-array-ext` to `mruby-enum-ext`

shuujii:


matz:
Is this pull-request combine two changes in one?

shuujii:
Dependency on `mruby-enumerator` dropped at #4595. However, dependency on `mruby-enum-ext` which `mruby-enumerator` depends on remains, so this PR drop it.

matz:
You removed test dependency from `mruby-array-ext` and it's OK.
You replaced `row.count` to `row.size` in `mrblib/array.rb` which does not seem to related to test dependency.

shuujii:
It is not test code but main code that depends on `mruby-enum-ext` (`Array#count` is defined in `mruby-enum-ext`).

It just works accidentally when `enable_test` is specified because there is no distinction between test dependencies and normal dependencies in current implementation.

#### Example

```shell
$ MRUBY_CONFIG=<(echo 'MRuby::Build.new{toolchain :gcc; gem core:"mruby-bin-mruby"; gem core:"mruby-array-ext"}') ./minirake
$ bin/mruby -e 'p [[1,2]].transpose'  #=> NoMethodError: undefined method 'count'
```

matz:
I understand.

PR #4601, opened 2019-07-28
Drop dependency from `mruby-enumerator` to `mruby-enum-ext`

shuujii:


dearblue:
The previous mrbgems with only `mruby-enumerator` without `mruby-enum-ext` as a dependency may not work.

I agree to increase independence.
However, users should be widely informed on the next release.

Actually, `mruby-cfunc` uses `#each_slice` method, but `CFunc::Struct#define` does not work because `mruby-enum-ext` is not specified.

```console
% ./build/host+mruby-cfunc/bin/mruby -e '
	class Foo < CFunc::Struct
	  define CFunc::Int, :bar,
	         CFunc::Double, :quux
	end'
        [0] -e:2
        [1] -e:4
        [2] build/mrbgems/mruby-cfunc/mrblib/cfunc_rb.rb:426:in define
build/mrbgems/mruby-cfunc/mrblib/cfunc_rb.rb:426: undefined method 'each_slice' (NoMethodError)
```

PR #4604, opened 2019-07-29
Fix Time#to_s encoding on Windows

kou:
strftime() on Windows returns locale encoding time zone for "%z" even
if MSDN says "%z" is "The offset from UTC in ISO 8601 format; no
characters if time zone is unknown" in MSDN:

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?view=vs-2019

So we need to convert encoding of string from strftime().

### DIFF
--- a/mrbgems/mruby-time/src/time.c
+++ b/mrbgems/mruby-time/src/time.c
@@ -19,6 +19,8 @@
 #include <string.h>
 #endif
 
+#include <stdlib.h>
+
 #define NDIV(x,y) (-(-((x)+1)/(y))-1)
 #define TO_S_FMT "%Y-%m-%d %H:%M:%S "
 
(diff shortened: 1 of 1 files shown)


kou:
"東京 (標準時)" in CP932 is returned for "%z" on Japanese Windows.

shuujii:
@kou Thank you for your report.

I've noticed time zone offset seems to be "Coordinated Universal Time" etc., on MSVC 2013 in AppVeyor. Is it not digits even with newer MSVC?

kou:
I don't know it.
I tested this with MinGW build.

How did you confirm it in AppVeyor? If you can confirm it in AppVeyor, you can confirm it with newer MSVC in AppVeyor.

kou:
Ah, you may be able to confirm it with https://ci.appveyor.com/api/buildjobs/3ytqapsla21aei4e/artifacts/groonga-9.0.5-e8277e6-x64-vs2017.zip .

```text
PS> bin\groonga.exe
> plugin_register ruby/eval
> ruby_eval 'Time.now.to_s'
```

shuujii:
Time zone offset seems to be digits on VS 2015 and VS 2017 in AppVeyor when I checked before. However, I wondered if this issue happens on newer VS because your report points to the VS 2019 documentation.

kou:
I misunderstood what you said. Sorry.
Visual Studio 2013 returns "東京 (標準時)". Visual Studio 2015 or later return "+09:00".

shuujii:
Thank you for your information. I tried to use own implementation to calculate UTC offset on Visual Studio 2015 or earlier or MinGW via #4607. 

BTW, UTC offset becomes like `+0900` on Visual Studio 2015, 2017 and 2019 in AppVeyor. Are there environments where it becomes like `+09:00` (includes `:`)?

kou:
Sorry. ":" isn't included.

PR #4606, opened 2019-07-30
Refine message to `skip` in nested `assert`

shuujii:
- I think "Info" is used only to `skip` (@dearblue right?), so change to "Skip".
- Changed the default value of `assert` and specify the argument explicitly at the caller of `assert` because it is unnatural "Assertion failed" is output even though the assertion doesn't fail.

### Example:

```ruby
def assert_foo(exp, act)
  assert do
    assert_equal exp[0], act[0]
    assert_equal exp[1], act[1]
  end
end

def assert_bar(exp, act)
  assert do
    skip
  end
end

def assert_baz(exp, act)
  assert do
    assert_equal exp, act
    assert_bar exp, act
  end
end

assert 'test#skip_in_nested_assert' do
  assert_baz 1, 1
end
```

#### Before this patch:

```text
?..
Info: test#skip_in_nested_assert (core)
 - Assertion[1]
    Info: Assertion failed (core)
     - Assertion[1-2]
        Skip: Assertion failed (core)
  Total: 3
     OK: 2
     KO: 0
  Crash: 0
Warning: 0
   Skip: 1
```

#### After this patch:

```text
???
Skip: test#skip_in_nested_assert (core)
 - Assertion[1]
    Skip: assert (core)
     - Assertion[1-2]
        Skip: assert (core)
  Total: 3
     OK: 0
     KO: 0
  Crash: 0
Warning: 0
   Skip: 3
```

### DIFF
--- a/mrbgems/mruby-bin-mruby/bintest/mruby.rb
+++ b/mrbgems/mruby-bin-mruby/bintest/mruby.rb
@@ -3,7 +3,7 @@ require 'open3'
 
 def assert_mruby(exp_out, exp_err, exp_success, args)
   out, err, stat = Open3.capture3(cmd("mruby"), *args)
-  assert do
+  assert "assert_mruby" do
     assert_operator(exp_out, :===, out, "standard output")
     assert_operator(exp_err, :===, err, "standard error")
     assert_equal(exp_success, stat.success?, "exit success?")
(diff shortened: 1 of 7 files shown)


dearblue:
@shuujii,
> I think "Info" is used only to `skip` (@dearblue right?),

Yes, "Info:" is used only for "Skip:" (and "Info:").
The reason is that even if the child assert skips, the parent assert continues without being skipped on the fly.
I needed to mention this in #4540.
If the parent assert should also be skipped on the fly, changes are welcome.

> Changed the default value of `assert` and specify the argument explicitly at the caller of `assert` because it is unnatural "Assertion failed" is output even though the assertion doesn't fail.

It is not desirable to be "Skip: Assertion failed", and I welcome the correction.

PR #4612, opened 2019-08-03
Add constants for floating point number

dearblue:
- Added macros available from C to `include/mruby/numeric.h`.
- Added the constants available from Ruby space to `mruby-numeric-ext`.
   Since `MRB_FLT_ROUNDS` can change its behavior, it is defined as a class method.

### DIFF
--- a/mrbgems/mruby-numeric-ext/src/numeric_ext.c
+++ b/mrbgems/mruby-numeric-ext/src/numeric_ext.c
@@ -1,5 +1,6 @@
 #include <limits.h>
 #include <mruby.h>
+#include <mruby/numeric.h>
 
 static inline mrb_int
 to_int(mrb_state *mrb, mrb_value x)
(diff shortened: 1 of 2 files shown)


matz:
Is there any clear intention to provide constants that even CRuby does not provide?
`Float::EVAL_METHOD`  for example?

dearblue:
There is no clear intention.
I just added it with momentum.

Should I delete the constants from `Float` class?
Or should I also delete related C macros?

matz:
We should probably remove the constants. Macros do not consume any resource at run-time, so we don't have to remove them. But there's no reason to keep unused macros in the source code.

dearblue:
I did rebase and force push.

- added method:
  `Float.rounds`
- added constants:
  `Float::RADIX`
  `Float::MANT_DIG`
  `Float::EPSILON`
  `Float::DIG`
  `Float::MIN_EXP`
  `Float::MIN`
  `Float::MIN_10_EXP`
  `Float::MAX_EXP`
  `Float::MAX`
  `Float::MAX_10_EXP`

matz:
Thank you.

mrbgems/mruby-numeric-ext/src/numeric_ext.c
```diff
@@ -64,6 +73,21 @@ mrb_mruby_numeric_ext_gem_init(mrb_state* mrb)
   mrb_define_method(mrb, i, "allbits?", mrb_int_allbits, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, i, "anybits?", mrb_int_anybits, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, i, "nobits?", mrb_int_nobits, MRB_ARGS_REQ(1));
+
+#ifndef MRB_WITHOUT_FLOAT
+  mrb_define_class_method(mrb, mrb->float_class, "rounds", mrb_float_rounds, MRB_ARGS_NONE());
```
matz: CRuby defines `Float::ROUNDS` but not `Float.rounds`.

dearblue:
`FLT_ROUNDS` can be changed at runtime by the `fesetround()` function.
Since I do not want to define as a constant by myself, I deleted it from the patch.

matz:
That means we should fix CRuby first.

dearblue:
Thank you for review and merge this patch.

matz:
I proposed to remove `Float::ROUNDS` from CRuby.

PR #4613, opened 2019-08-03
Change second argument to `%l` of `mrb_vformat()` to `size_t` from `mrb_int`

shuujii:
- `size_t` is more commonly used.
- `len` argument of `mrb_str_new()` is `size_t`.

NOTE:

The test for `%l` is temporarily disabled because adding a new type to
`mrbgems/mruby-test/vformat.c` causes an error (memory error?) on Visual
Studio 2017 in AppVeyor.

PR #4622, opened 2019-08-09
Retrieve test for `%l` of `mrb_vformat()`; ref #4613

shuujii:
The following compilation error occurs on Visual Studio 2017 in AppVeyor if
commented out part in previous `mrbgems/mruby-test/vformat.c` is enabled.

```text
CC    mrbgems/mruby-test/vformat.c -> build/cxx_abi/mrbgems/mruby-test/vformat.obj
  C:\projects\mruby\mrbgems\mruby-test\vformat.c(176): fatal error C1060: compiler is out of heap space
  INTERNAL COMPILER ERROR in 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe'
      Please choose the Technical Support command on the Visual C++
      Help menu, or open the Technical Support help file for more information
```

This error seems to occur only C++, and not occur Visual Studio 2013/2015.

The cause seems to be deep nested conditional operator (`? :`). So I decided
to generate `mrbgems/mruby-test/vformat.c` with ERB to avoid it.


matz:
Introducing more dependency (`erb`) is not preferable. Is there any other workaround?

shuujii:
I didn't use `erb` command but ERB library which is a Ruby standard library, so I thought it was acceptable.

If it is not acceptable, what about the following actions?

- Commit the generated file like `lex.def` as well.
- Incorporates a small template engine implementation such as [Mote](https://github.com/soveran/mote), [Ate](https://github.com/piscolomo/ate) and [Erubi](https://github.com/jeremyevans/erubi). If we only support `%` and `<%= ... %>` (newlines are not allowed), the following one-liner can also be used :)

  ```ruby
  # one-liner-erb.rb
  eval$<.map{|l|l=~/^%/?$':%|$><<"#{l.gsub(/<%=(.*?)%>/,'#{\1}')}"|}*$/
  ```

  ```c
  %# example.erb
  % native_types = {
  %   c: {type: "char", code: "RSTRING_PTR(obj)[0]"},
  %   d: {type: "int" , code: "(int)mrb_fixnum(obj)"},
  % }
  typedef struct {
  % native_types.each do |spec, attrs|
    <%=attrs[:type]%> <%=spec%>;
  % end
  } VFNative;
  ```

  ```console
  $ ruby one-liner-erb.rb example.erb
  typedef struct {
    char c;
    int d;
  } VFNative;

  $ diff -s =(ruby one-liner-erb.rb example.erb) =(erb example.erb)
  Files /tmp/zshq5HoG1 and /tmp/zshuoc8X6 are identical
  ```

matz:
I addressed this issue in a different way.

PR #4633, opened 2019-08-16
SHARED/FSHARED string is not required when sharing NOFREE string

shuujii:
I think the string buffer of NOFREE string always exists and does not need
to be released, so it can be shared as another NOFREE string.

Also changed the `mrb_shared_string` field order so that eliminate padding if
`int` and `mrb_int` sizes are less than pointer size.


matz:
Nice idea. Thank you!

PR #4636, opened 2019-08-18
Generate doxygen docs for mruby

davidsiaw:
### Context

This commit adds commands to the makefile to generate a documentation website for mruby. With this, inline documentation can show up on web pages in an easy to read manner.

```
make docs
```

For now documentation isn't as complete and useful as it should be but hopefully over time this will encourage writing more useful information into both the source code and extra documentation pages.

### Results

To save you the trouble of installing Doxygen dependencies, this is the resulting generated site:

https://mruby.davidsiaw.net

You can also go to documents like https://mruby.davidsiaw.net/md_doc_guides_mrbconf.html, and find clickable links like these that lead you to the actual documentation of the objects.

image

### Doxyfile

I tried to make the Doxyfile as readable as possible, so that it makes more sense to the human eye.

### Documentation

All documentation to be added to Doxygen need to start with the double asterisk, which is followed in some parts already.
```c
/**
 * Custom data type description.
 */
typedef struct mrb_data_type {
```

You can link to method documentation by going `method()`

```c
/**
 * Create a symbol using mrb_intern_cstr()
 */
```

Files that have important documentation need to be given a file header that is not too different from what we already have (with the double asterisk and `@file` marker):

```
/**
** @file mruby/data.h - Data class
**
```

### Why

Searching for mruby things on google now always makes it think that I typed in mruby by mistake. Having some searchable and indexable documentation will help improve the situation.


davidsiaw:
A strange crash occurred on https://travis-ci.org/mruby/mruby/jobs/573331846. Not sure if this is a known flicker.

dearblue:
Perhaps you have encountered cryptid! :ghost: :jack_o_lantern:

When I checked it a little, it seemed that symbolic links with the same name collided when the tests were executed in parallel.

There was no problem with your patch because there were no failures in other tests.

shuujii:
- Originally YARD document is generated by `rake doc`, so it is difficult to understand that another document is generated by `make docs`.

- It may be better to generate Doxygen document under `doc/doxgen/`, etc. because YARD document is generated under `doc/api/`.

- I think Rake task should be used instead of Make task.

- YARD warnings are increasing (originally different warning appeared).

  ```terminal
  $ rake doc
  [warn]: Unknown tag @class in file `` near line
  [warn]: Unknown tag @brief in file `` near line
  [warn]: Unknown tag @class in file `` near line
  [warn]: Unknown tag @brief in file `` near line
  [warn]: Unknown tag @class in file `` near line
  [warn]: Unknown tag @brief in file `` near line
  ...
  ```

davidsiaw:
> * It may be better to generate Doxygen document under `doc/doxgen/`

Makes sense. I will make another PR to correct that.

> * I think Rake task should be used instead of Make task.

👍 

> * Originally YARD document is generated by `rake doc`

I'll just make it so rake doc generates both the Doxygen and yard docs then

PR #4637, opened 2019-08-18
Prohibit changes to iseq in principle

dearblue:
I thought about inline mruby bytecode like 5b1f25a, but iseq is basically no problems even if it is immutable.

However, there is a compatibility problem because the structure and functions are changed as an API.

Third party programs that incorporate mruby may stop working.


matz:
I think it's OK.

PR #4654, opened 2019-08-21
Take commit hash of mruby too

take-cheeze:


shuujii:
This PR seems to add the following information into the lock file:

```yaml
mruby_version:
  version: 2.0.1
  release_no: 20001
  git_commit: 246179518d8e3641dcb026fb24e8b06f9878faa8
```

- Is this information just output and not used?
- It seems redundant that `version` is repeated like `mruby_version.version`, so I think just `mruby` is better.

PR #4656, opened 2019-08-22
Refine processing for gem lock file

shuujii:
- Defer YAML library and lock file loading until needed.
- Don't write empty parts into lock file.
- Extract code to read/write lock file to `MRuby::Lockfile`.
- `MRuby::Lockfile.disable` disables the use of lock file.

---

The use of lock file is enabled by default (without external gem) and only the `mruby_version` part is written. However, since `mruby_version` part is unused and lock file doesn't make sense, I think it is better to disable the use of lock file by default.

### DIFF
--- a/lib/mruby/lockfile.rb
+++ b/lib/mruby/lockfile.rb
@@ -1,0 +1,73 @@
+autoload :YAML, 'yaml'
+
+module MRuby
+  autoload :Source, 'mruby/source'
+
+  class Lockfile
+    class << self
+      def enable
+        @enabled = true
+      end
+
+      def disable
+        @enabled = false
+      end
+
+      def enabled?
+        @enabled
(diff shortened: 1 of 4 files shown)


lib/mruby/lockfile.rb
```diff
@@ -0,0 +1,73 @@
...
+  autoload :Source, 'mruby/source'
+
+  class Lockfile
+    class << self
+      def enable
+        @enabled = true
+      end
+
+      def disable
+        @enabled = false
```
take-cheeze: You should check this that it's already disabled

shuujii: What is the concrete problem with multiple `MRuby::Lockfile.disable`? I think idempotent is not bad.

PR #4669, opened 2019-08-25
Fixes for Doxygen docs

davidsiaw:
As discussed in https://github.com/mruby/mruby/pull/4636
- I moved the doxygen docs directory to /doc/capi alongside /doc/api which are the yard docs
- I made it so doxygen does not spam so many warnings by basically fixing up most of the warnings (which were mainly in mruby.h)
- I made it so docs are generated in a rake task instead of using make

This PR also fixes a few smaller issues such as the pipe symbol not displaying in the generated docs and code examples not being preformatted.

This PR also adds a note in the README about generating docs.

As before you can see the resulting doxygen docs in https://mruby.davidsiaw.net


shuujii:
I have some comments:

- I think it is better to place all documentation tasks under the `doc` namespace (add `doc_` prefix because minirake can't use namespace). Because it is consistent with GitLab tasks and user friendly for `rake -T`, etc. and shell completion.

- The description of `clean_doc`/`clean_api_doc`/`clean_capi_doc` is all the same.

- There is `sudo apt-get install doxygen` in the failure message of `capi_doc` task, but I think this is somewhat inaccurate because it only works on specific platforms.

- `view` tasks also only work on specific platforms, so I think they are unnecessary (or it is better to work on Linux/Mac/Windows at least).

PR #4673, opened 2019-08-28
Fix `Kernel#initialize_copy` to frozen object

shuujii:


matz:
I have no objection to this change, but since `initialize_copy` is a private method in CRuby, we should not test like `a.initialize_copy(a)`. Probably we should call `initialize_copy` indirectly via `clone` or `dup`.

matz:
The comment above does not mean we need to ensure test suits to be executable by CRuby. But we may support public/private visibility in mruby in the future. So I think it's better to avoid direct invoking of methods marked as private in CRuby.

shuujii:
I agree with your comment. And I withdraw this PR because `Kernel#initialize_copy` doesn't change anything so this PR itself doesn't make much sense.

Thank you.

PR #4677, opened 2019-08-30
`Array#(permutation|combination)` without block should return `self`

shuujii:


shuujii:
Sorry, I made a mistake in the commit message ("without block" -> "with block").

PR #4679, opened 2019-08-31
Implement `mrb_immediate_p` macro for `MRB_WORD_BOXING`

shuujii:
The default implementation of `mrb_immediate_p` uses `mrb_type`. However,
in `MRB_WORD_BOXING`, `mrb_type` has many branches and is inefficient, so
provide an implementation that does not use `mrb_type`.

### DIFF
--- a/include/mruby/boxing_word.h
+++ b/include/mruby/boxing_word.h
@@ -28,11 +28,6 @@ struct RCptr {
 };
 
 #define MRB_FIXNUM_SHIFT 1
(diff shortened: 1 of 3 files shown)


shuujii:
BTW, the default implementation of `mrb_immediate_p` returns true for `MRB_TT_FREE`. Is this intentional?

matz:
`MRB_TT_FREE` objects should not be considered as immediate. Although `MRB_TT_FREE` should not be leaked in the userspace, thus should not cause any problem.

PR #4680, opened 2019-08-31
Implement numbered parameters

ukrainskiysergey:
The implementation of numbered parameters from Ruby 2.7.
Still not fully compatible with ruby's implementation.


matz:
Thank you. May I take care of this patch to adopt the latest behavior in CRuby?

ukrainskiysergey:
No problem.

matz:
The PR is merged (with modification to support numbered parameters like `_1`).

PR #4698, opened 2019-09-09
Deduplicate constant names

dearblue:
If the module/class do `include`, `mod.constants` returned duplicate constant names.

Before patched:

```ruby
p Object.constants.grep(:BasicObject)
# => [:BasicObject, :BasicObject]
# BasicObject::BasicObject and Kernel::BasicObject
```

**NOTE**:

  - The patch includes the `mrb_ary_index()` function as an API.
  - Since the situation continues from before mruby-1.0.0, I think that it is also one hand to make the specification.
    At least I'm not in trouble.


matz:
I made my own fix because I don't want to disclose `mrb_ary_index` that calls `mrb_equal` which could reenter into VM via `mrb_funcall`.

dearblue:
OK, thank you.

PR #4699, opened 2019-09-09
add special local variable _ in mirb

takkaw:
I think underscore variable in irb is useful.
This pull request is adding it to mirb.


matz:
Thank you. But I think this useful `_` variable should be turned on by default. Instead of
```C
#ifdef ENABLE_MIRB_UNDERSCORE
```
You should
```C
#ifndef DISABLE_MIRB_UNDERSCORE
```

takkaw:
Thank you for your comment.
I changed it to be enabled by default.

PR #4716, opened 2019-09-16
Rename MRB_USE_ETEXT_EDATA to MRB_USE_LINK_TIME_RO_DATA_P and support lld linked programs

MaskRay:
In lld linked programs, .rodata comes before .text, thus mrb_ro_data_p
will return false for strings in .rodata. Change the lower bound from
_etext to __ehdr_start to catch these cases. This works for ld.bfd, gold
and lld, and it does not have false positives even if .init_array does
not exist.

Remove the branch that uses _edata: strings in .data can be modified so
this is semantically incorrect. Delete the `__APPLE__` branch (its
manpages say get_etext() and get_edata() are strongly discouraged).
.init_array has been adopted by most ELF platforms to supersede .ctors.

Neither _etext nor _edata is used, so rename MRB_USE_ETEXT_EDATA to
MRB_USE_EHDR_START.

-------------------------------

This removes code added by #3639 @Asmod4n You can use `MRB_USE_CUSTOM_RO_DATA_P` instead. The `_edata` approach can be wrong if the user modifies `.data`, so I don't think it makes lots of sense to support it by default.

Most ELF platforms support .init_array. For example, glibc has supported .init_array since 1999.

To use lld (LLVM linker), you can add `-fuse-ld=lld` to `conf.linker.flags`.
This option is supported by clang newer than Nov 9, 2013, and GCC 9 and newer.
Some Debian derivatives back port a patch to GCC 7 and 8.

Here is the readelf -S dump of my mruby executable:

```
...
  [11] .rodata           PROGBITS        0000000000004da0 004da0 03fdf3 00  AM  0   0 32
  [12] .eh_frame_hdr     PROGBITS        0000000000044b94 044b94 0039d4 00   A  0   0  4
  [13] .eh_frame         PROGBITS        0000000000048568 048568 00ee4c 00   A  0   0  8
  [14] .text             PROGBITS        0000000000058000 058000 08cf00 00  AX  0   0 16
  [15] .init             PROGBITS        00000000000e4f00 0e4f00 000017 00  AX  0   0  4
  [16] .fini             PROGBITS        00000000000e4f18 0e4f18 000009 00  AX  0   0  4
  [17] .plt              PROGBITS        00000000000e4f30 0e4f30 000690 00  AX  0   0 16
  [18] .fini_array       FINI_ARRAY      00000000000e6000 0e6000 000008 08  WA  0   0  8
  [19] .init_array       INIT_ARRAY      00000000000e6008 0e6008 000008 08  WA  0   0  8
```

### DIFF
--- a/doc/guides/mrbconf.md
+++ b/doc/guides/mrbconf.md
@@ -134,21 +134,16 @@ largest value of required alignment.
 
 ## Reduce heap memory configuration.
 
-`MRB_USE_ETEXT_EDATA`
+`MRB_USE_LINK_TIME_RO_DATA_P`
+* Only available on ELF platforms.
 * If you specify the address of a read-only section when creating a symbol or string, that string will be used as it is.
 * Heap memory can be saved.
-* Uses `_etext` and `__init_array_start`.
-* It must be `_etext < data_addr < &__init_array_start`.
-
-`MRB_NO_INIT_ARRAY_START`
-* Ignored if `MRB_USE_ETEXT_EDATA` is not defined.
-* Please try if `__init_array_start` is not available.
-* Uses `_etext` and `_edata`.
-* It must be `_etext < data_addr < _edata`.
+* Uses `__ehdr_start` and `__init_array_start`.
+* It must be `__ehdr_start < data_addr < __init_array_start`.
 
 `MRB_USE_CUSTOM_RO_DATA_P`
-* Takes precedence over `MRB_USE_ETEXT_EDATA`.
-* Please try if both `MRB_USE_ETEXT_EDATA` and `MRB_NO_INIT_ARRAY_START` are not available.
+* Takes precedence over `MRB_USE_LINK_TIME_RO_DATA_P`.
+* Please try if `MRB_USE_LINK_TIME_RO_DATA_P` is not available.
 * The `mrb_ro_data_p()` function is implemented by the user in an arbitrary file.
 * The prototype declaration is `mrb_bool mrb_ro_data_p(const char *ptr)`.
 * Return `TRUE` if `ptr` is in read-only section, otherwise return `FALSE`.
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -88,14 +88,10 @@
 /* number of object per heap page */
 //#define MRB_HEAP_PAGE_SIZE 1024
 
-/* if _etext and _edata available, mruby can reduce memory used by symbols */
-//#define MRB_USE_ETEXT_EDATA
+/* if __ehdr_start is available, mruby can reduce memory used by symbols */
+//#define MRB_USE_LINK_TIME_RO_DATA_P
 
-/* do not use __init_array_start to determine readonly data section;
-   effective only when MRB_USE_ETEXT_EDATA is defined */
-//#define MRB_NO_INIT_ARRAY_START
-
-/* if do not works both MRB_USE_ETEXT_EDATA and MRB_NO_INIT_ARRAY_START,
+/* if MRB_USE_LINK_TIME_RO_DATA_P does not work,
    you can try mrb_ro_data_p() that you have implemented yourself in any file;
    prototype is `mrb_bool mrb_ro_data_p(const char *ptr)` */
 //#define MRB_USE_CUSTOM_RO_DATA_P
--- a/include/mruby/value.h
+++ b/include/mruby/value.h
@@ -309,37 +309,25 @@ mrb_undef_value(void)
   return v;
 }
 
+#if defined(MRB_USE_ETEXT_EDATA) && !defined(MRB_USE_LINK_TIME_RO_DATA_P)
+# ifdef __GNUC__
+#  warning MRB_USE_ETEXT_EDATA is deprecated. Define MRB_USE_LINK_TIME_RO_DATA_P instead.
+# endif
+# define MRB_USE_LINK_TIME_RO_DATA_P
+#endif
+
 #if defined(MRB_USE_CUSTOM_RO_DATA_P)
 /* If you define `MRB_USE_CUSTOM_RO_DATA_P`, you must implement `mrb_ro_data_p()`. */
 mrb_bool mrb_ro_data_p(const char *p);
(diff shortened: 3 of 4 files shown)


MaskRay:
This is how I build mruby. I hope I do it correctly:

```ruby
# build_config.rb
MRuby::Build.new('test') do |conf|
...
  conf.cc.defines = %w(MRB_USE_LINK_TIME_RO_DATA_P)
```

```zsh
# By default ./minirake also builds host-debug and host, which can be time consuming.
bear ./minirake ~/Dev/mruby/build/test/bin/mirb
```

Asmod4n:
This makes mrb_read_irep leak memory on Darwin platforms when the supplied buffer isn’t statically allocated.

matz:
@MaskRay If you address the memory leak issue reported by @Asmod4n, I'd happy to merge the PR.
If you add migration from `MRB_USE_ETEXT_EDATA`, I will be more than happy.

MaskRay:
@Asmod4n Can you please check what memory leak this change causes? If this change causes a memory leak, I am certain w/o this change `-UMRB_USE_ETEXT_EDATA` will also cause a memory leak. Then, we should figure out the memory leak, though, I don't think it is caused by this change.

Using the end of `.data` or `__data` is dangerous. `.data` is writable (SHF_WRITE on ELF platforms), so its contents (e.g., interned strings) can be modified.

Uses who understand the implication can write
```c
static mrb_bool
mrb_ro_data_p(const char *p)
{
  return (const char*)get_etext() < p && p < (const char*)get_edata();
}
```
and link it with mruby.

> If you add migration from MRB_USE_ETEXT_EDATA, I will be more than happy.

How can I add the migration plan? I just started at looking at this repo 2 days ago...

matz:
Some people may have already added `MRB_USE_ETEXT_EDATA` to their build configuration, so when `MRB_USE_ETEXT_EDATA` is defined, `MRB_USE_LINK_TIME_RO_DATA_P` should be defined too, i.e.
```C
#ifdef MRB_USE_ETEXT_EDATA
# define MRB_USE_LINK_TIME_RO_DATA_P
#endif 
```

MaskRay:
I decide to go with:

```c
#if defined(MRB_USE_ETEXT_EDATA) && !defined(MRB_USE_LINK_TIME_RO_DATA_P)
# warning MRB_USE_ETEXT_EDATA is deprecated. Define MRB_USE_LINK_TIME_RO_DATA_P instead.
# define MRB_USE_LINK_TIME_RO_DATA_P
#endif
```

They will see a plethora of warnings because `mruby/value.h` is transitively included in many files, but it should help users to migrate.

```
CC    build/test/mrblib/mrblib.c -> build/test/mrblib/mrblib.o
In file included from /home/ray/Dev/mruby/include/mruby.h:92,
                 from /home/ray/Dev/mruby/build/test/mrblib/mrblib.c:1:
/home/ray/Dev/mruby/include/mruby/value.h:313:3: warning: #warning MRB_USE_ETEXT_EDATA is deprecated. Define MRB_USE_LINK_TIME_RO_DATA_P instead. [-Wcpp]
 # warning MRB_USE_ETEXT_EDATA is deprecated. Define MRB_USE_LINK_TIME_RO_DATA_P instead.
```

Asmod4n:
My bad, it doesn’t leak memory but it stops working when you pass it a dynamically allocated buffer. I believe that’s the reason I made the PR in the past.

MaskRay:
mrb_ro_data_p should return false for a dynamically allocated buffer. The approach makes an assumption: all non-writable sections are placed before writable sections. This is true for the default cases of ELF linkers:

* ld.bfd -z noseparate-code: `R RX R RW`
* ld.bfd -z separate-code (default since binutils 2.31 on x86_64): `R RX R RW`
* lld 8: `R RX RW`
* lld 9: `R RX RW(PT_GNU_RELRO) RW(non PT_GNU_RELRO)`  # Yes, there are 2 PF_R|PF_W PT_LOAD segments. I'm pretty sure about the implications because I made this change.

dearblue:
@MaskRay

The `#warning` preprocessor directive is not standard.
I think many compilers support it, but it's a proprietary extension (you can check it with `gcc -Wpedantic`, `clang -Weverything`, etc.).
A compiler that does not understand `#warning` may generate an error.

This is preferable because `#pragma message("Warning message")` is not an error even if the compiler does not understand it.

However, it is annoying that some compilers understand `#warning` but don't understand `#pragma message`...
As a simple example, I confirmed it with Portable C Compiler (pcc).

If you are only interested in GCC (including clang), you can use `#pragma GCC warning "Warning message"`.
That said, it's relatively new.
<https://github.com/gcc-mirror/gcc/commit/a1f1781ae8b54c384b7d193c2fd3ac8efc86932a>

I would also like to hear the opinions of people who use compilers other than recent GCC and clang.

MaskRay:
@dearblue 

```c
#warning "warning"
```

clang -fsyntax-only -pedantic a.c
```
<source>:1:2: warning: #warning is a language extension [-Wpedantic]
```

gcc -fsyntax-only -pedangic a.c
```
<source>:1:2: warning: #warning is a GCC extension

    1 | #warning warning

      |  ^~~~~~~
```

Compilers reusing clang/gcc frontend will issue such warnings. This doesn't matter, though, as the purpose is to remind people. MSVC errors on the construct, which does not matter, because this macro should not be used on Windows.

```
<source>(1): fatal error C1021: invalid preprocessor command 'warning'
```

dearblue:
Regardless of the Windows, if you use a compiler that does not support the `#warning`, it does not have any confusion by unrelated error message?

Currently, on Windows, the linker will complain about missing `_etext` and `__init_array_start`.
Since this is shown is a direct reason for the build to fail, I think the more preferable.

dearblue:
Or you may want to surround `#warning` with `#ifdef __GNUC __... # endif`.

MaskRay:
Added `#ifdef __GNUC __`

> Since this is shown is a direct reason for the build to fail, I think the more preferable.

Though I think for this build (Windows?), `MRB_USE_ETEXT_EDATA` should not be defined.

dearblue:
Thank you.

PR #4737, opened 2019-09-29
Add an annotation about the return value

dearblue:


matz:
This is not required. The return value from `mrb_hash_fetch` is either the value from the hash or the default value provided. You only need to protect the value created in the function.

dearblue:
This annotation is on `mrb_hash_delete_key ()`.

matz:
Oh sorry.

PR #4753, opened 2019-10-06
Replace in safe `mrb_get_argv()`

dearblue:
The `mrb_get_argv()` function now returns a copy of the argument on the stack or a splat argument.

With this change, `bin/mruby` built with `CFLAGS=-Os` will increase by 200 bytes.

```console
% ls -l mruby@*
-rwxr-xr-x  1 dearblue  dearblue  615448 10月  6 14:06 mruby@master
-rwxr-xr-x  1 dearblue  dearblue  615648 10月  6 14:05 mruby@patched
```

----

Since the `mrb_get_argv ()` function is `MRB_API`, it is expected that the arguments can be obtained correctly.
Or I think better to should delete `MRB_API` as an internal function.

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -520,32 +520,47 @@ to_hash(mrb_state *mrb, mrb_value val)
 
 #define to_sym(mrb, ss) mrb_obj_to_sym(mrb, ss)
 
+static mrb_value*
+get_argv(mrb_state *mrb, mrb_int *argc)
+{
+  mrb_value *array_argv;
+
+  *argc = mrb->c->ci->argc;
+  if (*argc < 0) {
+    struct RArray *a = mrb_ary_ptr(mrb->c->stack[1]);
+
+    *argc = ARY_LEN(a);
+    array_argv = ARY_PTR(a);
+  }
+  else {
+    array_argv = NULL;
+  }
+
+  return array_argv;
+}
+
 MRB_API mrb_int
 mrb_get_argc(mrb_state *mrb)
 {
-  mrb_int argc = mrb->c->ci->argc;
-
-  if (argc < 0) {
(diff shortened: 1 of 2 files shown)


src/class.c
```diff
@@ -586,9 +601,9 @@ mrb_get_args(mrb_state *mrb, const char *format, ...)
   char c;
   mrb_int i = 0;
   va_list ap;
-  mrb_int argc = mrb_get_argc(mrb);
+  mrb_int argc;
   mrb_int arg_i = 0;
-  mrb_value *array_argv = mrb_get_argv(mrb);
+  mrb_value *array_argv = get_argv(mrb, &argc);
```
dearblue: Previously used by `mrb_get_args()`, but replaced by `get_argv()`.

matz:
I am not sure about your intention.
* Why do you combine `mrb_get_argc` and `mrb_get_argv`, where `mrb->argc` is not supposed to be changed?
* Why do you always copy `argv`? I know it's safer, but copying is slower. Maybe we should update the documentation instead.

dearblue:
Currently if `mrb_get_argc()` returns more than 0, `mrb_get_argv()` may return `NULL`.

This PR is to be able to get the correct argument array with `mrb_get_argv()` as one of the solutions.
I propose removing `MRB_API` from `mrb_get_argv()` as another solution.

> * Why do you combine `mrb_get_argc` and `mrb_get_argv`, where `mrb->argc` is not supposed to be changed?

This is because I thought it was not good to have three functions that perform the same preprocessing.

> * Why do you always copy `argv`?

The argument is copied when the stack is passed. Does not copy for splatter arguments.
Also, if you don't copy, you will eventually have to follow `mrb->c->ci->argc` to distinguish between stack and splatter.

dearblue:
I will reconsider this PR.

PR #4754, opened 2019-10-06
Get keyword arguments with `mrb_get_args()`

dearblue:
Keyword arguments are now expanded by the `mrb_get_args()` function.
As before, mruby core and core gems should assume that keyword arguments cannot be handled directly from C.

For the interface, I referred to CRuby's `rb_get_kwargs()`.
For implementation, I referred to `OP_KARG` or etc.

----

When `MRB_ENABLE_KEYWORD_SPECIFIER` was defined, `bin/mruby` increased by about 800 bytes when building with `gcc -Os` on FreeBSD 12.
Even in the raw state, it has increased by about 130 bytes due to the addition of error messages.

```console
% ls -l mruby@*
-rwxr-xr-x  1 dearblue  dearblue  615448 10月  6 18:06 mruby@master
-rwxr-xr-x  1 dearblue  dearblue  615576 10月  6 18:06 mruby@patched
-rwxr-xr-x  1 dearblue  dearblue  616344 10月  6 18:05 mruby@patched+kwargs
```

----

You can check the test code at <https://github.com/dearblue/mruby-test-kwargs/blob/master/test/kwargs.rb>.
You can find out how to use it in C in the documentation or <https://github.com/dearblue/mruby-test-kwargs/blob/master/test/kwargs.c>.

Freedback is welcome.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -870,10 +870,66 @@ MRB_API struct RClass * mrb_define_module_under(mrb_state *mrb, struct RClass *o
  * | `*`  | rest arguments | {mrb_value} *, {mrb_int} | Receive the rest of arguments as an array; `*!` avoid copy of the stack.  |
  * | <code>\|</code> | optional     |                   | After this spec following specs would be optional. |
  * | `?`  | optional given | {mrb_bool}        | `TRUE` if preceding argument is given. Used to check optional argument is given. |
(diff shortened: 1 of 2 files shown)


matz:
Since the keyword argument is a part of Ruby spec, I don't think we need a macro to enable it.

dearblue:
I rebase and forced push.
The `:` specifier is now always available.

PR #4758, opened 2019-10-07
Implement Ruby2.7's frozen strings from `Module#name`

shuujii:


shuujii:
I changed to reduce `mrb_str_dup`.

PR #4771, opened 2019-10-13
include mruby.h before checking MRB_WITHOUT_FLOAT

takkaw:


matz:
`MRB_WITHOUT_FLOAT` may be defined in `mrbconf.h` which is included from `mruby.h`.
This change requires `MRB_WITHOUT_FLOAT` to be defined in `CFLAGS`.

takkaw:
> MRB_WITHOUT_FLOAT may be defined in mrbconf.h which is included from mruby.h.

Yes, I think so too.
Some files include mrbconf.h through including mruby.h to check MRB_WITHOUT_FLOAT.
But, including mruby.h is above checking MRB_WITHOUT_FLOAT in math.c and fmt_fp.c.


> This change requires MRB_WITHOUT_FLOAT to be defined in CFLAGS.

No. It's OK to be defined MRB_WITHOUT_FLOAT in mrbconf.h after this change.
Rather, MRB_WITHOUT_FLOAT is needed to be defined in CFLAGS without this change.

matz:
I am sorry but probably I missed your point. Is it OK to define `MRB_WITHOUT_FLOAT` in `mrbconf.h` or not? It seems this PR prohibits `MRB_WITHOUT_FLOAT` in `mrbconf.h`.

takkaw:
I intended that it's OK to define MRB_WITHOUT_FLOAT in mrbconf.h.
I'm sorry, I might have made a mistake so please let me check.

dearblue:
When I proposed #4576, 'mrbconf.h' was missing completely.
Although it was late, I confirmed that if `MRB_WITHOUT_FLOAT` was defined in 'mrbconf.h', the expected error would occur.
Thanks to @takkaw!

PR #4779, opened 2019-10-18
Print mrbc command on verbose mode

take-cheeze:


matz:
Could you tell us your intention? Other commands, e.g. CC does not print on verbose mode. Why `mrbc` only?

take-cheeze:
`sh` method prints the command:
https://github.com/mruby/mruby/blob/20d01f118ddb7e7f2f36926a7a3db35573611857/minirake#L327
Though `popen` won't

matz:
OK, now I understand.

PR #4789, opened 2019-10-25
Always add trailing `.0` in `Float#inspect`; ref #4225

shuujii:
Trailing `.0` is removed from `Float#to_s` and `Float#inspect` at
9d08025b. However, I think the more human-readable format is better
for `Float#inspect`.

For example, in the `Float#to_s` format, the failure message is not
well understood when testing values including types by `eql?` (e.g.
`Numeric#step` test.

  ```ruby
  assert "example" do
    exp = 1.0
    act = 1
    assert_operator(exp, :eql?, act)  #=> Expected 1 to be eql? 1.
  end
  ```

### DIFF
--- a/src/numeric.c
+++ b/src/numeric.c
@@ -195,19 +195,8 @@ integral_coerce_step_counter(mrb_state *mrb, mrb_value self)
  *  representation.
  */
 
-/* 15.2.9.3.16(x) */
-/*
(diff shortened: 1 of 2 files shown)


matz:
This PR is merged to `mryby3` branch (will be pushed soon).

PR #4823, opened 2019-11-14
Fix here document with EOF

shuujii:
#### Before this patch:

  ```
  $ bin/mruby -e 'p <<EOS
1
EOS'
  -e:4:0: can't find heredoc delimiter "EOS" anywhere before EOF
  -e:4:0: syntax error, unexpected $end, expecting tHEREDOC_END or tHD_STRING_PART or tHD_STRING_MID
  ```

  ```
  $ bin/mruby -e 'p <<EOS
1
EOS'
  "1\n"
  ```

### DIFF
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -4381,12 +4381,7 @@ parse_string(parser_state *p)
           }
         }
         if ((len-1 == hinf->term_len) && (strncmp(s, hinf->term, len-1) == 0)) {
-          if (c < 0) {
-            p->parsing_heredoc = NULL;
-          }
-          else {
-            return tHEREDOC_END;
-          }
+          return tHEREDOC_END;
         }
       }
       if (c < 0) {


mrbgems/mruby-compiler/core/parse.y
```diff
@@ -4381,12 +4381,7 @@ parse_string(parser_state *p)
           }
         }
         if ((len-1 == hinf->term_len) && (strncmp(s, hinf->term, len-1) == 0)) {
-          if (c < 0) {
-            p->parsing_heredoc = NULL;
-          }
-          else {
-            return tHEREDOC_END;
-          }
+          return tHEREDOC_END;
```
matz: Should it be the following, just in case parsing continues after `EOF` using `partial_hook`?
```C
if (c<0) {
  p->parsing_heredoc = NUL;
}
return tHEREDOC_END;
```

shuujii: Thank you for your comment.

I think `p->parsing_heredoc` will become `NULL` via `heredoc_end()`. At least the above code will crash at:

https://github.com/mruby/mruby/blob/db43738397c2236adbdd9750db971106d5335ab6/mrbgems/mruby-compiler/core/parse.y#L1276

matz: I see. Thank you for the input.

PR #4842, opened 2019-11-26
Embed short symbol name (<= 2 words - 2 bytes) to `struct symbol_name`

shuujii:


matz:
Embedded symbols are already implemented.

PR #4853, opened 2019-12-06
Add `MRB_ALLOCV()` for temporary memory allocation

shuujii:
`MRB_ALLOCV(mrb, type, var, n)` defines `type *var` variable and assign
temporarily allocated memory for _n_ objects of type _type_. If the
allocation size is less than or equal to `MRB_ALLOCV_ON_STACK_MAX`, allocate
on the stack, otherwise use `mrb_alloca()`.

`MRB_ALLOCV_ON_STACK_MAX` is configurable in `mrbconf.h` (default is 256).

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -120,6 +120,9 @@
 /* fixed size state atexit stack */
 //#define MRB_FIXED_STATE_ATEXIT_STACK
 
+/* maximum size to allocate on the stack in MRB_ALLOCV  */
+//#define MRB_ALLOCV_ON_STACK_MAX 256
+
 /* -DMRB_DISABLE_XXXX to drop following features */
 //#define MRB_DISABLE_STDIO /* use of stdio */
 
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1355,6 +1355,45 @@ MRB_API void mrb_show_copyright(mrb_state *mrb);
 
 MRB_API mrb_value mrb_format(mrb_state *mrb, const char *format, ...);
 
+/*
+ * MRB_ALLOCV(mrb, type, var, n)
+ *
+ * Define `type *var` variable and assign temporarily allocated memory for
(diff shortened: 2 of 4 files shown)


matz:
Thank you for this PR. After review, I'd like to close it without merging, for these reasons:

1. **mruby targets embedded environments with tight C stack budgets.** We generally prefer explicit small `char onstack[N]` buffers, where the stack usage is directly visible to a code reader, over abstraction macros that hide platform-specific stack allocation behind preprocessor branches.

2. **Implementation issues with the macro:**
   - In the VLA branch (when `__STDC_NO_VLA__` is not defined), `type buf[n]` is allocated **before** the runtime size check. A large `n` blows the C stack even though the macro intends to route to `mrb_alloca` for that case.
   - The GCC version gate for `__builtin_alloca_with_align` is incorrect — that builtin requires GCC 6.1+, but the check is `__GNUC__ >= 4 && __GNUC_MINOR__ >= 7`.
   - `n` is evaluated multiple times in the macro expansion (in `sizeof(type)*(n)` and again in the on-stack expression), so callers with side-effecting expressions (`MRB_ALLOCV(..., x++)`) would behave incorrectly.
   - `_MRB_ALLOCV` and `_MRB_ALLOCV_ON_STACK_BUFFER` are reserved identifiers (file-scope names beginning with `_` followed by uppercase).

3. **Architectural fit:** the two call sites this PR modifies (`mrb_id_attrset` in mruby-struct, `prepare_name_common` in src/class.c) have kept their explicit `char onstack[32]` + dispatch pattern for the past 6+ years. Other large-allocation sites (`mruby-random`, `mruby-array-ext`) use bare `mrb_alloca` (which is now an alias for `mrb_temp_alloc` — arena-managed heap, not real stack). Both patterns make the trade-off explicit at each call site, which fits mruby's "make stack usage visible" stance better than a hidden-dispatch macro.

Thanks again for the contribution. Closing without merge.

PR #4863, opened 2019-12-12
Add `mrb_num_args_error()` for "wrong number of arguments" error

shuujii:
To unify the style of messages.


dearblue:
A little comment.

CRuby has `rb_error_arity ()` and `rb_check_arity ()` functions, so I would be happy if you could name them together.

dearblue:
Sorry.

After checking `rb_error_arity()` and `rb_check_arity()`, it was declared in `include/ruby/intern.h`, so it was an internal function.

Forget the previous comment.

PR #4864, opened 2019-12-12
Enable sanitizer in travis test

take-cheeze:


take-cheeze:
Seems to get leak report in `mruby -h` will check

PR #4867, opened 2019-12-13
Set alloca size as string capa for debug purpose

take-cheeze:


matz:
Could you please tell us the reason behind the change? The `RString` struct will not be accessed but from `mrb_gc_free_str` and the function does not refer `capa `.

matz:
Thank you for this PR. Closing it without merge after re-reviewing — my December 2019 question about the rationale never received a follow-up, and the situation hasn't changed since.

Recap of the concern: the `RString` constructed by `mrb_alloca` (now `mrb_temp_alloc`) is a *fake string* used purely as a heap container — its `heap.ptr` points to the user buffer, and the surrounding object only exists so the arena cleanup machinery will free it later. Nothing reads `aux.capa` for these objects. Setting `capa = size` would add a field-write that no current code consumes; the change is harmless today but introduces a maintenance signal ("here's a value we set, but for what?") without an accompanying use site.

Arguably the *absence* of `capa` initialization is a deliberate marker that these objects are not full-fledged strings — useful as a defensive signal if they're ever accessed by mistake from string code paths.

If a future debugging tool (heap dumper, gdb introspection helper, etc.) adopts a contract of reading `aux.capa` for `mrb_temp_alloc` buffers, this can be reintroduced at that point with a comment naming the consumer. Until there's a use site, leaving the field zero is cleaner.

Closing in favor of "introduce when needed." Thanks again for the work.

PR #4868, opened 2019-12-13
Fix ownership error in vformat test

take-cheeze:


matz:
I don't see the reason to protect the `%C` class only. Does this PR fix any real problem?

matz:
I don't see the reason to protect the `%C` class only. Does this PR fix any real problem?

take-cheeze:
This bothers me when testing with `-DMRB_GC_STRESS`.
And object under `RData`'s memory don't get marked so it's potentially unsafe operation anyway.

shuujii:
This issue seems to be resolved by #4622. Will it not be merged?

matz:
I don't think `%C` (class/module) needs to be saved in `TestVFormat::Native`. `%C` should be handled like `%t`.

Issue #4618, opened 2019-08-06
`mrb_gc_unregister()` may access freed memory


shuujii:
`mrb_gc_unregister()` may be called from `mrb_gc_destroy()` in `mrb_close()`, but the registered list (global variable) has already freed by previous `mrb_gc_free_gv()`.

matz:
I didn't find the call path from `mrb_gc_destroy` to `mrb_gc_unregister`.  Did I miss something?

shuujii:
Sorry for the lack of explanation.

This is a problem when we create custom class and call `mrb_gc_unregister()` in `mrb_data_type::dfree`.

```text
mrb_gc_destroy
-> free_heap
  -> obj_free
    -> mrb_data_type::dfree
      -> mrb_gc_unregister
```

There seems to be such a usage in H2O:

https://github.com/h2o/h2o/blob/af171560cfa21d16a65da61f91bb7201d094e20e/lib/handler/mruby/channel.c#L37-L45

matz:
I clear `mrb->globals` in `mrb_gc_free_gv` so that access to the global variables should not cause any errors. It should cover this issue too.

shuujii:
`mrb_intern_lit` is called from `mrb_gc_unregister`, but `mrb_free_symtbl` is called before `mrb_gc_destroy`. This seems to be a similar problem.

PR #4883, opened 2019-12-19
Retrieve `mrb_vfromat()` tests at 1e9cb74c

shuujii:


matz:
I am not going to simply revert the change. Show us your fix.

shuujii:
> Show us your fix.

My fix is this PR.

shuujii:
Sorry, the subject of PR might have been confusing. This PR fixes to support `%l` test without code generation.

matz:
I understand. Let me see your code for `%l` support.

matz:
My opinion (the current master (matz) vs this PR (shuuji)):
* matz supports `%l` naturally
* shuuji supports the following:
  * some combination of type specifiers
  * some implementation_dependant tests
 * I am not sure if above tests are absolutely needed

If those combination tests and implementation-dependent tests, I will implement them on top of the master.

shuujii:
* In the first place, I just made the `%l` test temporarily disabled at 1e9cb74c work again, and have not changed the test itself before that commit ([diff: test/t/vformat.rb](https://github.com/mruby/mruby/compare/a4243360..78e5a46#diff-6cbfe9124f1e05525955a5c218c482df)).

* I think my implementation also naturally supports `%l`, so I am not sure the essential difference from the matz implementation. My implementation is a method of specifying C type from Ruby (`n.d(1)` is `int`, `n.s('str')` is `char*`, `n.l(2)` is `size_t`, etc).

* My implementation supports all specifiers combination up to 2 arguments.

* I think that it is necessary to test patterns that are easy to make mistakes, such as when multiple specifiers, ending with `%`, etc. At least I can't always implement without mistakes when refactoring, improvement, etc. The implementation_dependant tests are also intended to show that users should not depend on these behaviors (e.g. [ending with `%`in Ruby](https://twitter.com/igaiga555/status/1202106036338884608))

* I think `mrb_gc_register()` and `mrb_gc_unregister()` are important APIs (used in H2O, etc). However, they are not used in the core and no tests, so I intentionally use them in this PR. Actually, I found a bug (#4618).

PR #4885, opened 2019-12-20
Fix potentially crash in `%n` of `mrb_vformat()` with 64-bit `int`

shuujii:
If `mrb_sym` is smaller than `int`, it is promoted to `int`.

### DIFF
--- a/src/error.c
+++ b/src/error.c
@@ -302,7 +302,11 @@ mrb_vformat(mrb_state *mrb, const char *format, va_list ap)
           mrb_gc_arena_restore(mrb, ai);
           break;
         case 'n':
+#if UINT32_MAX < INT_MAX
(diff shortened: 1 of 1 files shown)


src/error.c
```diff
@@ -302,7 +302,11 @@ mrb_vformat(mrb_state *mrb, const char *format, va_list ap)
           mrb_gc_arena_restore(mrb, ai);
           break;
         case 'n':
+#if UINT32_MAX < INT_MAX
+          obj = (mrb_sym)mrb_symbol_value(va_arg(ap, int));
```
matz: I thin the line here should be:
``` c
obj = mrb_symbol_value((mrb_sym)va_arg(ap, int));
```
Right?

shuujii: You are right, thank you! I will fix it.

matz:
OK, thank you!

PR #4887, opened 2019-12-20
Support `-v` flag of rake

take-cheeze:


Rakefile
```diff
@@ -5,6 +5,10 @@ MRUBY_ROOT = File.dirname(File.expand_path(__FILE__))
 MRUBY_BUILD_HOST_IS_CYGWIN = RUBY_PLATFORM.include?('cygwin')
 MRUBY_BUILD_HOST_IS_OPENBSD = RUBY_PLATFORM.include?('openbsd')
 
+if Object.const_defined? :Rake
```
shuujii: I think `Rake` is always defined because there is `Rake = MiniRake` in `minirake`.

PR #4902, opened 2019-12-23
Handle CR LF newline natively in lexer

take-cheeze:


mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c
```diff
@@ -183,7 +183,7 @@ partial_hook(struct mrb_parser_state *p)
     return -1;
   }
   fn = args->argv[args->idx++];
-  p->f = fopen(fn, "r");
+  p->f = fopen(fn, "rb");
```
matz: Does "rb" mode conform C99 standard? I know it works for Linux, but not sure about other platforms.

matz: I fount the following description in the Linux man page for `fopen`:

> The mode string can also include the letter 'b' either as a last character or as a character between the characters in any of the two-character strings described above.  This is strictly for compatibility with C89 and has no effect; the 'b' is ignored on all POSIX conforming systems, including Linux.  (Other systems may treat text files and binary files differently, and adding the 'b' may be a good idea if you do I/O to a bi‐ nary file and expect that your program may be ported to non-UNIX environments.)

That means C89 (thus C99 as well) requires `fopen` to understand "rb" mode.

take-cheeze: Just some "kusoripu" : https://github.com/mruby/mruby/blob/18248cfd000cac220c17d7b3f2b49bbc0e0d90bd/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c#L311

PR #4907, opened 2019-12-25
Fix potentially use of wrong method cache

shuujii:
#### Example (with `MRB_METHOD_CACHE`)

  ```ruby
  GC.start
  c = Class.new
  p c            #=> #<Class:0x7fd6a180e790>
  c.new          #=> cache `c.new`
  c = nil
  GC.start       #=> `c` is GCed
  r = Range.dup
  p r            #=> #<Class:0x7fd6a180e790>
                 #   [same pointer as `c`]
  r.new(2, 3)    #=> ArgumentError: 'initialize':
                 #   wrong number of arguments (2 for 0)
                 #   [`c.new` is called instead of `r.new`]
  ```

#### Cause

  An entry of method cache is identified by class pointer and method
  id. However, reusing memory after GC may create a class with the same
  pointer as the cached class.

#### Treatment

  Cleared method caches of the class when the class is GCed.

### DIFF
--- a/include/mruby/class.h
+++ b/include/mruby/class.h
@@ -95,6 +95,12 @@ void mrb_gc_mark_mt(mrb_state*, struct RClass*);
 size_t mrb_gc_mark_mt_size(mrb_state*, struct RClass*);
 void mrb_gc_free_mt(mrb_state*, struct RClass*);
 
+#ifdef MRB_METHOD_CACHE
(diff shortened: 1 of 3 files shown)


matz:
First I worried little about performance impact. But classes and modules are rarely collected, so I will merge this PR as it is.

PR #4908, opened 2019-12-25
Introduce Encoding core mrbgem

take-cheeze:
For better scripting side UTF-8 support


matz:
The PR is not very useful for mruby, since it does not convert encoding at all. What is the motivation behind the PR? Compatibility?

matz:
At least for the time being, we are not going to add `Encoding` to `mruby`.

PR #4909, opened 2019-12-26
Enable `MRB_GC_STRESS` test on CI

shuujii:
The `debug` build target (`MRB_GC_STRESS` is enabled) on CI have been
compiled but not tested so far. However, I think testing with
`MRB_GC_STRESS` is effective (in fact, I found #4907 bug).

Therefore, I integrated `debug` and `full-debug` build targets to enable
`MRB_GC_STRESS` testing. Testing with `MRB_GC_STRESS` takes a little time,
but compiling takes more time, so CI execution time does not increase due to
decrease of build target.

### DIFF
--- a/appveyor_config.rb
+++ b/appveyor_config.rb
@@ -1,23 +1,10 @@
-MRuby::Build.new('debug') do |conf|
-  toolchain :visualcpp
-  enable_debug
-
-  # include all core GEMs
-  conf.gembox 'full-core'
-  conf.compilers.each do |c|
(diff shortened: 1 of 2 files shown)


matz:
PR merged.

shuujii:
In my PR I integrated the build target `debug` into `full-debug`, but the `debug` target remains in `travis_config.rb`. Is this intentional?

matz:
It's not intentional. Should I remove `debug` target from `travis_config.rb`?

shuujii:
I think `debug` target is unneeded because its configuration is almost same as `full-debug` target (it has been removed from `appveyor_config.rb`).

matz:
Understood. Done in f7abda14.

PR #4917, opened 2020-01-01
Call `va_end()` before return

dearblue:
The behavior when returning from a function without `va_end()` is undefined.

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -897,7 +897,7 @@ mrb_get_args(mrb_state *mrb, const char *format, ...)
       }
       break;
     case '|':
-      if (opt_skip && i == argc) return argc;
+      if (opt_skip && i == argc) goto finish;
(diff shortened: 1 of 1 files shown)


src/class.c
```diff
@@ -1023,8 +1023,10 @@ mrb_get_args(mrb_state *mrb, const char *format, ...)
   if (!c && argc > i) {
     mrb_raise(mrb, E_ARGUMENT_ERROR, "wrong number of arguments");
   }
+
+finish:
   va_end(ap);
-  return i;
+  return argc;
```
matz: Is there any reason to replace `i` with `argc` here? We checked `i == argc` before jumping to `finish`.

dearblue: That is my mistake. sorry.
I will fix it.

PR #4919, opened 2020-01-01
Revert "SHARED string is not required when sharing POOL string" (75949836)

shuujii:
Because literal pool may be released by GC.

#### Example:

  ```ruby
  s1 = eval('"abcdefghijklmnopqrstuvwxyz01"')
  GC.start
  p s1  #=> "\x00\x00\x00\x00\x00\x00\x00\x90\x00\x00\x00\x00\x00\x00\x00\x90\x03\x00stuvwxyz01"
  ```


shuujii:
I am terribly sorry for introducing the serious bug to mruby 2.1.0.

PR #4920, opened 2020-01-04
Avoid creating temporary objects in `read_irep_record_1()`

dearblue:
Since the irep pool element is now NUL terminated, break compatibility for mruby binary format.

### DIFF
--- a/include/mruby/string.h
+++ b/include/mruby/string.h
@@ -448,6 +448,7 @@ MRB_API int mrb_str_cmp(mrb_state *mrb, mrb_value str1, mrb_value str2);
 MRB_API char *mrb_str_to_cstr(mrb_state *mrb, mrb_value str);
 
 mrb_value mrb_str_pool(mrb_state *mrb, mrb_value str);
+mrb_value mrb_str_pool_from_str_len(mrb_state*, const char*, size_t, mrb_bool);
 uint32_t mrb_str_hash(mrb_state *mrb, mrb_value str);
 mrb_value mrb_str_dump(mrb_state *mrb, mrb_value str);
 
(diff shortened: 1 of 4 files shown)


matz:
I like the idea, but I worry about the compatibility issue. Let me investigate the idea for a while.

shuujii:
Null termination seems necessary for `mrb_cstr_to_dbl` to request it.

In most cases, if null termination is performed on the area allocated by `MRB_ALLOCV` (#4853), it can be realized without heap allocation or object creation, I think.

Alternatively, we may be able to use `mrb_str_to_dbl` with the following treatments:

1. Perform null termination using `MRB_ALLOCV` instead of `RSTRING_CSTR` in `mrb_str_to_dbl`.
2. Introduce something like CRuby's fake string (using `RString` on the stack).

FYI, there is the following bug (unreported). I think this can be fixed via 1.

  ```console
  $ ruby -e 'p "1\x002".to_f'
  1.0
  ```

  ```console
  $ mruby -e 'p "1\x002".to_f'
  -e:1: string contains null byte (ArgumentError)
  ```

dearblue:
Here are some things I want to appeal to in this proposal:

 1. Work object creation is reduced (= GC is reduced).
 2. Reduced transfer between memories.
 3. String literals are already NUL-terminated and can be extracted as C strings.

(1) and (2) can be especially effective on slow devices because they affect irep read time.
It's now that I came up with (3). :smile_cat:

matz:
111045e is my attempt to address this issue that:
* Avoid creating temporary strings
* Do not change the binary format
* `"1\x002".to_f` works OK
* `Float("0x10")` now works

dearblue:
I think it is a good change. Thank you!

PR #4933, opened 2020-01-19
Fix take over file scope variables with `mruby` and `mirb` command

dearblue:
- Add `mrbc_cleanup_local_variables()` function with `MRB_API`
- In `mirb` command, the `_` variable is defined after the `-r` switch

### DIFF
--- a/include/mruby/compile.h
+++ b/include/mruby/compile.h
@@ -42,6 +42,7 @@ MRB_API mrbc_context* mrbc_context_new(mrb_state *mrb);
 MRB_API void mrbc_context_free(mrb_state *mrb, mrbc_context *cxt);
 MRB_API const char *mrbc_filename(mrb_state *mrb, mrbc_context *c, const char *s);
 MRB_API void mrbc_partial_hook(mrb_state *mrb, mrbc_context *c, int (*partial_hook)(struct mrb_parser_state*), void*data);
+MRB_API void mrbc_cleanup_local_variables(mrb_state *mrb, mrbc_context *c);
 
 /* AST node structure */
 typedef struct mrb_ast_node {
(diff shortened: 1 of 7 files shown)


shuujii:
The following example doesn't seem to work as expected:

#### Files

  ```ruby
  # a.rb
  a = 1
  A = ->{a}
  ```

  ```ruby
  # b.rb
  p A.call
  ```

#### mruby

  ```console
  $ bin/mruby -r ./a.rb b.rb
  main
  ```

#### Ruby

  ```console
  $ ruby -r ./a.rb b.rb
  1
  ```

dearblue:
:crying_cat_face: That's true.

I was worried and tried `mrbc`, and it was the same without applying this patch.

```console
% bin/mrbc a.rb
% bin/mrbc b.rb
% bin/mruby -br a.mrb b.mrb
main
```

It seems that can deal with it using `REnv`, but I may not be able to handle it...

matz:
@dearblue It works as CRuby (`1`) on my box.

dearblue:
On FreeBSD 12.1/amd64, after slightly modifying `mrbgems/mruby-print/mrblib/print.rb` of mruby-head (59464f3), the result is as follows.

```console
% git checkout --detach origin/master
HEAD is now at 59464f35 Merge pull request #4935 from mimaki/delete-duplicated-Array-delete_if
% echo 'a,b,c,d,e,f,g,h,i,j,k,l,m,n=1,2,3,4,5,6,7,8,9,10,11,12,13,14; A=->{[a,b,c,d,e,f,g,h,i,j,k,l,m,n]}' >> mrbgems/mruby-print/mrblib/print.rb
% rake -m clean all
...
% bin/mruby -e 'a=999; b=123; c="ccc"; p A.call'
[999, 123, "ccc", main, main, nil, 999, 123, "ccc", main, main, nil, 999, 123]
```

dearblue:
I fixed the problem pointed out by @shuujii and the problem with mruby binary.

dearblue:
I made a mistake.

PR #4950, opened 2020-03-07
Absolute path for windows

Reckordp:
On mirb
```
File.expand_path("~")
```

Result in windows
```
(mirb):1: non-absolute home (ArgumentError)
```

I have added some code for this. I like it when combined :)

### DIFF
--- a/mrbgems/mruby-io/src/file.c
+++ b/mrbgems/mruby-io/src/file.c
@@ -282,7 +282,12 @@ mrb_file__getwd(mrb_state *mrb, mrb_value klass)
 static int
 mrb_file_is_absolute_path(const char *path)
 {
+#ifdef _WIN32
+  if (strlen(path) < 2) return 0;
+  return path[0] >= 64 && path[0] <= 90 && path[1] == ':'; // 64 to 90 is ASCII
+#else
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-io/src/file.c
```diff
@@ -282,7 +282,12 @@ mrb_file__getwd(mrb_state *mrb, mrb_value klass)
 static int
 mrb_file_is_absolute_path(const char *path)
 {
+#ifdef _WIN32
+  if (strlen(path) < 2) return 0;
+  return path[0] >= 64 && path[0] <= 90 && path[1] == ':'; // 64 to 90 is ASCII
```
mattn: It is not enoguh to check paths with slash or backslash. And "c:" is not relative paths on Windows.
```c
#define IS_PATHSEP(x) (x == '/' || x == '\')
    return (isalpha(path[0]) && path[1] == ':' && IS_PATHSEP(path[2]));
#undef IS_PATHSEP
```
Also if you want to handle UNC paths, it should check more strictly.

Reckordp: I will adjust it like the source in ruby. just found out about udc thanks

PR #4951, opened 2020-03-07
Fix unfinished changes

Reckordp:
Me again
There was a typo resulting in an error during compile.

```
#define IS_PATHSEP(x) (x == '/' || x == '\')
```
to
```
#define IS_PATHSEP(x) (x == '/' || x == '\\')
```

Just that.

### DIFF
--- a/mrbgems/mruby-io/src/file.c
+++ b/mrbgems/mruby-io/src/file.c
@@ -283,7 +283,7 @@ static int
 mrb_file_is_absolute_path(const char *path)
 {
 #ifdef _WIN32
-#define IS_PATHSEP(x) (x == '/' || x == '\')
+#define IS_PATHSEP(x) (x == '/' || x == '\\')
   if (strlen(path) < 3) return 0;
   return (isalpha(path[0]) && path[1] == ':' && IS_PATHSEP(path[2]));
 #undef IS_PATHSEP
--- a/mrbgems/mruby-io/src/file.c
+++ b/mrbgems/mruby-io/src/file.c
@@ -284,8 +284,10 @@ mrb_file_is_absolute_path(const char *path)
 {
 #ifdef _WIN32
 #define IS_PATHSEP(x) (x == '/' || x == '\\')
-  if (strlen(path) < 3) return 0;
-  return (isalpha(path[0]) && path[1] == ':' && IS_PATHSEP(path[2]));
+  if (isalpha(path[0]))
+    return (strlen(path) > 2 && path[1] == ':' && IS_PATHSEP(path[2]));
+  else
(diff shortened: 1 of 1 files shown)


Reckordp:
also added the absolute path for the ANC case

matz:
Wait, how should `mrb_file_is_absolute_path("\\")` return? true or false?

Reckordp:
```
set USERPROFILE=\
mirb
> File.expand_path("~")
non-absolute home (ArgumentError)
```
`mrb_file_is_absolute_path("\\")` return false

```
set USERPROFILE=\\
mirb
> File.expand_path("~")
"/"
```
`mrb_file_is_absolute_path("\\\\")` return true

matz:
Sorry for my ignorance. So, in Windows, `\` is not an absolute path, right?
In that case, the current implementation access the trailing `NUL` (L290) in `path`. Is it intentional?

Reckordp:
first I just wanted to solve my own problem and share it with people, but it seemed that the problem was getting bigger: p

Sorry, this is a translation from Google

according to the documentation from windows that i read, there are three types of paths: traditional dos, unc, and dos devices.

only the traditional dos pathway has a relative path.

traditional path absolute dos must include drive letters, colons, and separators. for unc double separator without dots and question marks that follow. for multiple dos separator devices, followed by a period and a question mark.

That's all I know, I'll fix my program style.

mrbgems/mruby-io/src/file.c
```diff
@@ -283,9 +283,11 @@ static int
...
 {
 #ifdef _WIN32
-#define IS_PATHSEP(x) (x == '/' || x == '\')
-  if (strlen(path) < 3) return 0;
-  return (isalpha(path[0]) && path[1] == ':' && IS_PATHSEP(path[2]));
+#define IS_PATHSEP(x) (x == '/' || x == '\\')
+  if (isalpha(path[0]))
+    return (strlen(path) > 2 && path[1] == ':' && IS_PATHSEP(path[2]));
+  else
+    return (IS_PATHSEP(path[0]) && IS_PATHSEP(path[1]));
```
kou: It seems that we need to add `path[0] != '\0'` check.

dearblue: I think that there is no problem as it is...

Reckordp: watching over the length of the string might be more appropriate. when trying to retrieve `path[1]` it will collapse.

Reckordp: Pull Request #4952

kou: Ah, sorry. I was wrong. `path[0] != '\0'` is needless.

PR #4952, opened 2020-03-08
Traditional, UNC, and Drive Path

Reckordp:
added several types of paths for windows only
and some additions :D

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,3 +42,6 @@ of this list.
    Masahiro Wakame
    YAMAMOTO Masaya
    KOBAYASHI Shuji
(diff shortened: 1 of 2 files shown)


shuujii:
I think that the argument `len` of `mrb_file_is_absolute_path()` is unnecessary because differences in the signature of the function between Windows and others cause confusion.

And the type of `len` should be `size_t`, I think.

shuujii:
Sorry, I didn't realize @matz had already made the same fix as my comment. Thank you.

Reckordp:
Nice

Reckordp:
Nice

PR #4953, opened 2020-03-08
Some rule to CrossBuild Android

Reckordp:
usually there are things that are not the same as others have. the addition of some smooth rules on my device


matz:
Next time, separate the PR for each concern (e.g. patches for `mrbgem.rake` and `android.rake` in this case), please.

Reckordp:
Sure

PR #4959, opened 2020-04-01
build when directories and files have spaces

dsisnero:
Modified the build to quote filenames so that it builds when files have spaces

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -36,9 +36,14 @@ module MRuby
 
     private
     def _run(options, params={})
-      return sh command + ' ' + ( options % params ) if NotFoundCommands.key? @command
+      
+      cmd_string = if NotFoundCommands.key?(@command)
+                     command + ' ' + ( options % params )
+                   else
+                     build.filename(command) + ' ' + ( options % params )
+                   end
       begin
-        sh build.filename(command) + ' ' + ( options % params )
+        sh cmd_string
       rescue RuntimeError
         NotFoundCommands[@command] = true
         _run options, params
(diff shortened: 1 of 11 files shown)


lib/mruby/build/command.rb
```diff
@@ -36,9 +36,14 @@ def shellquote(s)
...
+                     command + ' ' + ( options % params )
+                   else
+                     build.filename(command) + ' ' + ( options % params )
+                   end
       begin
-        sh build.filename(command) + ' ' + ( options % params )
+        sh cmd_string
       rescue RuntimeError
         NotFoundCommands[@command] = true
         _run options, params
```
shuujii, line 49: Is the logic changed intentionally?

With this change, it will be an infinite loop if neither `command` nor `build.filename(command)` exists, I think.

shuujii: BTW, here, we just need to execute `build.filename(command)`, I think.

It seemed to fall back to `command` to support `ccache gcc`, etc. because previously `build.filename(command)` quoted the whole (9968af45). I don't think it is necessary to fall back to `command` because `build.filename(command)` is not quoted now.

shuujii: If there is a compile error, it will also loop infinitely.

```console
$ echo 1 > src/a.c
$ rake -v
mkdir -p /mruby/mruby/build/host/src
gcc -std=gnu99 -g -O3 -Wall -Wundef -Wdeclaration-after-statement -Werror-implicit-function-declaration -Wwrite-strings -I"/mruby/mruby/include" -MMD -o "/mruby/mruby/build/host/src/a.o" -c "/mruby/mruby/src/a.c"
/mruby/mruby/src/a.c:1:1: error: expected identifier or '('
1
^
1 error generated.
gcc -std=gnu99 -g -O3 -Wall -Wundef -Wdeclaration-after-statement -Werror-implicit-function-declaration -Wwrite-strings -I"/mruby/mruby/include" -MMD -o "/mruby/mruby/build/host/src/a.o" -c "/mruby/mruby/src/a.c"
/mruby/mruby/src/a.c:1:1: error: expected identifier or '('
1
^
1 error generated.
gcc -std=gnu99 -g -O3 -Wall -Wundef -Wdeclaration-after-statement -Werror-implicit-function-declaration -Wwrite-strings -I"/mruby/mruby/include" -MMD -o "/mruby/mruby/build/host/src/a.o" -c "/mruby/mruby/src/a.c"
/mruby/mruby/src/a.c:1:1: error: expected identifier or '('
1
^
1 error generated.
...
```

PR #4980, opened 2020-04-26
Reduce `struct mrb_context`

dearblue:


shuujii:
In the case of MSVC, the size doesn't seem to be reduced just by making it a bit field. It seems that it is necessary to make bit field members all the same type, so it becomes smaller if the type is changed to `uint8_t`.

It is also even smaller if the structure members are sorted by the size of the type.

#### Examples:
* [MSVC](https://ci.appveyor.com/project/shuujii/windows/build/job/kp9j5jeu68bpbyj4)
* [GCC](https://wandbox.org/permlink/mfGunwL0jTHMBukX)
* [Clang](https://wandbox.org/permlink/f6gHNJbQyQQv1CHc)

matz:
The effectiveness of bit fields is implementation-dependent. I myself don't care that much. But if OP cares, PR will be welcome.

dearblue:
I now plan to remove `ecall()` and embed the `ensure` block into the same irep as the `begin` (as well as the `rescue`).

This will merge the "exception handler stack" and "ensure handler stack" in the `struct mrb_context`.

https://github.com/mruby/mruby/blob/0f229bfbbf4fe5a8b7faf802750bda60096b9409/include/mruby.h#L162-L165

I'm having trouble with `break` and `return` in the `yield` across methods, so it's sad that it's not ready to publish soon...

matz:
It seemed so hard when I designed the VM. If you can remove `ecall`, it's awesome!

PR #4981, opened 2020-04-27
Test mruby-io in tmpdir when AF_UNIX cannot be created on cwd

udzura:
mruby-io cannot create temporary AF_UNIX socket when working directory is vboxfs or NFS.

This is a workaround to test in such kind of filesystems.

### DIFF
--- a/mrbgems/mruby-io/test/mruby_io_test.c
+++ b/mrbgems/mruby-io/test/mruby_io_test.c
@@ -1,7 +1,6 @@
 #include <mruby/common.h>
 #include <sys/types.h>
 #include <errno.h>
-#include <assert.h>
 
 #if defined(_WIN32) || defined(_WIN64)
 
@@ -54,11 +53,12 @@ mkdtemp(char *temp)
   #include <sys/socket.h>
   #include <unistd.h>
   #include <sys/un.h>
+  #include <assert.h>
+  #include <fcntl.h>
 #endif
 
 #include <sys/stat.h>
 #include <stdlib.h>
-#include <fcntl.h>
 
 #include "mruby.h"
 #include "mruby/array.h"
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-io/test/mruby_io_test.c
```diff
@@ -80,6 +113,16 @@ mrb_io_test_io_setup(mrb_state *mrb, mrb_value self)
...
+
+  if(!(socket_available_p = mrb_io_socket_available())) {
+    char *tmpdir;
+    wd_save = open(".", O_DIRECTORY);
+    tmpdir = getenv("TMPDIR");
+    if (tmpdir)
+      assert(!chdir(tmpdir));
+    else
+      assert(!chdir("/tmp"));
+  }
```
shuujii, line 125: I think the side effect in assertions are not good because it changes the behavior depending on whether it is in debug mode or not.

In the first place, it's better not to change the directory if possible, but at least it should go back to the original directory, I think.

FYI: It may be better to use `mrb_assert()` instead of `assert()`.

shuujii: > In the first place, it's better not to change the directory if possible, but at least it should go back to the original directory, I think.

I'm sorry, I misunderstood about the following, so please ignore the quoted comments.
* Moves to a temporary directory when some tests don't work in the current directory.
* Returns to the original directory via `mrb_io_test_io_cleanup()`.

shuujii:
Running the test leaves `tmp.mruby-io-socket-ok.????????`. The return value of `mkstemp()` is an error when it is `-1` (negative), so I think that the check is inappropriate.

mrbgems/mruby-io/test/mruby_io_test.c
udzura: Yeah, `assert(3)` is ZATSU... I expected that I got some review

udzura: `mrb_assert()` looks better

PR #4984, opened 2020-04-29
Sorting the list of gems at build time by name

dearblue:
Printing them in sorted order makes it easier to find the desired gem.
But it has come to completely ignore the dependency.


matz:
I am not sure about your intention. Sorting is better because we can find gems by names easier, but it has the drawback of losing dependency info. So you are asking for my decision, right?

dearblue:
The only thing that is aligned is the display.
The inside information is not affected at all.

As for the display so far, I could only predict that it would depend on certain items being on the upper side.
I don't think there's much point in that.

PR #4948, opened 2020-03-04
Enable MRB_METHOD_T_STRUCT by default on 32bit Windows

kou:
Because we can't use the highest 2 bits of function pointers.

I couldn't reproduce this with the existing tests.
https://ci.appveyor.com/project/kou/mruby/builds/31199291/job/tkh6ady3xwxnvogj works well without `MRB_METHOD_T_STRUCT`.

But I added some 32bit Windows entries to `appveyor.yml`. I don't add "Visual Studio 2013 32bit" case because it failed with `Math.atan2` test:
https://ci.appveyor.com/project/kou/mruby/builds/31199291/job/tkh6ady3xwxnvogj

```text
Fail: Math.atan2 (mrbgems: mruby-math)
 - Assertion[5]
    Expected NaN to be 2.356194490192345.
 - Assertion[6]
    Expected NaN to be -2.356194490192345.
 - Assertion[7]
    Expected NaN to be 0.7853981633974483.
 - Assertion[8]
    Expected NaN to be -0.7853981633974483.
```

It's not related to this case.

I couldn't reproduce this with existing tests but our product, Groonga, is crashed without `MRB_METHOD_T_STRUCT`:
https://ci.appveyor.com/project/groonga/groonga/builds/31118559/job/8i0hq7crg0rpaucw#L7297

```text
#|C| c:\projects\groonga\vendor\mruby-source\src\vm.c:525:0: mrb_funcall_with_block(): <libgroonga>: <C:\projects\groonga\install\groonga-10.0.0-b5d38cc-x86-vs2017\bin\libgroo>
#|C| c:\projects\groonga\vendor\mruby-source\src\vm.c:441:60: mrb_funcall_with_block(): <libgroonga>: <C:\projects\groonga\install\groonga-10.0.0-b5d38cc-x86-vs2017\bin\libgroo>
#|C| c:\projects\groonga\vendor\mruby-source\src\vm.c:541:63: mrb_funcall_argv(): <libgroonga>: <C:\projects\groonga\install\groonga-10.0.0-b5d38cc-x86-vs2017\bin\libgroo>
#|C| c:\projects\groonga\vendor\mruby-source\src\class.c:1553:31: mrb_obj_new(): <libgroonga>: <C:\projects\groonga\install\groonga-10.0.0-b5d38cc-x86-vs2017\bin\libgroo>
[...]

PR #4990, opened 2020-05-04
Enable MRB_METHOD_T_STRUCT by default on 32bit GUN/Linux

komainu8:
Hello,

We can't also use the highest 2 bits of function pointers on 32bit GNU/Linux same as https://github.com/mruby/mruby/pull/4948.

How about we decide whether 32bit architecture by using `MRB_32BIT`?


matz:
I am not sure `MRB_32BIT` is a correct predicate. It sounds OK but not for sure. If we see any problem, we will revert the change.

komainu8:
I see.

PR #4992, opened 2020-05-06
Add integer division like 7 / 2 => 3

LanzaSchneider:


matz:
The integer division is the biggest difference from CRuby. Since we don't implement `Bignums` in mruby, we use `Float` numbers instead on overflow. So some integer calculations may result in float numbers. That's the reason we use float division by default instead of integer division.

The options for the future are:
1.  make division integer division by default (this PR; compatible with Ruby)
    1.1.  overflow should make float as it does now (inconsistency? maybe)
    1.2.  overflow should raise an exception (big limitation; acceptable?)
    1.3.  implement `Bignum` using `mpz` or some other library (need more RAM)
2. make division float division as it is now
    2.1. Python3 chose this change; CRuby can follow after mruby

Any opinion?

fourmond:
From my narrow perspective of someone who embeds mruby into a program for data analysis, the current status (option 2, i.e. float division) is much less confusing for my users than integer division. Some data plotting programs like gnuplot default to integer division, but it's the kind of things that cause severe headaches from time to time, even to someone like me who understands what type casting is about.

If you choose to make option 1 the default, is there any possibility for changing the default (back to 2) at mruby compile time ? Thanks a lot.

pulsejet:
For whatever my opinion might count for, I am primarily using mruby to port legacy code to the web (getting cruby to cross compile to wasm and actually work is a nightmare), so this is the only patch that I had to make to mruby for everything to be consistent. I don't really see the harm of having such a switch that will change integer division to cruby behavior (which will be turned off by default).

matz:
At the moment I started `mruby`, I had a plan to make Ruby integer division to result in float numbers, but due to the compatibility issue, I abandoned the plan. So there's little reason to keep this behavior as an experiment for Ruby's future.
So I will merge this pull-request to `mruby3` branch for the future.

PR #5006, opened 2020-05-23
Do not destruct rest arguments for __send__

mame:
Formerly, `__send__(*args)` modified `args` with `Array#shift`.
This bug affects optcarrot.

This changeset avoids the array destruction by using
`args = args[1, len-1]`.


mame:
This code demonstrates the bug behavior:

```ruby
def foo(x)
end

ary = [:foo, :bar]
p ary #=> [:foo, :bar]
send(*ary)
p ary #=> [:bar]
```

PR #5015, opened 2020-05-31
Fix `Fixnum` and `Float` comparison in `Hash` lookup

shuujii:
#### Before this patch:

  ```console
  $ bin/mruby -e 'p({1 => 2}.key?(1.0))'
  true
  ```

#### After this patch (same as Ruby)

  ```console
  $ bin/mruby -e 'p({1 => 2}.key?(1.0))'
  false
  ```

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -96,27 +96,13 @@ ht_hash_equal(mrb_state *mrb, htable *t, mrb_value a, mrb_value b)
     return mrb_symbol(a) == mrb_symbol(b);
 
   case MRB_TT_FIXNUM:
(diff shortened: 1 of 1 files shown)


shuujii:
I think current implementation does not follow ISO Ruby because ISO Ruby requires `eql?` method to be used for hash key matching and `1.0.eql?(1)` is `false`.

matz:
Yes, but we should fix (or change) the relationship of `Integer` and `Float` first. I am working on it.

matz:
Merged to `mruby3` branch.

PR #5016, opened 2020-05-31
Remove `patch_irep()` in `mruby-eval`

dearblue:
- It can now deal with operands in the range of `OP_EXT*`.
- It can now call the same method as the variable name without arguments.

  ```ruby
  def a
    "Safe!"
  end

  a = "Auto!"

  eval "a()" # call method `a`
  ```

----

It also includes changes in `mrbgems/mruby-compiler/core/y.tab.c`, so the difference is inflated by more than six times.
The net is as follows:
```console
% git diff --stat master.. -- :^mrbgems/mruby-compiler/core/y.tab.c
 include/mruby/compile.h               |   4 +-
 mrbgems/mruby-compiler/core/codegen.c | 101 +++++++++-------
 mrbgems/mruby-compiler/core/parse.y   |  15 ++-
 mrbgems/mruby-eval/src/eval.c         | 215 +---------------------------------
 mrbgems/mruby-eval/test/eval.rb       |  21 ++++
 5 files changed, 98 insertions(+), 258 deletions(-)
```

I wish there was a better way to do it.


shuujii:
> It also includes changes in mrbgems/mruby-compiler/core/y.tab.c, so the difference is inflated by more than six times.
> I wish there was a better way to do it.

For the time being, why not just ignore `#line` directive and `yyrline` in the difference in the following steps?

```console
$ touch .gitattributes
$ echo '/mrbgems/mruby-compiler/core/y.tab.c diff=tabc' >> .gitattributes
$ git config diff.tabc.textconv "ruby --disable-gems -0pe 'gsub(/^\\x23line [^\\n]+?\\n|^[^\\n]+?YYRLINEYYN.+?^}\\x3b\\n/m, %())'"
```

dearblue:
Thanks for the comment.
The difference displayed for the time being will be less, but hmm, it is different from what I thought...
Up until about a week ago, FreeBSD was a 3.5 series binary package of bison, so the difference was huge.
It's not always up-to-date bison available, so in that case I may need someone to help me out...

matz:
Merged with the `y.tab.c` generated by the latest bison.

dearblue:
Thanks @matz. And I am sorry for the lack of words.
In this PR, I was using bison-3.6.2 to convert, so it was possible to merge as is (if there are no other problems).

PR #5023, opened 2020-06-09
Remove `mrb_assert()` in `Module.nesting`

dearblue:
The following code was causing SIGSEGV:

```ruby
Module.method(:nesting).call
```


matz:
Oops, I pushed the wrong button. I tried to merge it.

PR #5028, opened 2020-06-26
Work around MSC optimization generating non functional code with Array#shuffle

RoryO:
Quite the interesting issue here.

Summary of changes
===

- Work around MSC and MSC++ compiler bug where it would optimize away a union access which may contain a static value in random.c

- Updated appveyor config, added MS 2019 compilers, and uses the highest Ruby version available on the appveyor image

- Removed special compiler flags in github actions MSC build

Description
===

Currently the tests with master fails with the Microsoft compiler v19 with optimizations enabled. This _only_ happens with MSC and from what I can determine only with MSC > 19. The simplest reproduction is the Ruby expression from the Random test

```ruby
[1, 2].shuffle Random.new
```

Passing a Random instance to an Array causes a segfault. I have some insight as to what happens though I do not know how to fix it besides disabling optimization entirely for the `mrb_ary_shuffle_bang` C function.

With optimizations enabled the MSC compiler infers some hard coded values incorrectly and stores them in registers. Code down the line treats the values in the registers as pointers which leads to a crash. I've tried expanding all the macros out, setting intermediate values as a part of the macro expansion, and marking the suspect variables as `volatile`. Using `declspec(noinline)` on all relevant functions it still does not fix it. Nothing I know how to do forces the compiler to stop over optimizing and generating hard numbers. However, I am no expert and I am well out of my depth at this point trying to debug optimized code.

More in depth, the MSC compiler hard sets register `rbx` to a literal `0x18` value when casting to `rand_state *`. This causes the pointer to go wild later in `random_rand` trying to follow a pointer to memory address `0x18` 

Annotation 2020-06-26 141756

Annotation 2020-06-26 143006

 It also sets`rsi` to `0x2` breaking C calling conventions and overwriting the pointer to the `mrb_state*` stored in`rsi`. 
Annotation 2020-06-26 131541

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,6 +101,3 @@ jobs:
         rake -E "STDOUT.sync=true" -m -j4 test
       env:
         MRUBY_CONFIG: appveyor_config.rb
-        # TODO(take-cheeze): Re-enable /O2
-        CFLAGS: "/c /nologo /W3 /we4013 /Zi /MD /D_CRT_SECURE_NO_WARNINGS"
-        CXXFLAGS: "/c /nologo /W3 /Zi /MD /EHs /D_CRT_SECURE_NO_WARNINGS"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,7 +41,7 @@ environment:
       machine: x86
 
 init:
-  - call "%visualcpp%" "%machine%"
+  - call "%visualcpp%" %machine%
   # For using Rubyins4aller's Ruby 2.6 64bit
   # 2.6 is the highest supported Ruby version across all historical
   # Visual Studio AppVeyor images. Ruby 2.7 is only on the 2019 image.
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,12 +32,12 @@ environment:
 
     - job_name: Visual Studio 2013 64bit
       visualcpp: C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat
-      appveyor_build_worker_image: Visual Studio 2013
+      appveyor_build_worker_image: Visual Studio 2015
       machine: x86_amd64
 
     - job_name: Visual Studio 2013 32bit
       visualcpp: C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\vcvarsall.bat
-      appveyor_build_worker_image: Visual Studio 2013
+      appveyor_build_worker_image: Visual Studio 2015
       machine: x86
 
 init:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,40 +4,44 @@ shallow_clone: true
 
 environment:
   matrix:
-    # Visual Studio 2019 64bit
-    - visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat
+    - job_name: Visual Studio 2019 64bit
+      visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat
       appveyor_build_worker_image: Visual Studio 2019
 
-    # Visual Studio 2019 32bit
-    - visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars32.bat
+    - job_name: Visual Studio 2019 32bit
+      visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars32.bat
       appveyor_build_worker_image: Visual Studio 2019
 
-    # Visual Studio 2017 64bit
-    - visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat
+    - job_name: Visual Studio 2017 64bit
+      visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat
       appveyor_build_worker_image: Visual Studio 2017
 
-    # Visual Studio 2017 32bit
-    - visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars32.bat
+    - job_name: Visual Studio 2017 32bit
+      visualcpp: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars32.bat
       appveyor_build_worker_image: Visual Studio 2017
 
-    # Visual Studio 2015 64bit
-    - visualcpp: C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvars64.bat
+    - job_name: Visual Studio 2015 64bit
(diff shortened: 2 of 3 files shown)


mmozeiko:
Some more investigations show that this is MSVC codegen bug introduced in version 19.23, and is in current 19.24.
Version 19.22 generates correct code.

Minimal example code that reproduces issue: https://gcc.godbolt.org/z/GBgNYT
Top-right is correct code with v19.22
Bottom-right is bad code with v19.23
It has following two instruction that causes crash:
```
        mov     edx, 51966                                ; 0000cafeH
        mov     eax, DWORD PTR [rdx]
```

It looks that newer MSVC versions do not like how mrb_value unions work in `mrb_ary_shuffle_bang` function - when changing nil type to Random type something goes wrong and pointer `p` in mrb_value is incorrectly used.

Alternative to fully disabling optimizations is to make mrb_value r volatile:
```
  volatile mrb_value r = mrb_nil_value();
```
Then accesses to it won't trigger this bug.

RoryO:
Thanks to @mmozeiko 's excellent compiler detective work I've narrowed the scope to adding `volatile` to `r`. The reproduction test case still passes. I never thought of marking `r` as `volatile` as it seemed completely unrelated.

RoryO:
Oh no, adding `volatile` to this actually increases the scope of impact. I'm sorry I didn't do a `deep_clean` before rebuilding and testing locally. `mrb_istruct_ptr(mrb_value)` needs a `volatile` qualifier and every caller to `mrb_istruct_ptr` also needs the `volatile` qualifier. Although there are only a few places where we call `mrb_istruct_ptr` it makes me uncomfortable scattering this fix.

```
src\class.c
831:          *p = mrb_istruct_ptr(ss);

mrbgems\mruby-rational\src\rational.c
17:#define rational_ptr(mrb, v) (struct mrb_rational*)mrb_istruct_ptr(v)

mrbgems\mruby-random\src\random.c
154:#define random_ptr(v) (rand_state*)mrb_istruct_ptr(v)

mrbgems\mruby-inline-struct\test\inline.c
9:  char *string = (char*)mrb_istruct_ptr(self);
35:  return mrb_str_new_cstr(mrb, (const char*)mrb_istruct_ptr(self));
53:  return mrb_bool_value(((char*)mrb_istruct_ptr(object))[0] == 's');
67:  char *ptr = (char*)mrb_istruct_ptr(self);

mrbgems\mruby-complex\src\complex.c
27:#define complex_ptr(mrb, v) (struct mrb_complex*)mrb_istruct_ptr(v)

include\mruby\istruct.h
38:MRB_INLINE void* mrb_istruct_ptr(mrb_value object)
```


@matz what would you like to do here? We can either

- Mark the function `mrb_shuffle_bang` as unoptimized for MSC >=1923 to whatever version fixes the compiler bug or
- Introduce a conditional `volatile` keyword a few places into the code with the same `#if defined` logic

mmozeiko:
Oh, I did not try with other compilers. Just the volatile there compiled for me on MSVC.
If you don't like that volatile. I think you can leave it without it, but do a roundtrip through volatile like this:
```
   volatile mrb_value rr;
   rr = r;
   r = rr;
```
Do this beore random_check, or maybe before mrb_nil_p. That should also prevent this bug. And it won't change type of r, so rest of functions should be fine.

RoryO:
Oh brilliant! Round tripping to a temporary variable does work. That limits the scope to just this function and eliminates the need for pervasive `volatile` qualifiers, and my previous question is now invalid. However, running `rake` locally for compilation works using the default configuration, while using the compile flags in `build.yml` returns an error `error C2678: binary '=': no operator found which takes a left-hand operand of type 'volatile mrb_value'`

I guess this goes back to my original question about CI. The CI compiler settings are different than what the end user gets when running `rake`. Should we just remove the custom compiler flags from CI?

I've been at this all day and out of steam. Let me know what you want to do and I'll do it.

RoryO:
Okay, I have finally finished and updated the PR description accordingly. The CIFuzz check fails that looks the fault of upstream.

shuujii:
I would like to have separate PR for handling optimization bugs and changing CI configurations. I have no objection to the former, but have some comments to the latter.

* It is intentional not to run tests in parallel due to the following reasons.
  * Prevent logs from mixing.
  * Rarely, tests may not be finished (it may depend on the environment). I'm wondering if Ruby crashed and it becomes like an infinite loop.
* I don't think similar environments (e.g. MSVC 2019 64-bit) need to be tested on multiple CI platforms.

RoryO:
> * It is intentional not to run tests in parallel due to the following reasons.
>   
>   * Prevent logs from mixing.
>   * Rarely, tests may not be finished (it may depend on the environment). I'm wondering if Ruby crashed and it becomes like an infinite loop.

Do you mean running rake with `-m`? I did not remove it, it was already there. The only CI changes I did was putting back the compiler optimization compiler flag from Github Actions, matching what the user gets when running `rake`; and adding the 2019 version to Appveyor. I suspect we removed the optimization flag from Github because we ran into compiler bugs at some point, and we did not have 2019 on Appveyor, so we were blind to this bug.

> * I don't think similar environments (e.g. MSVC 2019 64-bit) need to be tested on multiple CI platforms.

I agree. I've used the Windows Github actions platform and found it poor compared to Appveyor. I am in favor of removing the Windows Github action and make the Appveyor build the source for Windows.

I can create another PR which

- Removes the use of `rake -m` on CI
- Removes the Windows Github action

Do you agree?

shuujii:
> * Removes the use of `rake -m` on CI

I mean running tests (not running builds) with `rake -m`.

https://github.com/mruby/mruby/pull/5028/files#diff-180360612c6b8c4ed830919bbb4dd459L38-R48

Originally the builds were run in parallel (`rake -m`), but the tests were not run in parallel (`rake -E $stdout.sync=true test`).

> * Removes the Windows Github action

I don't have a strong opinion on removing it from either AppVeyor or GitHub Actions if the duplication is eliminated. However, AppVeyor can not execute jobs in parallel, so I am a little worried that the total CI time will increase steadily as the number of jobs increases.

RoryO:
> > * Removes the use of `rake -m` on CI
> 
> I mean running tests (not running builds) with `rake -m`.
> 
> https://github.com/mruby/mruby/pull/5028/files#diff-180360612c6b8c4ed830919bbb4dd459L38-R48

Oh I see! I was making the builds use exactly the same parameters, as shown in the GH action command

https://github.com/mruby/mruby/pull/5028/files#diff-898e737ee4e56ba042bda31764bed49eR101

Which is relevant to

> > * Removes the Windows Github action
> 
> I don't have a strong opinion on removing it from either AppVeyor or GitHub Actions if the duplication is eliminated. However, AppVeyor can not execute jobs in parallel, so I am a little worried that the total CI time will increase steadily as the number of jobs increases.

Well that's a different issue. We test all MSC compilers back to the first 2013 version when MS initially made the compilers free. It looks like we only use current LTS supported Ubuntu compilers; so the last 5 years. If we support only the current MSC compiler year and the previous release, 2019 and 2017, that cuts it down to four builds. So revising the proposal

- Remove `-m` for tests

- Remove GH Windows action, change Appveyor to use 2019 and 2017 compilers

Since we're on the subject, what's the Travis build for? Was it being replaced by Github Actions and wasn't removed yet?

matz:
We are not going to remove TravisCI soon. For some reason, Travis and GitHub Actions behave differently. Maybe in the future.

RoryO:
Thank you. @matz, do you have an opinion on how far back we should support Microsoft compilers?

PR #5030, opened 2020-07-06
Fix compiling mruby-io on msc with UTF8 enabled

RoryO:


mrbgems/mruby-io/src/io.c
```diff
@@ -18,6 +18,7 @@
 #if defined(_WIN32) || defined(_WIN64)
   #include <winsock.h>
   #include <io.h>
+  #include <BaseTsd.h>
   #define open  _open
```
matz, line 22: What is this `BaseTsd.h` for?

RoryO: `ssize_t` is not intrinsic to MSC like it is on other compilers which is why UTF8 enabled fails. It is available within the `BaseTsd.h` header, documented here https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types#ssize_t

> A signed version of SIZE_T.
> This type is declared in BaseTsd.h as follows:
> typedef LONG_PTR SSIZE_T;

matz: `ssize_t` is used in `mruby-io/src/file.c` and `mruby-socket/src/socket.c` too.  Are they OK?

dearblue: In MinGW, the file name is `basetsd.h` instead of `BaseTsd.h`.

  - mingw32
  
    ```console
    % git remote -v
    origin  https://scm.osdn.net/gitroot/mingw/mingw-org-wsl.git (fetch)
    origin  https://scm.osdn.net/gitroot/mingw/mingw-org-wsl.git (push)
    % git ls-files origin/5.4-trunk '**/BaseTsd.h' '**/basetsd.h'
    w32api/include/basetsd.h
    ```

  - mingw-w64

    ```console
    % git remote -v
    origin  https://git.code.sf.net/p/mingw-w64/mingw-w64 (fetch)
    origin  https://git.code.sf.net/p/mingw-w64/mingw-w64 (push)
    % git ls-files origin/master '**/BaseTsd.h' '**/basetsd.h'
    mingw-w64-headers/include/basetsd.h
    mingw-w64-tools/widl/include/basetsd.h
    ```

I actually got an error when compiling with MinGW32 GCC on FreeBSD.

```console
% mingw32-gcc -I include -I mrbgems/mruby-io/include -O3 -c mrbgems/mruby-io/src/io.c
mrbgems/mruby-io/src/io.c:21:23: fatal error: BaseTsd.h: No such file or directory
   #include <BaseTsd.h>
                       ^
compilation terminated.
```

This is solved by specifying it in lowercase, so I'm happy if it is set to `<basetsd.h>`.

RoryO: This is good information. I am also disturbed on how you can replicate it locally and CI does not fail.

RoryO: I downcased the include and it still works. This will cause problems if Microsoft ever implements case sensitive file system fully. The best solution is getting mingw to name their file correctly, however that probably isn't likely as it would break other projects.

@matz All files compile with the typedef set.

matz:
PR merged. Thank you!

dearblue:
Sorry, my confirmation wasn't enough.

The current MinGW seems to define `ssize_t` only if there is a special preprocessor definition in advance.
Therefore, I think this PR is generally okay.

However, Legacy MinGW on FreeBSD had its own `ssize_t`.
Also, maybe MinGW-w64 also has its own.

I about this will issue a new PR.

PR #5031, opened 2020-07-06
Update Windows CI settings

RoryO:
Following from discussion on #5028 

- Does not use `rake -m` for avoiding mixing output from rake tasks. Further, it appears building with `-m` introduces other issues while performing separate build and test tasks. The project builds in less than a minute so I do not think the parallelization helps.
- Removes MSC build from github actions
- Removes VS 2013 from Appveyor matrix as 2013 is no longer supported by MS (https://docs.microsoft.com/en-us/visualstudio/releases/2019/servicing)

@shuujii I think this addresses your comments?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     - name: build
       run: rake -m -j4 all
     - name: test
-      run: rake test
+      run: rake test -v
       env:
         MRUBY_CONFIG: travis_config.rb
 
@@ -25,7 +25,7 @@ jobs:
     - name: build
       run: rake -m -j4 all
     - name: test
-      run: rake test
+      run: rake test -v
       env:
         MRUBY_CONFIG: travis_config.rb
         CC: gcc
@@ -40,7 +40,7 @@ jobs:
     - name: build
       run: rake -m -j4 all
     - name: test
-      run: rake test
+      run: rake test -v
       env:
         MRUBY_CONFIG: travis_config.rb
         CC: clang
@@ -55,7 +55,7 @@ jobs:
     - name: build
       run: rake -m -j4 all
     - name: test
-      run: rake test
+      run: rake test -v
       env:
         MRUBY_CONFIG: travis_config.rb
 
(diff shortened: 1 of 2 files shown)


matz:
Did you give up `rake -m` (multi-tasking) to build? Any specific reason?

RoryO:
@shuujii has a good point on #5028 (https://github.com/mruby/mruby/pull/5028#issuecomment-651647168) where running tests with `-m` makes interpreting test output difficult as all the output is mixed. I tried separating the build command using `rake -m` and the test command which does not use `-m`. However, building with `-m` and then testing without `-m` leads to some strange segfaults on all nix system, seen in this commit https://github.com/mruby/mruby/pull/5031/commits/b3dace6ae3d28417bf7061a560671734a5568de5

I can put the test back as one line using `-m` as that is what it was before. One, it is strange how the build process with `-m` causes a segfault. Two, I think it is a good point that running test with multitask makes output hard to figure out

I did turn on `-v` with tests though so we can see exactly what tests failed.

shuujii:
> However, building with `-m` and then testing without `-m` leads to some strange segfaults on all nix system, seen in this commit [b3dace6](https://github.com/mruby/mruby/commit/b3dace6ae3d28417bf7061a560671734a5568de5)

It may be because `MRUBY_CONFIG` envvar is not set to the `build` action.

shuujii:
> It may be because `MRUBY_CONFIG` envvar is not set to the `build` action.

Sorry, it's not just `MRUBY_CONFIG`. It's also `CC`, `CXX`, `CFLAGS`, etc.

RoryO:
🤦 You're right, I read the yaml file wrong. I thought the `env` key applied to all of that step, not just the run step. That will fix it.

Apologies if there is notification spam as I push more commits testing this. Not possible working with CI otherwise.

RoryO:
Thanks @shuujii that was the issue. Everything works now.

matz:
@RoryO can you remove `-v` from tests? It's too verbose when I look into logs. I will use `-v` on my box.

matz:
Thank you!

PR #5032, opened 2020-07-10
Add ObjectSpace.memsize_of

RoryO:
Hello, this is my first significant contribution. I hope the review is not a burden.

This implements the mri method of `ObjectSpace.memsize_of`. With mruby we care a great deal about tracking resource use therefore I feel this is useful. All of the code is new, none copied. The mruby structures differ from C Ruby structures so existing C Ruby code isn't useful.

One difference between this implementation and the mri one is an addition of a `recurse` parameter. Specifying `ObjectSpace.memsize_of obj, recurse: true` will recursively descend into collection members and instance variables. Do let me know if I should not include this option and I will remove it.

Additionally, I added a protocol for `MRB_TT_DATA` objects. If a data object implements the `memsize` method this code will call that method and use it's return value for the size total. This allows a chance for including data objects into memory calculations, a gap I noticed from mri.

### DIFF
--- a/include/mruby/gc.h
+++ b/include/mruby/gc.h
@@ -21,6 +21,7 @@ struct mrb_state;
 #define MRB_EACH_OBJ_BREAK 1
 typedef int (mrb_each_object_callback)(struct mrb_state *mrb, struct RBasic *obj, void *data);
 void mrb_objspace_each_objects(struct mrb_state *mrb, mrb_each_object_callback *callback, void *data);
+const mrb_int mrb_objspace_page_slot_size();
 MRB_API void mrb_free_context(struct mrb_state *mrb, struct mrb_context *c);
 
 #ifndef MRB_GC_ARENA_SIZE
--- a/include/mruby/hash.h
+++ b/include/mruby/hash.h
@@ -23,6 +23,7 @@ struct RHash {
 #define mrb_hash_ptr(v)    ((struct RHash*)(mrb_ptr(v)))
 #define mrb_hash_value(p)  mrb_obj_value((void*)(p))
 
+mrb_int os_memsize_of_hash_table(mrb_value obj);
 MRB_API mrb_value mrb_hash_new_capa(mrb_state *mrb, mrb_int capa);
 MRB_API mrb_value mrb_ensure_hash_type(mrb_state *mrb, mrb_value hash);
 MRB_API mrb_value mrb_check_hash_type(mrb_state *mrb, mrb_value hash);
--- a/include/mruby/hash.h
+++ b/include/mruby/hash.h
@@ -23,7 +23,7 @@ struct RHash {
 #define mrb_hash_ptr(v)    ((struct RHash*)(mrb_ptr(v)))
 #define mrb_hash_value(p)  mrb_obj_value((void*)(p))
 
-mrb_int os_memsize_of_hash_table(mrb_value obj);
+mrb_int mrb_os_memsize_of_hash_table(mrb_value obj);
 MRB_API mrb_value mrb_hash_new_capa(mrb_state *mrb, mrb_int capa);
 MRB_API mrb_value mrb_ensure_hash_type(mrb_state *mrb, mrb_value hash);
 MRB_API mrb_value mrb_check_hash_type(mrb_state *mrb, mrb_value hash);
--- a/mrbgems/mruby-objectspace/src/mruby_objectspace.c
+++ b/mrbgems/mruby-objectspace/src/mruby_objectspace.c
@@ -270,7 +270,7 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_value recurse, mrb_int* 
     }
     case MRB_TT_HASH: {
       (*t) += mrb_objspace_page_slot_size() +
-              os_memsize_of_hash_table(obj);
+              mrb_os_memsize_of_hash_table(obj);
       if(!mrb_nil_p(recurse)) {
         os_memsize_of_object(mrb, mrb_hash_keys(mrb, obj), recurse, t);
         os_memsize_of_object(mrb, mrb_hash_values(mrb, obj), recurse, t);
--- a/mrbgems/mruby-objectspace/src/mruby_objectspace.c
+++ b/mrbgems/mruby-objectspace/src/mruby_objectspace.c
@@ -262,8 +262,12 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int* t
       break;
     }
     case MRB_TT_HASH: {
-      /*struct htable* htable = RHASH_TBL(obj);
-       * Need htable & segment struct defs */
+      (*t) += mrb_objspace_page_slot_size() +
+              os_memsize_of_hash_table(obj);
+      if(recurse) {
+        os_memsize_of_object(mrb, mrb_hash_keys(mrb, obj), recurse, t);
+        os_memsize_of_object(mrb, mrb_hash_values(mrb, obj), recurse, t);
+      }
       break;
     }
     case MRB_TT_ARRAY: {
@@ -325,7 +329,6 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int* t
     case MRB_TT_FREE:
     case MRB_TT_UNDEF:
     case MRB_TT_ENV:
-    case MRB_TT_ISTRUCT:
     /* never used, silences compiler warning
      * not having a default: clause lets the compiler tell us when there is a new
      * TT not accounted for */
--- a/mrbgems/mruby-objectspace/src/mruby_objectspace.c
+++ b/mrbgems/mruby-objectspace/src/mruby_objectspace.c
@@ -324,8 +324,10 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int* t
         ci_p++;
       }
 
-      for(i = 0; i <= f->cxt->esize; i++) {
-         os_memsize_of_irep(mrb, f->cxt->ensure[i]->body.irep, t);
+      if(f->cxt->esize) {
+        for(i = 0; i <= f->cxt->esize; i++) {
+          os_memsize_of_irep(mrb, f->cxt->ensure[i]->body.irep, t);
+        }
       }
 
       (*t) += mrb_objspace_page_slot_size() +
--- a/mrbgems/mruby-objectspace/src/mruby_objectspace.c
+++ b/mrbgems/mruby-objectspace/src/mruby_objectspace.c
@@ -196,7 +196,7 @@ os_memsize_of_ivars(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int *t)
   if(recurse) {
     mrb_int r = (mrb_int)recurse;
     mrb_int *cb_data[2] = { &r, t };
-    mrb_iv_foreach(mrb, obj, os_memsize_ivar_cb, t);
+    mrb_iv_foreach(mrb, obj, os_memsize_ivar_cb, cb_data);
   }
 }
 
@@ -261,8 +261,8 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int* t
       break;
     }
     case MRB_TT_HASH: {
-      struct htable* htable = RHASH_TBL(obj);
-      /* Need htable & segment struct defs */
+      /*struct htable* htable = RHASH_TBL(obj);
+       * Need htable & segment struct defs */
       break;
     }
     case MRB_TT_ARRAY: {
@@ -302,10 +302,11 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int* t
       (*t) += sizeof(struct mrb_range_edges);
     #endif
       break;
-    case MRB_TT_FIBER:
-      struct RFiber* fiber = (struct RFiber*)mrb_ptr(obj);
+    case MRB_TT_FIBER: {
+      /*  struct RFiber* fiber = (struct RFiber*)mrb_ptr(obj); */
       (*t) += sizeof(struct mrb_context);
       break;
+    }
     /*  zero heap size types.
      *  immediate VM stack values, contained within mrb_state, mrb_heap_page,
      *  or on C stack */
--- a/mrbgems/mruby-objectspace/src/mruby_objectspace.c
+++ b/mrbgems/mruby-objectspace/src/mruby_objectspace.c
@@ -2,6 +2,14 @@
 #include <mruby/gc.h>
 #include <mruby/hash.h>
 #include <mruby/class.h>
+#include <mruby/object.h>
+#include <mruby/numeric.h>
+#include <mruby/string.h>
+#include <mruby/array.h>
+#include <mruby/variable.h>
+#include <mruby/proc.h>
+#include <mruby/value.h>
+#include <mruby/range.h>
 
 struct os_count_struct {
   mrb_int total;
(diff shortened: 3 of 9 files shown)


mrbgems/mruby-objectspace/src/mruby_objectspace.c
```diff
@@ -168,12 +176,206 @@ os_each_object(mrb_state *mrb, mrb_value self)
...
+  mrb_int* total = &cb_data[1];
+
+  os_memsize_of_object(mrb, obj, (mrb_bool)(recurse), total);
+  return 0;
+}
+
+static void
+os_memsize_of_ivars(mrb_state* mrb, mrb_value obj, mrb_bool recurse, mrb_int *t)
+{
+  /* need iv segment table size */
```
RoryO, line 195: Currently hash tables and their data segments are within `.c` files not accessible outside that translation unit. Is it okay to pull the definitions of iv and hash tables into header files? Otherwise I have to copy the definition of these structs into this file so we can calculate them.

Something like this

```patch
diff --git a/include/mruby/hash.h b/include/mruby/hash.h
index 0052a110..18dadab5 100644
--- a/include/mruby/hash.h
+++ b/include/mruby/hash.h
@@ -20,6 +20,33 @@ struct RHash {
   struct htable *ht;
 };
 
+/* hash table structure */
+
+struct h_segkv {
+  mrb_value key;
+  mrb_value val;
+};
+
+typedef struct h_segindex {
+  size_t size;
+  size_t capa;
+  struct h_segkv *table[];
+} h_segindex;
+
+typedef struct h_segment {
+  uint16_t size;
+  struct h_segment *next;
+  struct h_segkv e[];
+} h_segment;
+
+typedef struct htable {
+  h_segment *rootseg;
+  h_segment *lastseg;
+  mrb_int size;
+  uint16_t last_len;
+  h_segindex *index;
+} htable;
+
 #define mrb_hash_ptr(v)    ((struct RHash*)(mrb_ptr(v)))
 #define mrb_hash_value(p)  mrb_obj_value((void*)(p))
 
diff --git a/include/mruby/variable.h b/include/mruby/variable.h
index 6e918cf5..d671d4d9 100644
--- a/include/mruby/variable.h
+++ b/include/mruby/variable.h
@@ -29,6 +29,23 @@ struct global_entry {
   mrb_sym id;
 };
 
+#ifndef MRB_IV_SEGMENT_SIZE
+#define MRB_IV_SEGMENT_SIZE 4
+#endif
+
+typedef struct iv_segment {
+  mrb_sym key[MRB_IV_SEGMENT_SIZE];
+  mrb_value val[MRB_IV_SEGMENT_SIZE];
+  struct iv_segment *next;
+} iv_segment;
+
+/* Instance variable table structure */
+typedef struct iv_tbl {
+  iv_segment *rootseg;
+  size_t size;
+  size_t last_len;
+} iv_tbl;
+
 mrb_value mrb_vm_special_get(mrb_state*, mrb_sym);
 void mrb_vm_special_set(mrb_state*, mrb_sym, mrb_value);
 mrb_value mrb_vm_cv_get(mrb_state*, mrb_sym);
diff --git a/src/hash.c b/src/hash.c
index 4d531090..99dc66bf 100644
--- a/src/hash.c
+++ b/src/hash.c
@@ -21,39 +21,13 @@ mrb_int mrb_float_id(mrb_float f);
 #endif
 #define HT_SEG_INCREASE_RATIO 6 / 5
 
-struct segkv {
-  mrb_value key;
-  mrb_value val;
-};
-
-typedef struct segment {
-  uint16_t size;
-  struct segment *next;
-  struct segkv e[];
-} segment;
-
-typedef struct segindex {
-  size_t size;
-  size_t capa;
-  struct segkv *table[];
-} segindex;
-
-/* hash table structure */
-typedef struct htable {
-  segment *rootseg;
-  segment *lastseg;
-  mrb_int size;
-  uint16_t last_len;
-  segindex *index;
-} htable;
-
 static /* inline */ size_t
 ht_hash_func(mrb_state *mrb, htable *t, mrb_value key)
 {
   enum mrb_vtype tt = mrb_type(key);
   mrb_value hv;
   size_t h;
-  segindex *index = t->index;
+  h_segindex *index = t->index;
   size_t capa = index ? index->capa : 0;
 
   switch (tt) {
@@ -121,7 +95,7 @@ ht_hash_equal(mrb_state *mrb, htable *t, mrb_value a, mrb_value b)
 
   default:
     {
-      segindex *index = t->index;
+      h_segindex *index = t->index;
       size_t capa = index ? index->capa : 0;
       mrb_bool eql = mrb_eql(mrb, a, b);
       if (index && (index != t->index || capa != index->capa)) {
@@ -170,8 +144,8 @@ ht_index(mrb_state *mrb, htable *t)
 {
   size_t size = (size_t)t->size;
   size_t mask;
-  segindex *index = t->index;
-  segment *seg;
+  h_segindex *index = t->index;
+  h_segment *seg;
   size_t i;
 
   /* allocate index table */
@@ -180,7 +154,7 @@ ht_index(mrb_state *mrb, htable *t)
   }
   power2(size);
   if (!index || index->capa < size) {
-    index = (segindex*)mrb_realloc_simple(mrb, index, sizeof(segindex)+sizeof(struct segkv*)*size);
+    index = (h_segindex*)mrb_realloc_simple(mrb, index, sizeof(h_segindex)+sizeof(struct h_segkv*)*size);
     if (index == NULL) {
       mrb_free(mrb, t->index);
       t->index = NULL;
@@ -221,9 +195,9 @@ ht_index(mrb_state *mrb, htable *t)
 static void
 ht_compact(mrb_state *mrb, htable *t)
 {
-  segment *seg;
+  h_segment *seg;
   uint16_t i, i2;
-  segment *seg2 = NULL;
+  h_segment *seg2 = NULL;
   mrb_int size = 0;
 
   if (t == NULL) return;
@@ -277,8 +251,8 @@ ht_compact(mrb_state *mrb, htable *t)
   }
 }
 
-static segment*
-segment_alloc(mrb_state *mrb, segment *seg)
+static h_segment*
+h_segment_alloc(mrb_state *mrb, h_segment *seg)
 {
   uint32_t size;
 
@@ -288,7 +262,7 @@ segment_alloc(mrb_state *mrb, segment *seg)
     if (size > UINT16_MAX) size = UINT16_MAX;
   }
 
-  seg = (segment*)mrb_malloc(mrb, sizeof(segment)+sizeof(struct segkv)*size);
+  seg = (h_segment*)mrb_malloc(mrb, sizeof(h_segment)+sizeof(struct h_segkv)*size);
   seg->size = size;
   seg->next = NULL;
 
@@ -299,9 +273,9 @@ segment_alloc(mrb_state *mrb, segment *seg)
 static void
 ht_index_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
 {
-  segindex *index = t->index;
+  h_segindex *index = t->index;
   size_t k, sp, step = 0, mask;
-  segment *seg;
+  h_segment *seg;
 
   if (index->size >= UPPER_BOUND(index->capa)) {
     /* need to expand table */
@@ -331,8 +305,8 @@ ht_index_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
   if (t->last_len < seg->size) {
     index->table[k] = &seg->e[t->last_len++];
   }
-  else {                        /* append a new segment */
-    seg->next = segment_alloc(mrb, seg);
+  else {                        /* append a new h_segment */
+    seg->next = h_segment_alloc(mrb, seg);
     seg = seg->next;
     seg->next = NULL;
     t->lastseg = seg;
@@ -349,7 +323,7 @@ ht_index_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
 static void
 ht_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
 {
-  segment *seg;
+  h_segment *seg;
   mrb_int i, deleted = 0;
 
   if (t == NULL) return;
@@ -361,7 +335,7 @@ ht_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
   while (seg) {
     for (i=0; i<seg->size; i++) {
       mrb_value k = seg->e[i].key;
-      /* Found room in last segment after last_len */
+      /* Found room in last h_segment after last_len */
       if (!seg->next && i >= t->last_len) {
         seg->e[i].key = key;
         seg->e[i].val = val;
@@ -382,7 +356,7 @@ ht_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
   }
   /* Not found */
 
-  /* Compact if last segment has room */
+  /* Compact if last h_segment has room */
   if (deleted > 0 && deleted > MRB_HT_INIT_SIZE) {
     ht_compact(mrb, t);
   }
@@ -394,8 +368,8 @@ ht_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
     i = t->last_len;
   }
   else {
-    /* append new segment */
-    seg = segment_alloc(mrb, t->lastseg);
+    /* append new h_segment */
+    seg = h_segment_alloc(mrb, t->lastseg);
     i = 0;
     if (t->rootseg == NULL) {
       t->rootseg = seg;
@@ -417,7 +391,7 @@ ht_put(mrb_state *mrb, htable *t, mrb_value key, mrb_value val)
 static mrb_bool
 ht_index_get(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 {
-  segindex *index = t->index;
+  h_segindex *index = t->index;
   size_t mask = HT_MASK(index);
   size_t k = ht_hash_func(mrb, t, key) & mask;
   size_t step = 0;
@@ -437,7 +411,7 @@ ht_index_get(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 static mrb_bool
 ht_get(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 {
-  segment *seg;
+  h_segment *seg;
   mrb_int i;
 
   if (t == NULL) return FALSE;
@@ -469,7 +443,7 @@ ht_get(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 static mrb_bool
 ht_del(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 {
-  segment *seg;
+  h_segment *seg;
   mrb_int i;
 
   if (t == NULL) return FALSE;
@@ -499,14 +473,14 @@ ht_del(mrb_state *mrb, htable *t, mrb_value key, mrb_value *vp)
 static void
 ht_foreach(mrb_state *mrb, htable *t, mrb_hash_foreach_func *func, void *p)
 {
-  segment *seg;
+  h_segment *seg;
   mrb_int i;
 
   if (t == NULL) return;
   seg = t->rootseg;
   while (seg) {
     for (i=0; i<seg->size; i++) {
-      /* no value in last segment after last_len */
+      /* no value in last h_segment after last_len */
       if (!seg->next && i >= t->last_len) {
         return;
       }
@@ -529,7 +503,7 @@ mrb_hash_foreach(mrb_state *mrb, struct RHash *hash, mrb_hash_foreach_func *func
 static htable*
 ht_copy(mrb_state *mrb, htable *t)
 {
-  segment *seg;
+  h_segment *seg;
   htable *t2;
   mrb_int i;
 
@@ -557,12 +531,12 @@ ht_copy(mrb_state *mrb, htable *t)
 static void
 ht_free(mrb_state *mrb, htable *t)
 {
-  segment *seg;
+  h_segment *seg;
 
   if (!t) return;
   seg = t->rootseg;
   while (seg) {
-    segment *p = seg;
+    h_segment *p = seg;
     seg = seg->next;
     mrb_free(mrb, p);
   }
@@ -1009,7 +983,7 @@ mrb_hash_delete(mrb_state *mrb, mrb_value self)
 static void
 ht_shift(mrb_state *mrb, htable *t, mrb_value *kp, mrb_value *vp)
 {
-  segment *seg = t->rootseg;
+  h_segment *seg = t->rootseg;
   mrb_int i;
 
   while (seg) {
diff --git a/src/variable.c b/src/variable.c
index 030aa7b0..197a83a3 100644
--- a/src/variable.c
+++ b/src/variable.c
@@ -11,23 +11,6 @@
 #include <mruby/string.h>
 #include <mruby/variable.h>
 
-#ifndef MRB_IV_SEGMENT_SIZE
-#define MRB_IV_SEGMENT_SIZE 4
-#endif
-
-typedef struct segment {
-  mrb_sym key[MRB_IV_SEGMENT_SIZE];
-  mrb_value val[MRB_IV_SEGMENT_SIZE];
-  struct segment *next;
-} segment;
-
-/* Instance variable table structure */
-typedef struct iv_tbl {
-  segment *rootseg;
-  size_t size;
-  size_t last_len;
-} iv_tbl;
-
 /* Creates the instance variable table. */
 static iv_tbl*
 iv_new(mrb_state *mrb)
@@ -46,9 +29,9 @@ iv_new(mrb_state *mrb)
 static void
 iv_put(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value val)
 {
-  segment *seg;
-  segment *prev = NULL;
-  segment *matched_seg = NULL;
+  iv_segment *seg;
+  iv_segment *prev = NULL;
+  iv_segment *matched_seg = NULL;
   size_t matched_idx = 0;
   size_t i;
 
@@ -86,7 +69,7 @@ iv_put(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value val)
     return;
   }
 
-  seg = (segment*)mrb_malloc(mrb, sizeof(segment));
+  seg = (iv_segment*)mrb_malloc(mrb, sizeof(iv_segment));
   seg->next = NULL;
   seg->key[0] = sym;
   seg->val[0] = val;
@@ -104,7 +87,7 @@ iv_put(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value val)
 static mrb_bool
 iv_get(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value *vp)
 {
-  segment *seg;
+  iv_segment *seg;
   size_t i;
 
   if (t == NULL) return FALSE;
@@ -130,7 +113,7 @@ iv_get(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value *vp)
 static mrb_bool
 iv_del(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value *vp)
 {
-  segment *seg;
+  iv_segment *seg;
   size_t i;
 
   if (t == NULL) return FALSE;
@@ -158,7 +141,7 @@ iv_del(mrb_state *mrb, iv_tbl *t, mrb_sym sym, mrb_value *vp)
 static void
 iv_foreach(mrb_state *mrb, iv_tbl *t, mrb_iv_foreach_func *func, void *p)
 {
-  segment *seg;
+  iv_segment *seg;
   size_t i;
 
   if (t == NULL) return;
@@ -186,7 +169,7 @@ iv_foreach(mrb_state *mrb, iv_tbl *t, mrb_iv_foreach_func *func, void *p)
 static size_t
 iv_size(mrb_state *mrb, iv_tbl *t)
 {
-  segment *seg;
+  iv_segment *seg;
   size_t size = 0;
 
   if (t == NULL) return 0;
@@ -208,7 +191,7 @@ iv_size(mrb_state *mrb, iv_tbl *t)
 static iv_tbl*
 iv_copy(mrb_state *mrb, iv_tbl *t)
 {
-  segment *seg;
+  iv_segment *seg;
   iv_tbl *t2;
 
   size_t i;
@@ -235,11 +218,11 @@ iv_copy(mrb_state *mrb, iv_tbl *t)
 static void
 iv_free(mrb_state *mrb, iv_tbl *t)
 {
-  segment *seg;
+  iv_segment *seg;
 
   seg = t->rootseg;
   while (seg) {
-    segment *p = seg;
+    iv_segment *p = seg;
     seg = seg->next;
     mrb_free(mrb, p);
   }
```

mrbgems/mruby-objectspace/src/mruby_objectspace.c
```diff
@@ -168,12 +176,206 @@ os_each_object(mrb_state *mrb, mrb_value self)
...
+  (*t) += sizeof(struct RProc);
+  if(!MRB_PROC_CFUNC_P(proc)) os_memsize_of_irep(mrb, proc->body.irep, t);
+}
+
+static void
+os_memsize_of_methods(mrb_state* mrb, mrb_value obj, mrb_int* t)
+{
+  mrb_value method_list;
+  mrb_int i;
+  if(!mrb_respond_to(mrb, obj, mrb_intern_lit(mrb, "instance_methods"))) return;
```
RoryO, line 232: Similar situation to the hash tables, the functions and definitions for accessing these are within .c translation units. However, we can achieve the same thing by calling the Ruby functions so it's not so necessary. This is why we need `mruby-metaprog` and `mruby-method` as a test dependency.

mrbgems/mruby-objectspace/src/mruby_objectspace.c
```diff
@@ -168,12 +176,206 @@ os_each_object(mrb_state *mrb, mrb_value self)
...
+      break;
+    #endif
+    case MRB_TT_RANGE:
+    #ifndef MRB_RANGE_EMBED
+      (*t) += sizeof(struct mrb_range_edges);
+    #endif
+      break;
+    case MRB_TT_FIBER: {
+      /*  struct RFiber* fiber = (struct RFiber*)mrb_ptr(obj); */
+      (*t) += sizeof(struct mrb_context);
```
RoryO, line 307: I'm not completely sure on how to calculate a Fiber. I think it's the size of the context and the size of the program irep, but I think I'm missing more. Any advice?

mrbgems/mruby-objectspace/test/objectspace.rb
```diff
@@ -58,3 +59,78 @@
...
+  assert_equal int_size, 0, 'int zero'
+
+  sym_size = ObjectSpace.memsize_of :foo
+  assert_equal sym_size, 0, 'sym zero'
+
+  assert_equal ObjectSpace.memsize_of(true), int_size
+  assert_equal ObjectSpace.memsize_of(false), int_size
+
+  float_size = if Object.const_defined? :Float
+                 ObjectSpace.memsize_of 1.0
```
RoryO, line 75: Here is an example of where it is useful exposing internal details. The memory size of a float is 0 with default no boxing, but is the size of an `mrb_value` when using word boxing.

mrbgems/mruby-objectspace/test/objectspace.rb
```diff
@@ -58,3 +59,78 @@
...
+      a = 0
+      a + 1
+    end
+  end
+
+  m_size = ObjectSpace.memsize_of class_with_methods.instance_method(:foo)
+  assert_not_equal m_size, 0, 'method size not zero'
+
+  # collections
+  assert_equal ObjectSpace.memsize_of([]), 0, 'empty array size zero'
```
RoryO, line 128: It's also useful exposing Array and String embed length size.

mrbgems/mruby-objectspace/src/mruby_objectspace.c
matz: I don't think exposing internal data structure is a wise idea. How about providing a function to get the size of instance variable table e.g. `mrb_obj_iv_memsize()` in `variable.c`?

mrbgems/mruby-objectspace/src/mruby_objectspace.c
matz: You need to count the size of the stack (`c->stend - c->stbase`) as well.

mrbgems/mruby-objectspace/test/objectspace.rb
matz: It's weird that the embedded array returns 0, since it occupies one slot (48 bytes in 64bit).

RoryO: That is my first question. Should we always return the size of one slot for an object that does not perform any more allocations?

matz: I think so. At least CRuby `memsize_of` seems to work so, including the size of `RVALUE` slot, and excluding the size of `VALUE` that refers to the slot. That's why immediate values like `nil` return 0.

RoryO: I will do so then. Thank you for your advice!

matz: I am looking forward to the update.

include/mruby/gc.h
```diff
@@ -21,6 +21,7 @@ struct mrb_state;
 #define MRB_EACH_OBJ_BREAK 1
 typedef int (mrb_each_object_callback)(struct mrb_state *mrb, struct RBasic *obj, void *data);
 void mrb_objspace_each_objects(struct mrb_state *mrb, mrb_each_object_callback *callback, void *data);
+const mrb_int mrb_objspace_page_slot_size();
```
RoryO, line 24: This feels like it should be just a #DEFINE, however when using a `const int` return value compilers optimize this function call to a single `MOV` instruction instead, which is okay

https://godbolt.org/z/1b6Prq

mrbgems/mruby-objectspace/src/mruby_objectspace.c
RoryO: I _think_ I have this correct now.

include/mruby/gc.h
dearblue: Since the translation unit is divided, it is usually a function call (I don't know `gcc -flto` :smiley_cat:).
However, I don't think there is any problem with function calls.
If you care about that, you should be fine with global variables (with const).

Also, if you qualify a non-pointer return type with `const`, the compiler will ignore it (`gcc -Wignored-qualifiers`).

include/mruby/hash.h
```diff
@@ -23,6 +23,7 @@ struct RHash {
 #define mrb_hash_ptr(v)    ((struct RHash*)(mrb_ptr(v)))
 #define mrb_hash_value(p)  mrb_obj_value((void*)(p))
 
+mrb_int os_memsize_of_hash_table(mrb_value obj);
```
dearblue, line 26: The `mrb_` prefix is required.

src/variable.c
```diff
@@ -1128,6 +1129,14 @@ mrb_class_find_path(mrb_state *mrb, struct RClass *c)
   return path;
 }
 
+mrb_int
+mrb_obj_iv_tbl_memsize(mrb_state* mrb, mrb_value obj)
+{
+  return sizeof(iv_tbl) +
+    (sizeof(segment) * ceil(iv_size(mrb, mrb_obj_ptr(obj)->iv)/
```
dearblue, line 1136: Because of `MRB_WITHOUT_FLOAT`, it needs to replace the `<math.h>` header file and the `ceil()` function.

include/mruby/gc.h
RoryO: You're right I keep forgetting about translation unit boundaries.

It feels silly for a function call which always returns the same value. What I mean is because of the way the code is arranged in the c and header file, and because matz prefers not exposing internal structures in the header, it's not possible to create a global const. It's not a big deal either way.

src/variable.c
RoryO: Thank you. I forget that we sometimes do not have floats.

matz:
I missed that `ObjectSpace.memsize_of` does not recurse. It was originally designed to analyze single object allocation.
Do we really need recurring enhancement for mruby?

RoryO:
If you wish I can remove it.

matz:
I will do some fixes and refactoring to improve compatibility with CRuby. I am working on it.
Thank you!

RoryO:
Please show me if there's anything I can do to help, including working on C Ruby as well. I do not wish to burden!

matz:
Never mind. It's fun!

shuujii:
I've noticed that `ObjectSpace.memsize_of` doesn't exist in JRuby and so on. Do we need to maintain incompatible features among Ruby implementations in the mruby repository?

`ObjectSpace.memsize_of` may not return an exact value, so I personally find it difficult to use. Therefore, I feel it's better to provide it as an external gem.

matz:
Separating to a gem. Hmm. OK.

RoryO:
> ObjectSpace.memsize_of may not return an exact value

Can you provide some examples? I want to make it accurate along with matz' corrections. I think this is an important tool for an embedded project.

matz:
I think he meant the result does not count alignment gaps nor allocation overhead, which are not easy to calculate.

matz:
I made a separated gem for `memsize_of`. The diff is here:
https://gist.github.com/matz/0ece6c6fbeeb7609dd44e894f276ee7e
If I commit this change, we will lose @RoryO's name and history from the repository.
Maybe he can commit the above change as a PR.

shuujii:
> > ObjectSpace.memsize_of may not return an exact value
>
> Can you provide some examples? I wanted to make it accurate along with matz' corrections. I think this is an important tool for an embedded project.

Here are some examples. I'm not sure what value users typically expect or appropriate for the return value of `ObjectSpace.memsize_of`.

* Currently, the method returns `0` for `Symbol`. However, creating one non-inline symbol uses one slot in `mrb_state::symtbl` (which may also uses a slot in `mrb_state::symhash`). Also, if the symbol name is not `ro_data`, allocate a buffer for it.
* Currently, the method returns `RVALUE` size for shared `String`. In the case of shared `Strings`, I think shared buffer size (including size of `mrb_shared_string`) should be accounted for in one of the objects that share the same buffer. However, I don't think it's obvious which objects should be accounted for. Information may need to be added to `RString` to identify the object, even if it is clear. I don't think it's desirable to increase runtime costs and so on just for the feature.
* Currently, the method retuerns `0` for `Fixnum`. However, it may be using a slot in literal pool.

> I made a separated gem for `memsize_of`.

My intention was to provide it in a different repository than `mruby/mruby` (e.g. `RoryO/mruby-os-memsize`).

I personally feel that return value definition of the method is unclear. The meaning of the value may change due to changes in internal implementation because it strongly depends on the internal implementation.

Still, the methods may have some uses depending on the use case, but I'm a little wondering if it is necessary to officially provide such a seemingly unclear feature (also, I'm a little worried that the maintenance cost seems to be high against the frequency of use).

I agree with separating gems when maintaining in `mruby/mruby` because the method is required `require 'objspace'` in CRuby.

matz:
Considering the fact `memsize_of` is that highly depends on the mruby internal, it should be maintained mainly by the core developer (i.e. me).

src/hash.c
```diff
@@ -518,6 +518,20 @@ ht_foreach(mrb_state *mrb, htable *t, mrb_hash_foreach_func *func, void *p)
...
+  mrb_int segkv_size = 0;
+
+  if(h->index) segkv_size = (sizeof(struct segkv) * h->index->capa);
+
+  return sizeof(htable) +
+    sizeof(segindex) +
+    (sizeof(segment) * h->size) +
+    segkv_size;
+}
+
```
shuujii, line 534: * The result seems to be incorrect.

Example (64-bit no-boxing):

| Hash Size  | Expected Result | Actual Result |
|-----------:|----------------:| -------------:|
|          1 |             184 |            72 |
|          8 |             360 |           184 |
|         16 |             600 |           312 |
|         24 |            1176 |          1464 |
|         32 |            1800 |          2616 |

* IV table size should also be included.

* The result may not fit into `mrb_int`.

RoryO: Thank you for this example! Do you mean that a Hash object should also include `mrb_obj_iv_tbl_memsize`, like this

```diff
--- a/mrbgems/mruby-os-memsize/src/memsize.c
+++ b/mrbgems/mruby-os-memsize/src/memsize.c
@@ -74,6 +74,7 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj)
     }
     case MRB_TT_HASH: {
       size += mrb_objspace_page_slot_size() +
+              mrb_obj_iv_tbl_memsize(mrb, obj) +
               mrb_os_memsize_of_hash_table(obj);
       break;
     }
```

Do you mean the numeric limit of `mrb_int`? That has a minimum limit of `INT32_MAX`. I suppose it's possible for one object consuming 2gb of memory.

shuujii: > Do you mean that a Hash object should also include `mrb_obj_iv_tbl_memsize`, like this

Since IV table is part of a `Hash` object, I think it's natural that its memory size be included in the result of `mrb_os_memsize_of_hash_table()`.

However, in CRuby, for example, the result of `rb_ary_memsize()` seems not to include memory size of IV table, so I think that it will look like the code you presented if follow CRuby.

> Do you mean the numeric limit of mrb_int?

Yes. I have no opinion on the behavior when it exceeds `MRB_INT_MAX`, but I'd like to know if it is not taken into consideration or if it is considered and fixed.

PR #5034, opened 2020-07-12
Fix `ssize_t` for mingw; ref #5030

dearblue:
Legacy MinGW and MinGW-w64 had own `ssize_t`.

### DIFF
--- a/mrbgems/mruby-io/src/io.c
+++ b/mrbgems/mruby-io/src/io.c
@@ -33,7 +33,10 @@
   typedef long fsuseconds_t;
   typedef int fmode_t;
   typedef int mrb_io_read_write_size;
+  #if !defined(_SSIZE_T_) && !defined(_SSIZE_T_DEFINED) && \
+      !defined(__have_typedef_ssize_t)
   typedef SSIZE_T ssize_t;
+  #endif
 
   #ifndef O_TMPFILE
     #define O_TMPFILE O_TEMPORARY


mrbgems/mruby-io/src/io.c
```diff
@@ -33,7 +33,10 @@
   typedef long fsuseconds_t;
   typedef int fmode_t;
   typedef int mrb_io_read_write_size;
+  #if !defined(_SSIZE_T_) && !defined(_SSIZE_T_DEFINED) && \
+      !defined(__have_typedef_ssize_t)
   typedef SSIZE_T ssize_t;
```
matz, line 38: Just to confirm. Here we define `ssize_t` using `SSIZE_T` but the condition above indicates `SSIZE_T` is **not** defined. Is it really OK?

dearblue: `_SSIZE_T_` (unlike `SSIZE_T`) is defined in Legacy MinGW when `ssize_t` is `typedef`.

I checked the build with `MRB_UTF8_STRING` in CI for MSVC and Modern MinGW, and it was ok.
- https://github.com/dearblue/mruby/runs/862624893
- https://github.com/dearblue/mruby/runs/862624882
- https://ci.appveyor.com/project/dearblue/mruby/builds/34048239

The changes about MinGW-w64 were only confirmed from the repository of the same project, so we have not confirmed by build.

RoryO: Throughout all this it’s been bothering me that we don’t test utf8 string enabled pathway on ci. To do that we’d have to double the amount of tests. I think we should do that because utf8 strings cause code to go down very different paths.

dearblue: I have also thought about it before.
However, there are `MRB_WITHOUT_FLOAT`, `MRB_WORD_BOXING` and more, so combining them increases the load on the CI.
It's difficult to balance.

PR #5045, opened 2020-07-21
Improve `mruby-os-memsize`

dearblue:


mrbgems/mruby-os-memsize/src/memsize.c
```diff
@@ -54,6 +54,7 @@ os_memsize_of_object(mrb_state* mrb, mrb_value obj)
       size += mrb_objspace_page_slot_size();
       if (!RSTR_EMBED_P(RSTRING(obj)) && !RSTR_SHARED_P(RSTRING(obj))) {
         size += RSTRING_CAPA(obj);
+        size ++; /* NUL terminator */
```
matz, line 73: I prefer not having space before `++` operator.

mrbgems/mruby-os-memsize/src/memsize.c
```diff
@@ -44,6 +44,22 @@ os_memsize_of_method(mrb_state* mrb, mrb_value method_obj)
   return size;
 }
 
+static mrb_bool
+obj_is_kind_of_checked(mrb_state *mrb, mrb_value obj, mrb_sym classname)
+{
+  mrb_value objclass = mrb_obj_value(mrb->object_class);
+
+  if (mrb_const_defined(mrb, objclass, classname)) {
+    mrb_value klass = mrb_const_get(mrb, objclass, classname);
```
matz, line 53: Probably you can use `mrb_class_get` here.

mrbgems/mruby-os-memsize/test/memsize.rb
```diff
@@ -26,14 +26,20 @@
     @a = 1
     @b = 2
   end
-  class_total_size = empty_class_def_size + (int_size * 2)
+  ivnum = 2
+  segnum = ((ivnum + Mrbtest::COUNTOF_IV_SEGMENT - 1) / Mrbtest::COUNTOF_IV_SEGMENT).floor
+  class_total_size = empty_class_def_size + (int_size * ivnum) +
+                     Mrbtest::SIZEOF_IV_SEGMENT * segnum
```
matz, line 32: I know this is kind of controversial, but I don't think we need these precise tests for `memsize_of`.
The fundamental behavior of `memsize_of` is "you can get the size of a various object in integer".

mrbgems/mruby-test/driver.c
```diff
@@ -232,6 +235,9 @@ mrb_init_test_driver(mrb_state *mrb, mrb_bool verbose)
 #endif
 #endif
 
+  mrb_define_const(mrb, mrbtest, "SIZEOF_IV_SEGMENT", mrb_fixnum_value(mrb_sizeof_iv_segment));
+  mrb_define_const(mrb, mrbtest, "COUNTOF_IV_SEGMENT", mrb_fixnum_value(mrb_countof_iv_segment));
```
matz, line 239: I  don't think that disclosing (possibly fragile) internal for continuous tests is a good idea. This is my personal opinion.
Besides that, too many test checks slow down the test performance.

src/variable.c
```diff
@@ -1132,11 +1131,17 @@ mrb_class_find_path(mrb_state *mrb, struct RClass *c)
 mrb_int
 mrb_obj_iv_tbl_memsize(mrb_state* mrb, mrb_value obj)
 {
-  return sizeof(iv_tbl) +
-    (sizeof(segment) * ceil(iv_size(mrb, mrb_obj_ptr(obj)->iv)/
-                            MRB_IV_SEGMENT_SIZE));
+  size_t ivsize = iv_size(mrb, mrb_obj_ptr(obj)->iv);
+  size_t ivsegs = (ivsize + MRB_IV_SEGMENT_SIZE - 1) / MRB_IV_SEGMENT_SIZE;
+  return sizeof(iv_tbl) + (sizeof(segment) * ivsegs);
```
matz, line 1136: This change in `mrb_obj_iv_tbl_memsize` is great. It should remain.

shuujii: It seems that the memory size cannot be calculated by this way because deleted instance variable slots remain.

I have a feeling that deleted slots will disappear if `iv_del` does the followings:

* Move the last instance variable to the deleted slot.
* Free the last segment if it is empty.

mrbgems/mruby-os-memsize/src/memsize.c
dearblue: I changed.

mrbgems/mruby-os-memsize/src/memsize.c
dearblue: I made the `obj_is_kind_of_checked()` function take a string as an argument.

mrbgems/mruby-os-memsize/test/memsize.rb
dearblue: This test itself has been deleted.

src/variable.c
matz: I tried to address `iv_tbl` shrink but found out that removing instance variables is very rare. So I concluded it's not worth the complexity. Maybe we will reimplement the `iv_tbl` in the future (to reduce memory). We will revisit removing `iv` then.

PR #5051, opened 2020-07-26
Add `mrb_static_assert1()` macro

dearblue:
This macro function is a variant of `mrb_static_assert()` that takes only one argument.

----

I'll use it in a patch that removes `ecall()` that I think I can suggest in the near future.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -62,10 +62,14 @@
 #define mrb_assert_int_fit(t1,n,t2,max) ((void)0)
 #endif
 
-#if defined __STDC_VERSION__ && __STDC_VERSION__ >= 201112L
-#define mrb_static_assert(exp, str) _Static_assert(exp, str)
+#if (defined __STDC_VERSION__ && __STDC_VERSION__ >= 201112L) || \
+    (defined __cplusplus && __cplusplus >= 201103L)
+#include <assert.h>
+#define mrb_static_assert(exp, str) static_assert(exp, str)
+#define mrb_static_assert1(exp) static_assert(exp, #exp)
 #else
 #define mrb_static_assert(exp, str) mrb_assert(exp)
+#define mrb_static_assert1(exp) mrb_assert(exp)
 #endif
 
 #include "mrbconf.h"


matz:
Could you explain the intention behind the pull-request?
Actually those `static_assert` lines are introduced with the past PR without considering much.
Do we really need it?

dearblue:
I want to check the size of the structure at compile time.
The existing `mrb_static_assert()` takes two arguments, but the motivation is that it's hard to think of the message and you want to omit it.

In a commit a week ago, when using `mrb_static_assert()` the same expression and message was applied.
https://github.com/dearblue/mruby/commit/6c4ed12900de90218d004d9112a2e569976b2f67#diff-0508f329afb350b1d3332d25d56166adR3214

```c
   mrb_static_assert(sizeof(struct mrb_irep_catch_hander) == 7, "sizeof(struct mrb_irep_catch_hander) == 7");
```

matz:
After some consideration, I decided to **remove** `static_assert` for several reasons:

- those assertions are incomplete. We can find those issues with ASAN as well.
- the current fallback implementation uses *dynamic* assertion, which does not really work like `static_assert`

So, I will remove the lines that use `mrb_static_assert` from the core and bundled gems, and make fallback implementation empty for compatibility.

shuujii:
There is a way to emulate a `static_assert` that is evaluated at compile time.

```c
#define _mrb_static_assert_cat0(a, b) a##b
#define _mrb_static_assert_cat(a, b) _mrb_static_assert_cat0(a, b)
#ifdef __COUNTER__
# define _mrb_static_assert_id(prefix) _mrb_static_assert_cat(prefix, __COUNTER__)
#else
# define _mrb_static_assert_id(prefix) _mrb_static_assert_cat(prefix, __LINE__)
#endif
#define mrb_static_assert(exp, str) typedef char _mrb_static_assert_id(_mrb_static_assert_)[(exp) ? 1 : -1]

mrb_static_assert(sizeof(RVALUE) <= sizeof(void*) * 6, "RVALUE size must be within 6 words");
```

Since assertions themselves are useful, they should at least be enabled in the environment with `static_assert`, rather than removed, I think.

In particular, I think the above `RVALUE` size check is very useful since it is often unintentionally violated.

matz:
I tried your version of `static_assert` but I get the following warnings for every `mrb_static_assert`.
```
warning: typedef ‘_mrb_static_assert_0’ locally defined but not used [-Wunused-local-typedefs]
```
I will try different versions from the Internet.

shuujii:
> I tried your version of `static_assert` but I get the following warnings for every `mrb_static_assert`.
>
> ```
> warning: typedef ‘_mrb_static_assert_0’ locally defined but not used [-Wunused-local-typedefs]
> ```

Could you please tell me about the following, as it doesn't reproduce in my environment?

* What is the compiler version?
* Will the warning be given even if used outside of functions?

matz:
- `gcc -v` says `gcc version 9.3.0 (Ubuntu 9.3.0-10ubuntu2)`. I compiled with `-x c++ -std=gnu++03`.
- if `mrb_static_assert()` is used outside of functions, it doesn't warn anything.

shuujii:
Thank you for your reply.

Maybe the warning can be suppressed by `_Pragma`. If anything, I personally think it's better to be able to use `mrb_static_assert` outside of functions as well as `static_assert`.

matz:
I understand. If there's any way we can use `static_assert()` outside of functions without warnings, I'd love to merge.

dearblue:
@shuujii
> ```c
> #define mrb_static_assert(exp, str) typedef char _mrb_static_assert_id(_mrb_static_assert_)[(exp) ? 1 : -1]
> ```

Using `struct` instead of `typedef` seems to get rid of the warning.

```c
#define mrb_static_assert(exp, str) struct _mrb_static_assert_id(_mrb_static_assert_) { char x[(exp) ? 1 : -1]; }
```

I confirmed with gcc 6, 7, 8, 9, 10, 11 and clang 6, 7, 8, 9, 10, 11.

Do you make a pull request?

shuujii:
@dearblue Great! I will make a pull request. Thank you!

PR #5052, opened 2020-07-26
Extend the `cipush()` and `cipop()` functions

dearblue:
- Returns the updated call info.
- Unify the processing around `cipush()`.
- `cipop()` restores the stack.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -269,13 +269,12 @@ top_proc(mrb_state *mrb, struct RProc *proc)
 #define CI_ACC_RESUMED -3
 
 static inline mrb_callinfo*
-cipush(mrb_state *mrb)
+cipush(mrb_state *mrb, const mrb_code *pc, int push_stacks, int acc,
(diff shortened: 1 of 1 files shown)


matz:
From those three changes, the first and the last ones look OK to me.
I am a little bit concerned about the middle one. It's shorter but harder to read. Caller arguments do not have names, so you have to remember the meaning of those 8 arguments. I know it has a positive effect that prevents initialization errors, but still. Any opinion?

dearblue:
The operation of `cipush()` and `cipop()` was one of the parts that was hard to understand when I started reading `src/vm.c`.
It meant that it was hard to figure out which range in the code was involved.
If it is said that it is "just getting used to", it is until that point, but I think that it is easier to understand if it is organized.

Should I reduce the arguments to about `cipush(mrb_state *mrb, int push_stacks, int acc)`?
Or are there still more?

matz:
After some consideration, I decided to accept this PR.

PR #5063, opened 2020-08-11
Use `struct _stat32` instead of `struct __stat32`

dearblue:
It is described as `struct __stat32` in the MSVC reference manual.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2019
But it doesn't really exist, so it must use `struct _stat32`.

It also replaces `struct __stat64` with `struct _stat64` to make it look nicer.

### DIFF
--- a/mrbgems/mruby-io/src/file.c
+++ b/mrbgems/mruby-io/src/file.c
@@ -71,11 +71,11 @@ typedef struct stat         mrb_stat;
 # define mrb_stat(path, sb) stat(path, sb)
 # define mrb_fstat(fd, sb)  fstat(fd, sb)
 #elif defined MRB_INT32
-typedef struct __stat32     mrb_stat;
+typedef struct _stat32      mrb_stat;
 # define mrb_stat(path, sb) _stat32(path, sb)
 # define mrb_fstat(fd, sb)  _fstat32(fd, sb)
 #else
-typedef struct __stat64     mrb_stat;
+typedef struct _stat64      mrb_stat;
 # define mrb_stat(path, sb) _stat64(path, sb)
 # define mrb_fstat(fd, sb)  _fstat64(fd, sb)
 #endif


matz:
Thank you for the PR. But on 32bit MinGW, we got the following error:
```
D:/a/mruby/mruby/mrbgems/mruby-io/src/file.c:400: undefined reference to '__imp__fstat32'
```

The stackoverflow says it is caused by calling 32bit library function from 64bit application. Is there any workaround?

I wanted to use 32bit `time_t` because `(mrb_int)time64_t` caused out-of-range error. If we can cast 64bit `time_t` into 32bit `mrb_int`, it is OK too.

matz:
By the latest CI build, everything OK on GitHub Actions. On AppVeyor, a test fails with 32bit build with the error.
```
ArgumentError: File#mtime => -550329529 out of Time range (mrbgems: mruby-io)
```
which indicates something was wrong with `time_t` to `mrb_int` conversion. But the time (current time) should fit in 32bit, so I have no idea why it fails.

dearblue:
It's about the `mruby3` branch, right?

The current time uses up to 31 bits.
However, in `MRB_INT32` + `MRB_WORD_BOXING`, the sign is up to 31 bits, so it seems that it is erroneously converted to a negative integer.

```console
% ruby27 -e'n = Time.now.to_i; p n | n[30] * -0x80000000'
-550290679
```

The workaround I came up with is:
- Use `MRB_NO_BOXING` (it crashed...)
- Convert with `mrb_float_value()` instead of `mrb_fixnum_value()` (not tried)

I don't know `__imp__fstat32`.

dearblue:
Sorry.

> - Use `MRB_NO_BOXING` (it crashed...)

It was a mistake to get caught in `mrb_static_assert` instead of crash.

```
In file included from /tmp/mruby3/src/gc.c:9:0:
/tmp/mruby3/src/gc.c: In function 'mrb_init_gc':
/tmp/mruby3/include/mruby.h:72:38: error: static assertion failed: "RVALUE size must be within 6 words"
 # define mrb_static_assert(exp, str) _Static_assert(exp, str)
                                      ^
/tmp/mruby3/src/gc.c:1623:3: note: in expansion of macro 'mrb_static_assert'
   mrb_static_assert(sizeof(RVALUE) <= sizeof(void*) * 6,
   ^
```

I tried this on FreeBSD 12.1 using old mingw32 (not 64 bit version) and wine (i386-wine).

`target/cross-mingw32.rb`
```ruby
load File.join(__dir__, "host.rb")

MRuby::CrossBuild.new("cross-mingw32") do |conf|
  toolchain :gcc

  conf.compilers.each do |cc|
    cc.command = "mingw32-gcc"
    #cc.defines << %w(MRB_NAN_BOXING)
    cc.defines << %w(MRB_NO_BOXING)
    #cc.defines << %w(MRB_NO_BOXING MRB_INT64)
  end
  exts.executable = ".exe"
  archiver.command = "mingw32-ar"
  linker.command = "mingw32-gcc"
  linker.libraries << %w(ws2_32)

  gem core: "mruby-bin-mruby"
  gem core: "mruby-io"
  gem core: "mruby-print"
  gem core: "mruby-sprintf"
  gem core: "mruby-time"
  conf.test_runner.command = "wine"
end
```

matz:
Ah, I didn't notice 31bit overflow. Let me think a while.

Issue #5047, opened 2020-07-23
Introduce malloc_trim on linux to fix dirty page leak


RoryO:
To begin please understand that mruby is not doing anything incorrectly. There is no issue with the way mruby allocates and releases memory and does not have a memory leak. This is a discussion on using the glibc malloc/realloc/free as default will not return dirty allocated pages back to the kernel managed heap automatically. This issue exists on any dynamic language using glib allocator. 

Summary
===

- The glib allocator doesn't return dirty pages back to the OS heap automatically until the program allocations meet rare conditions
- Calling glib specific `malloc_trim` forces glib allocator to return dirty heap pages
- I propose adding a call to `malloc_trim` on a major GC and as a final step in `mrb_close`
- Asking for advice on the lowest impact way of adding this code

Research
===

Following up on @FooBarWidget [absolutely wonderful research about the glibc allocator](https://www.joyfulbikeshedding.com/blog/2019-03-14-what-causes-ruby-memory-bloat.html) I wanted to test the same with mruby. The short summary from his research is by default glibc `free` does not automatically mark free pages as unused with `madvise` without hitting a frankly absurd page usage threshold. 

This is not a new discovery. It is obscure though discussed elsewhere through the years, and on C Ruby
- https://stackoverflow.com/a/47061458
- https://stackoverflow.com/a/3860495
- https://bugs.ruby-lang.org/issues/15667

We can force marking the free pages by calling the glib specific `malloc_trim`. The results I observed with mruby are the same as presented with C Ruby, which is not a surprise.

Here is what I used for a test program along with a Rake file for automation. First, I built `libmruby.a` from master using defaults, and then `libmruby-gcpatched.a` using this patch. It calls `malloc_trim` on a major GC cycle

```patch
diff --git a/src/gc.c b/src/gc.c
index fd4fb240..8d653dfb 100644
--- a/src/gc.c
+++ b/src/gc.c
@@ -19,6 +19,7 @@
 #include <mruby/gc.h>
 #include <mruby/error.h>
 #include <mruby/throw.h>
+#include <malloc.h>

 /*
   = Tri-color Incremental Garbage Collection
@@ -1314,7 +1315,7 @@ mrb_full_gc(mrb_state *mrb)
     gc->majorgc_old_threshold = gc->live_after_mark/100 * MAJOR_GC_INC_RATIO;
     gc->full = FALSE;
   }
-
+  malloc_trim(0);
   GC_TIME_STOP_AND_REPORT;
 }
```

[...]

PR #5069, opened 2020-08-30
Adding malloc_trim call on full GC

RoryO:
Follow up on #5047. Does a couple things:

- Adds a few methods to `MRuby::Command::Compiler` for asking about headers and functions. I took inspiration from [how meson does it](https://github.com/mesonbuild/meson/blob/6f74692ae0fe4fda8a133cc1627c4742331a52c7/mesonbuild/compilers/mixins/clike.py), which has a nice API.
- Wraps the necessary includes and call within a #define set inside `gcc.rake`. That way if a user has a custom `build_config.rb` they can remove the define if they wish, like if they are using a different allocator than glib.

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -1,4 +1,5 @@
 require 'forwardable'
+require 'tempfile'
 
 module MRuby
   class Command
@@ -126,6 +127,31 @@ module MRuby
       end
     end
 
+    def compiles?(source_text)
+      infile = Tempfile.new ['', '.c']
+      infile.write source_text
+      infile.close
+      cwd = Dir.pwd
+      is_success = false
+      # Change to a tmp dir when compiling so we don't litter compiler artifacts
+      Dir.mktmpdir do |tmpdir|
+        Dir.chdir tmpdir
+        sh(command, infile.path, verbose: false) { |retval, _| is_success = retval }
+      end
+      infile.delete
+      Dir.chdir cwd
+
+      return is_success
+    end
+
+    def has_header?(header_name)
+      compiles? test_code_template header: header_name
+    end
+
+    def has_function?(function_name, with_header: nil)
+      compiles? test_code_template function: function_name, header: with_header
+    end
+
     private
 
     #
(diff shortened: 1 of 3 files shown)


lib/mruby/build/command.rb
```diff
@@ -162,6 +188,46 @@ def get_dependencies(file)
       end
       deps << MRUBY_CONFIG
     end
+
+    def test_code_template(function: nil, header: nil)
```
RoryO, line 192: This function is kind of ugly. I didn't know how far back we support Ruby so I didn't use the tilde `<<~` form of heredocs to make it a bit less ugly.

lib/mruby/build/command.rb
```diff
@@ -126,6 +127,31 @@ def define_rules(build_dir, source_dir='')
       end
     end
 
+    def compiles?(source_text)
```
RoryO, line 130: @matz I hope this is sufficient for what you were thinking about adding autoconf like behavior when we need.

lib/mruby/build/command.rb
```diff
@@ -126,6 +127,31 @@ def define_rules(build_dir, source_dir='')
...
+      infile = Tempfile.new ['', '.c']
+      infile.write source_text
+      infile.close
+      cwd = Dir.pwd
+      is_success = false
+      # Change to a tmp dir when compiling so we don't litter compiler artifacts
+      Dir.mktmpdir do |tmpdir|
+        Dir.chdir tmpdir
+        sh(command, infile.path, verbose: false) { |retval, _| is_success = retval }
```
RoryO, line 139: There is a bit of a nuisance here. You cannot silence stderr with the Rake `sh` method. When testing for features it may be a legitimate failure, which is why we're asking the compiler. Unfortunately gcc and clang output the compiler errors to `stderr`. Doing a shell redirection isn't cross platform. We could use `Open3.capture3` instead of `sh`, I didn't know if it is okay to require and use.

RoryO:
It appears either mingw or the CI system doesn't allow removing temp directories. That is a strange situation either way. I will work around it.

matz:
I am wondering whether it may be better to specify manually via a macro define, e.g. `MRB_USE_MALLOC_TRIM`.
In that case, we don't have to add a complex detection code. No worry about cross-compiling nor compiler portability.

RoryO:
Oh yes it should be prefixed with `MRB_`.

Are you saying you would like it off by default and enabling it with a custom `build_config`?

matz:
@RoryO yes. Altough it's less intelligent. I'd like to follow KISS principle here.

RoryO:
Okay, I have done that.

matz:
Thank you! It will be merged to `mruby3` branch.

RoryO:
Oh that’s new! Okay I changed the base branch.

matz:
Oh, you didn't have to. I did it on my side.

matz:
Thank you anyway.

RoryO:
Should this be closed then?

matz:
I will close when I push to GitHub. I am struggling with other issues. Wait for a while.

matz:
Merged.

PR #5072, opened 2020-09-04
Skip C and Ruby comments from scan target

udzura:
Keep build OK when broken codes are included in comments.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -179,8 +179,8 @@ file presym_file => cfiles+rbfiles+psfiles+[__FILE__] do
   end
   rbsymbols = rbfiles.map do |f|
     src = File.read(f)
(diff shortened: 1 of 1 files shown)


matz:
I understand the concept. But mere skipping does not work, since test suites contain symbols like `:"#{1+2}`.
There are a few choices:
- avoid using those weird symbols in test suites (they are not used in the real applications)
- implement better code scanner

udzura:
At this moment the problem I got should be solved by https://github.com/mruby/mruby/pull/5074 !
I found scanning comments (and adding unused symbols to presym) itself might not be a big problem...

matz:
Addressed by b2e92f60c0e

PR #5074, opened 2020-09-05
Ensure exact match for symbols like `foo!/foo?/foo=` on presym.inc generation

udzura:
* e.g. symbols like "foo[]=" make invalid C codes... even in comments!
* https://github.com/mattn/mruby-http/blob/master/src/mrb_http.c#L652

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -218,13 +218,13 @@ file presym_inc => presym_file do
         f.print "MRB_PRESYM_CSYM(#{sym}, #{i+1})\n"
       elsif op_table.key?(sym)
         f.print "MRB_PRESYM_QSYM(\"#{sym}\", #{op_table[sym]}, #{i+1})\n"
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -218,13 +218,13 @@ file presym_inc => presym_file do
         f.print "MRB_PRESYM_CSYM(#{sym}, #{i+1})\n"
       elsif op_table.key?(sym)
         f.print "MRB_PRESYM_QSYM(\"#{sym}\", #{op_table[sym]}, #{i+1})\n"
-      elsif /\?\Z/ =~ sym
+      elsif /\A\w+\?\Z/ =~ sym
```
matz, line 221: It should be `[A-Za-z_]\w*` instead of `\w+` (`[A-Za-z0-9_]+`).

udzura: Sure, fixed!

PR #5079, opened 2020-09-17
Fix *.d parsing with gcc 9.3.0

wataash:
```
Before this commit:

  $ rake
  ...
  $ rm build/host/src/gc.o
  $ rake
  rake aborted!
  Don't know how to build task '/home/foo/mruby/build/host/src/gc.o' (See the list of available tasks with `rake --tasks`)
  ...

After this commit:

  $ rake
  ...
  $ rm build/host/src/gc.o
  $ rake
  CC    src/gc.c -> build/host/src/gc.o
  ...

With gcc 9.3.0 on Ubuntu 20.04, build/host/src/array.d looks like:

  /build/host/src/array.o: /src/array.c \
   /include/mruby.h /include/mrbconf.h \
   /include/mruby/common.h \
   ...
   /include/mruby/range.h \
   /src/value_array.h

and it has been parsed to:

  [
                                           # /src/array.c missing
    "/include/mruby.h /include/mrbconf.h", # incorrectly parsed
    "/include/mruby/common.h",
    ...
    "/src/value_array.h",
  ]

After this change, *.d will be parsed correctly.

  [
    "/src/array.c",
    "/include/mruby.h",
    "/include/mrbconf.h",
    "/include/mruby/common.h",
    ...
    "/src/value_array.h",
  ]
```

Ref. https://github.com/mruby/mruby/pull/4571
cc @shuujii

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -139,6 +139,15 @@ module MRuby
     #     /include/mruby/value.h \
     #     /src/value_array.h
     #
+    # ==== Without <tt>-MP</tt> compiler flag (gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0)
+    #
+    #   /build/host/src/array.o: /src/array.c \
+    #    /include/mruby.h /include/mrbconf.h \
+    #    /include/mruby/common.h \
+    #    ...
+    #    /include/mruby/range.h \
+    #    /src/value_array.h
+    #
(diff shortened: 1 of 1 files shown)


shuujii:
Thank you for your report.

It doesn't depend on the version of GCC, but if the paths are short, it seems to be concatenated without line breaks (reproduced in Clang and older GCC).

wataash:
ah thanks!
I generalized the comment on #5081.

PR #5084, opened 2020-09-18
Mruby3

matz:
Update to `mruby3`.

### DIFF
--- a/doc/mruby3.md
+++ b/doc/mruby3.md
@@ -1,7 +1,7 @@
 User visible changes in `mruby3`
 ===
 
-= Build System
+# Build System
 
 You can specify `TARGET` option to `rake` via a command line
 option, or via an environment variable, e.g.
@@ -15,13 +15,13 @@ or
 It's much easier to switch multiple targets than the
 previous `build_config.rb` system.
 
-== `presym` target
+## `presym` target
 
 The first compilation of `mruby` may require generation of a
 static symbol table named `build/presym`. You can generate
 the table by `rake gensym`.
 
-== `target` directory
+## `target` directory
 
 Build target specification files are loaded from `target`
 directory. The default `TARGET` is `host` which is described
@@ -39,15 +39,15 @@ If you want to have your target description out of the
 source tree, you can specify the path to the description
 file in `MRUBY_CONFIG`.
 
-= Build Target Contribution
+# Build Target Contribution
 
 When you write a new target description, please
 contribute. We welcome your contribution as a GitHub
 pull-request.
 
-= Languge Changes
+# Languge Changes
 
-== New Syntax
+## New Syntax
 
 We have ported some new syntax from CRuby.
 
@@ -55,54 +55,54 @@ We have ported some new syntax from CRuby.
 * Numbered block parameter (`x.map{_1 * 2}`)
 * End-less `def` (`def double(x) = x*2`)
 
-= Configuration Options Changed
+# Configuration Options Changed
 
 Some configuration macro names are changed for consistency
 
-== `MRB_NO_FLOAT`
+## `MRB_NO_FLOAT`
 
 Changed from `MRB_WITHOUT_FLOAT` to conform `USE_XXX` naming
 convention.
 
-== `MRB_USE_FLOAT32`
+## `MRB_USE_FLOAT32`
 
 Changed from `MRB_USE_FLOAT` to make sure `float` here means
 using single precision float, and not the opposite of
 `MRB_NO_FLOAT`.
 
-== `MRB_USE_METHOD_T_STRUCT`
+## `MRB_USE_METHOD_T_STRUCT`
 
 Changed from `MRB_METHOD_T_STRUCT`.
 
 To use `struct` version of `mrb_method_t`. More portable but consumes more memory.
 Turned on by default on 32bit platforms.
 
-== `MRB_NO_BOXING`
+## `MRB_NO_BOXING`
 
 Uses `struct` to represent `mrb_value`. Consumes more memory
 but easier to inveticate the internal and to debug. It used
 to be default `mrb_value` representation. Now the default is
 `MRB_WORD_BOXING`.
 
-== `MRB_WORD_BOXING`
+## `MRB_WORD_BOXING`
 
 Pack `mrb_value` in an `intptr_t` integer. Consumes less
 memory compared to `MRB_NO_BOXING` especially on 32 bit
 platforms. `Fixnum` size is 31 bits so some integer values
 does not fit in `Fixnum` integers.
 
-== `MRB_NAN_BOXING`
+## `MRB_NAN_BOXING`
 
 Pack `mrb_value` in a floating pointer number. Nothing
 changed from previous versions.
 
-== `MRB_USE_MALLOC_TRIM`
+## `MRB_USE_MALLOC_TRIM`
 
 Call `malloc_trim(0)` from mrb_full_gc() if this macro is defined.
 If you are using glibc malloc, this macro could reduce memory consumption.
 
-= Internal Changes
+# Internal Changes
 
-== `Random` now use `xoshiro128++`.
+## `Random` now use `xoshiro128++`.
 
 For better and faster random number generation.
--- a/doc/mruby3.md
+++ b/doc/mruby3.md
@@ -96,6 +96,11 @@ does not fit in `Fixnum` integers.
 Pack `mrb_value` in a floating pointer number. Nothing
 changed from previous versions.
 
+== `MRB_USE_MALLOC_TRIM`
+
+Call `malloc_trim(0)` from mrb_full_gc() if this macro is defined.
+If you are using glibc malloc, this macro could reduce memory consumption.
+
 = Internal Changes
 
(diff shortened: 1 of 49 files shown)


src/dump.c
```diff
@@ -918,4 +969,171 @@ mrb_dump_irep_cfunc(mrb_state *mrb, mrb_irep *irep, uint8_t flags, FILE *fp, con
...
+{
+  if (p->tt & IREP_TT_NFLAG) {  /* number */
+    switch (p->tt) {
+#ifdef MRB_64BIT
+    case IREP_TT_INT64:
+      if (p->u.i64 < INT32_MIN || INT32_MAX < p->u.i64) {
+        fprintf(fp, "{IREP_TT_INT64, {.i64=%" PRId64 "}},\n", p->u.i64);
+      }
+      else {
+        fprintf(fp, "{IREP_TT_INT32, {.i64=%" PRId64 "}},\n", p->u.i64);
```
dearblue, line 983: If it were `IREP_TT_INT32`, is it as follows?

```c
fprintf(fp, "{IREP_TT_INT32, {.i32=%" PRId32 "}},\n", (int32_t)p->u.i64);
```

include/mruby.h
```diff
@@ -943,11 +954,11 @@ typedef struct mrb_kwargs mrb_kwargs;
...
 struct mrb_kwargs
 {
-  uint32_t num;
-  mrb_value *values;
-  const char *const *table;
-  uint32_t required;
-  mrb_value *rest;
+  uint32_t num;                 /* number of keyword arguments */
+  uint32_t required;            /* number of required keyword arguments */
+  mrb_sym *table;               /* C array of symbols for keyword names */
```
dearblue, line 959: I think it's better to define it as `const mrb_sym *table` so that you can pass an array of `const mrb_sym`.

doc/limitations.md
```diff
@@ -17,7 +17,7 @@ Please help to improve it by submitting your findings.
 ## `1/2` gives `0.5`
 
 Since mruby does not have `Bignum`, bigger integers are represented
-by `Float` numbers. To enhance interoperability between `Fixnum`
+by `Float` numbers. To enhance interoperability between `Integer`
 and `Float`, mruby provides `Float#upto` and other iterating
 methods for the `Float` class.  As a side effect, `1/2` gives `0.5`
 not `0`.
```
dearblue, line 23: mruby3 now returns integers when dividing between integers and needs to be fixed.

doc/mruby3.md
```diff
@@ -0,0 +1,108 @@
+User visible changes in `mruby3`
+===
+
+= Build System
```
dearblue, line 4: The heading in the `mruby3.md` file is in rdoc format (`==`), so it needs to be modified to markdown format (`##`).

include/mruby/irep.h
```diff
@@ -16,28 +16,59 @@
...
+  IREP_TT_INT64 = 3,          /* 64bit integer */
+  IREP_TT_FLOAT = 5,          /* float (double/float) */
 };
 
-struct mrb_locals {
-  mrb_sym name;
-  uint16_t r;
+#define IREP_TT_NFLAG 1       /* number (non string) flag */
+#define IREP_TT_SFLAG 2       /* static string flag */
+#define IREP_TT_SFLAG 2       /* static string flag */
```
dearblue, line 28: The same macro is defined twice.

include/mruby/irep.h
```diff
@@ -46,6 +77,8 @@ typedef struct mrb_irep {
 } mrb_irep;
 
 #define MRB_ISEQ_NO_FREE 1
+#define MRB_IREP_NO_FREE 2
+#define MRB_IREP_STATIC  3
```
dearblue, line 81: `MRB_IREP_STATIC` probably means `(MRB_ISEQ_NO_FREE | MRB_IREP_NO_FREE)`.

dearblue (commented):
For these things, it's probably enough to merge the PRs and then work on them.

src/dump.c
matz: It should be. I will fix.

include/mruby.h
matz: Indeed. Thank you.

dearblue:
I have a question.

For now, `presym.inc` is commonly used for all targets, but are there any plans to separate it?
If not, is the concept acceptable?

As a factor when separating for each target I am thinking of:

- Instead of `build/repo/**/...`, get the symbols only from the required mrbgem. Symbols contained only in the test will be extracted if necessary.
- Instead of `build/presym.inc`, it will be output to `build/<TARGET>/include/mruby/build/presym.inc`.
- Add a switch to `bin/mrbc` to enter the contents of the `presym` file. It is used with the `-S` switch.
- It may need to separate the build part of the `bin/mrbc` file from the "host" target.

matz:
To avoid repeated compilation, `presym' need to be generated before `mrbc`. That is the biggest constraint.
If you have any idea, let me know.

dearblue:
If it can enter the `presym` file into `bin/mrbc`, then the "host" target `bin/mrbc` can be used for other targets as well.

matz:
I am not sure whether I understand your statement. `mrbc` is built using the core, core using symbols generated from `presym`. If we update the `presym` we need to compile everything again.
If we use `mrbc` to generate `presym`, the build steps would be:
* build core without `presym`
* build `mrbc` using the first core
* generate binary files and `presym` using `mrbc`
* recompile everything again (this time including gems)

It is redundant. I have several ideas to avoid recompilation, but it requires a big rewrite regarding `mrbc`.

dearblue:
In my opinion:

- "host" will generate and compile `presym` in the current way.
- Generate a `presym` for each target.
- The per-target `*.rb` file is compiled by passing the per-target `presym` to the `bin/mrbc` of "host".

I think it's true that the change to `mrbc.c` will be large due to the addition of the ability to read the` presym` file.

shuujii:
My understanding is that mruby should be able to build without presym for the following reasons:

1. Regular expressions are currently used to extract presym from Ruby files, but it is not strict, so I think it is better to use mruby itself (`mrbc`) to parse them.
2. For example, adding a class, method, or instance variable to one Ruby file will change symbol IDs, so most files, including C files, will need to be recompiled. This is inconvenient, especially for trial and error during development.

To achieve 1, the core would have to be built twice with and without presym, but I don't think that's a big deal for the following reasons:

1. If only `mrbc` is built, using ccache and parallel builds, even my low spec machine can build it in about 1 second.
2. If we can specify pre-built `mrbc`, we don't have to build it every time. Because I think there are not so many changes which affect the result produced by `mrbc`, this should  be fine in most cases.

Achieving 2 also has the side effect of speeding up compilation of Ruby files because we can always use optimally built `mrbc` (for this reason, my personal repository has been doing something similar for some time).

I'll try to build without presym. However, even if it does, it will probably be partially incompatible with the current presym (C macros change). Also, I think it will take a while.

matz:
> Currently it doesn't work correctly at least in GC terms.

Regarding `ObjectSpace.count_objects`, similar things could happen with `T_FLOAT` in CRuby. So I think that's OK not to count immediate values. But your example may indicate the existence of a bug. Let me investigate it later. I couldn't reproduce the issu on my machine yet.

matz:
Regarding `presym` generation, using `mrbc` sounds tempting. But rewriting the build process needs more time that I have at hand. Maybe later.

shuujii:
> Regarding ObjectSpace.count_objects, similar things could happen with T_FLOAT in CRuby. So I think that's OK not to count immediate values.

The variable `int` (`1<<30`) in that example is not immediate.

> But your example may indicate the existence of a bug.

As mentioned before, `mrb_immediate_p` is true even for `Integer` which is not an immediate, so the cause is the object is not a GC target, I think.

I think there are various other bugs caused by the same cause, such as the following.

```ruby
p 1073741824 == 1073741824+0  #=> false
```

dearblue:
`MRuby::Build.new` now always generates `bin/mrbc`, even on non-hosts.
Is this intentional?

```ruby
# build_config.rb

MRuby::Build.new do
  toolchain "clang"
end

MRuby::Build.new("host2") do
  toolchain "clang"
end
```

```console
% rake -m MRUBY_CONFIG=build_config.rb
...SNIP...

Build summary:

================================================
      Config Name: host
 Output Directory: build/host
         Binaries: mrbc
    Included Gems:
             mruby-bin-mrbc - mruby compiler executable
             mruby-compiler - mruby compiler library
================================================

================================================
      Config Name: host2
 Output Directory: build/host2
         Binaries: mrbc
    Included Gems:
             mruby-bin-mrbc - mruby compiler executable
             mruby-compiler - mruby compiler library
================================================
```

shuujii:
> As mentioned before, `mrb_immediate_p` is true even for `Integer` which is not an immediate

I'm very sorry, `mrb_immediate_p` is no problem with Word-boxing because it does not use type tag. Also, with No-boxing or NaN-boxing, `Integer` is always an immediate value, so` mrb_immediate_p` was not affected.

> so the cause is the object is not a GC target, I think.

This again is not a problem with `mrb_immediate_p`, so I'll create a PR later.

PR #5087, opened 2020-09-24
Prohibit string changes by "s"/"z" specifier of `mrb_get_args()`

dearblue:
- The `s` specifier is a string pointer obtained without performing `mrb_str_modify()`, so it cannot be changed.
- The `z` specifier cannot be changed because it is a string pointer obtained by `RSTRING_CSTR()` which returns `const char *`.

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -584,8 +584,8 @@ void mrb_hash_check_kdict(mrb_state *mrb, mrb_value self);
     S:      String         [mrb_value]            when ! follows, the value may be nil
     A:      Array          [mrb_value]            when ! follows, the value may be nil
     H:      Hash           [mrb_value]            when ! follows, the value may be nil
-    s:      String         [char*,mrb_int]        Receive two arguments; s! gives (NULL,0) for nil
-    z:      String         [char*]                NUL terminated string; z! gives NULL for nil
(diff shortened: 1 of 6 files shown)


src/class.c
```diff
@@ -584,8 +584,8 @@ void mrb_hash_check_kdict(mrb_state *mrb, mrb_value self);
     S:      String         [mrb_value]            when ! follows, the value may be nil
     A:      Array          [mrb_value]            when ! follows, the value may be nil
     H:      Hash           [mrb_value]            when ! follows, the value may be nil
-    s:      String         [char*,mrb_int]        Receive two arguments; s! gives (NULL,0) for nil
-    z:      String         [char*]                NUL terminated string; z! gives NULL for nil
+    s:      String         [char const*,mrb_int]  Receive two arguments; s! gives (NULL,0) for nil
+    z:      String         [char const*]          NUL terminated string; z! gives NULL for nil
```
matz, line 588: L775 uses `const char*` and here it's `char const*`. Aren't they inconsistent?

dearblue: Since `"d"` is `TYPENAME const`, I intended to match it.
C/C++ recognizes that it is equivalent to `const TYPENAME`, so there is no problem.
However, in addition to that, human beings are certainly confused when the pointer is involved, so I will correct it.

Issue #5085, opened 2020-09-19
[RFC] Support startless/endless `Range` (`N..`, `..N`)


wataash:
This RFC proposes:

1. Let mruby follow the Ruby's behavior of `Range` with `nil` argument:

```rb
                                # ruby 2.7.1p83   mruby-latest (2020-09-19)
puts(Range.new(1, 2))           # 1..2            1..2
puts(Range.new(1, nil))         # 1..             mruby: bad value for range (ArgumentError)
puts(Range.new(nil, 2))         # ..2             mruby: undefined method '<=>' (NoMethodError)
puts(Range.new(nil, nil))       # .. (!)          mruby: undefined method '<=>' (NoMethodError)

puts(Range.new(1, 2, true))     # 1...2           1...2
puts(Range.new(1, nil, true))   # 1...            mruby: bad value for range (ArgumentError)
puts(Range.new(nil, 2, true))   # ...2            mruby: undefined method '<=>' (NoMethodError)
puts(Range.new(nil, nil, true)) # ... (!)         mruby: undefined method '<=>' (NoMethodError)
```

Japanese Industrial Standards (JIS) X3017:2013 (which is equivalent to ISO/IEC 30170:2012) _5.2.14 Range_ section doesn't define this behavior.

2. If 1. is accepted, implement the startless/endless `Range` literal.

```rb
            # ruby 2.7.1p83                 mruby-latest (2020-09-19)
puts(1..2)  # 1..2                          1..2
puts(1..)   # 1..                           syntax error, unexpected ')'
puts(..2)   # ..2                           syntax error, unexpected tDOT2, expecting ')'
puts(1...2) # 1...2                         1..2
puts(1...)  # 1...                          syntax error, unexpected ')'
puts(...2)  # ...2                          syntax error, unexpected tDOT2, expecting ')'

# just for reference:
puts(..)    # syntax error, unexpected ')'  syntax error, unexpected tDOT2, expecting ')'
puts(...)   # unexpected ...                unexpected argument forwarding ...
```

Though startless/endless syntax is *NOT* allowed in JIS X3017:2013.

## Why?

Just to mitigate the cost of Ruby->mruby porting. Currently we have to do the following conversions:

- `arr[1..]` -> `arr[1..-1]`
- `arr[..2]` -> `arr[0..2]`

As my personal opinion, implementing this RFC could be accepted because it doesn't much increase the complexity of sources (just adding about 10 lines?).

## References

[...]

PR #5093, opened 2020-10-13
Introduce endless range (a part of #5085)

taiyoslime:
We introduced endless range feature, which is proposed in https://github.com/mruby/mruby/issues/5085 and already supported by CRuby.

### Current Behavior
```
> Range.new(1,nil)
bad value for range (ArgumentError)
```

### Example behavior after this commit
```
# mruby 
> Range.new(1,nil)
 => 1..
> (1...)
 => 1...
> (1..).each {|n| p n}
1
2
3
(snip)
> ('a'...).each {|c| p c; break if c>='f'}
"a"
"b"
"c"
"d"
"e"
"f"
 => nil
 > "Ruby"[1..]
 => "uby"
 > a = [0, 1, 2, 3, 4]; a[1..] = 5
 => 5
 > a
 => [0, 5]
```
```
# CRuby(2.6.6)
2.6.6 :001 > Range.new(1,nil)
 => 1.. 
2.6.6 :002 > (1...)
 => 1... 
2.6.6 :001 > (1..).each {|n| p n}
1
2
3
(snip)
2.6.6 :011 > ('a'...).each {|c| p c; break if c>='f'}
"a"
"b"
"c"
"d"
"e"
"f"
 => nil
2.6.6 :001 > "Ruby"[1..]
 => "uby" 
2.6.6 :002 > a = [0, 1, 2, 3, 4]; a[1..] = 5
2.6.6 :003 > a
 => [0, 5]
```

For other changes, please see the diff.

### Things to notice
- Based on CRuby, `Range#size` returns `Float::INFINITY` when it is endless. Under `MRB_WITHOUT_FLOAT` environment, we implemented to return nil for now.
```
# CRuby(2.6.6)
2.6.6 :015 > (1..).size
 => Infinity
 ```
```
# mruby
> (1..).size
=> Infinity
```
```
# mruby (built with MRB_WITHOUT_FLOAT flag)
> (1..).size
=> nil
```

- `Range#last`
    - According to ISO, `Range#last` is an alias of `Range#end` (15.2.14.4.10), which is actually not the case in current Ruby anymore (e.g. `Range#last` can take 0 or 1 argument, but `Range#end` does not take arguments.). Since the core of mruby basically needs to follow ISO, those features in CRuby have been implemented as extension(mruby-range-ext) in current mruby.
    - In CRuby, `Range#last` raises a RangeError when it is endless. (while `Range#end` returns nil). So we implemented this feature to extension(mruby-range-ext)'s `Range#last` following that implementation.
```
# CRuby(2.6.6)
2.6.6 :002 > (1..).last
Traceback (most recent call last):
        5: from /home/wataru/ruby/bin/irb:23:in `<main>'
        4: from /home/wataru/ruby/bin/irb:23:in `load'
        3: from /home/wataru/.rvm/gems/ruby-2.6.6/gems/irb-1.2.4/exe/irb:11:in `<top (required)>'
        2: from (irb):2
        1: from (irb):2:in `last'
RangeError (cannot get the last element of endless range)
```
```
# mruby (core)
> (1..).last
 => nil
```
```
# mruby (with mruby-range-ext)
> (1..).last
cannot get the last element of endless range (RangeError)
```

- To implement `Range#each`, we need to prepare `upto` method of String corresponding to endless range separately from the existing `String#upto` (CRuby also takes the same steps). According to the other implementation in mruby, we implement `String#__upto_endless` for private in use.


matz:
I will merge this PR after updating `master` by `mruby3` branch.

PR #5097, opened 2020-10-18
Prohibit changes by `RARRAY_PTR()` and `ARY_PTR()`

dearblue:
Previously, explicit write-barriers were required to change content using these macros.

From now on it can use `ARY_PTR_NEED_WB()` instead if it wants to get a pointer without `const` that requires write-barrier.
But if it wants to change only a few contents of the array object, the `mrb_ary_set()` function will be more useful.

With this change, the following function specifications will change:

- The `mrb_get_argv()` function will now return `const mrb_value *`.
  This is because it is difficult for the caller to check if it is a splat argument (array object) and to write-barrier if necessary.
- The "a"/"*" specifier of the `mrb_get_args()` function will now return `const mrb_value *`.
  This is because it is difficult for the caller to check if it is an array object and write-barrier if necessary.
  And it requires calling `mrb_ary_modify()` on the unmodified array object, which is also difficult (this is similar to #5087).

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -889,7 +889,7 @@ MRB_API struct RClass* mrb_define_module_under_id(mrb_state *mrb, struct RClass 
  * | `H`  | {Hash}         | {mrb_value}       | when `!` follows, the value may be `nil`           |
  * | `s`  | {String}       | const char *, {mrb_int} | Receive two arguments; `s!` gives (`NULL`,`0`) for `nil` |
  * | `z`  | {String}       | const char *      | `NULL` terminated string; `z!` gives `NULL` for `nil` |
- * | `a`  | {Array}        | {mrb_value} *, {mrb_int} | Receive two arguments; `a!` gives (`NULL`,`0`) for `nil` |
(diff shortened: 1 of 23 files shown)


shuujii:
I'm against this breaking changes because I think changing the behavior of `RARRAY_PTR`/`ARY_PTR` have a big impact. At least my product at hand will not work.

If you want to ensure that array content is not changed, it would be better to create another APIs (e.g. `RARRAY_CONST_PTR`/`ARY_CONST_PTR`). I kind of remember that similar APIs was added to CRuby when RGenGC or something was introduced.

matz:
I agree with @shuujii regarding `MRB_ARY_PTR_NEED_WB`. It is better but incompatible too much.
Maybe some changes regarding `mrb_get_args` are acceptable.

dearblue:
I checked CRuby's `RARRAY_PTR()` again, but it doesn't have `const`.
What I made a mistake, and I feel like I was to be bewitched by a fox...

I will withdraw this proposal because their opinions are plausible.
I will make a pull request related to `mrb_get_args()` later.

Issue #5096, opened 2020-10-16
Build system compatibility


shuujii:
### Default build configuration file

It seems that default build configuration file has changed from `MRUBY_ROOT/build_config.rb` to `MRUBY_ROOT/target/host.rb`, but I think this is unnecessary incompatibility and has a large impact.

For compatibility, I think it's better to use `MRUBY_ROOT/build_config.rb` if it exits when `MRUBY_CONFIG` or `MRUBY_TARGET` is not specified.

### `enable_debug` and `enable_test`

`enable_debug` and `enable_test` were not specified in `host` target in mruby2's default build configuration file, but they are specified in mruby3's default build configuration file. I think it is better not to specify these for the following reasons.

* `enable_debug` is turned off at #4045.
* Since default configurations are used in `ruby-build` etc., it is better not to have unnecessary options.
* `target/host-debug.rb` can be used for debugging. We may prepare `target/host-test.rb` for testing.

### The term "target"

I think the term "target" has been used mainly for things that can be obtained by `MRuby.each_target` and `MRuby.targets` (instance name of `MRuby::Build`). However, `MRUBY_TARGET` is used as name of build configuration file, which is a bit confusing.

Especially the following points are confusing.

* Multiple `MRuby::Build` instances can be specified in the build configuration file, so there is no one-to-one connection with `MRUBY_TARGET`.
* The required name `host` as instance name of `MRuby::Build` is the same as default value for `MRUBY_TARGET`.

matz:
`enable_debug` was turned off by default by 3c55cb0. `enable_test` was kept as it was turned on by default in mruby2's `build_config.rb`.
Regarding the term "target", I have no strong opinion. Do you have any idea? "config"? "build"?

shuujii:
> `enable_debug` was turned off by default by [3c55cb0](https://github.com/mruby/mruby/commit/3c55cb089514d40ff17e9f29f9195c85111e0c10). `enable_test` was kept as it was turned on by default in mruby2's `build_config.rb`.
> Regarding the term "target", I have no strong opinion. Do you have any idea? "config"? "build"?

shuujii:
I'm sorry, I made a wrong comment and closed this issue. I will comment again later.

shuujii:
> `enable_test` was kept as it was turned on by default in mruby2's `build_config.rb`.

`enable_test` was turned on in the `test` target, but off in the `host` target.

shuujii:
[...]

PR #5098, opened 2020-10-19
Revert `MRUBY_TARGET` mechanism; ref #5096

shuujii:
* In explanation of mruby, the expression `build_config.rb` is frequently
  used including official documents, so I think that it will not make sense
  if the file is no longer used.
* The `MRUBY_TARGET` mechanism seems to have little improvement, so I don't
  think it should be changed to avoid unnecessary confusion.
* `MRUBY_TARGET` and `MRuby.targets` represent somewhat different things,
  so using the same term "target" is a bit confusing.

The mechanism that can be written short when using a file under
`build_config` (renamed from `target`) directory remains
(`build_config/${MRUBY_CONFIG}.rb` is used if the path specified
in `MRUBY_CONFIG` doesn't exist).


tasks/benchmark.rake
```diff
@@ -9,7 +9,11 @@ def bm_files
 end
 
 def build_config_name
-  MRUBY_TARGET
+  if !ENV['MRUBY_CONFIG'].to_s.empty?
+    File.basename(ENV['MRUBY_CONFIG'], '.rb').gsub('build_config_', '')
+  else
+    "build"
```
matz, line 15: I am not sure the rationale behind this change. Why "build"?
Could you explain?

shuujii: I just put it back (I'm not sure intention of original author).

https://github.com/mruby/mruby/commit/fdbfeaf533fd75143f5d3af3ec6c585b9da746c4#diff-3c833bf555cc95c834d1765cffde8d995db0413412d33785fc0ea872cc74946a

However, in `Rakefile`, `ENV['MRUBY_CONFIG']` equated unspecified with empty, but that wasn't the case here, so I've unified it.

shuujii:
Thank you for merging!

Issue #5114, opened 2020-11-05
mruby-io doesn't work in process started as Windows service


kou:
#4968 added file descriptor check in `IO#initialize`.
It raises an exception at `IO.open(0, "r")` https://github.com/mruby/mruby/blob/master/mrbgems/mruby-io/mrblib/io.rb#L363 when a process is started as Windows service. Because a process started as Windows service doesn't have a valid standard input. `fstat(0, buf)` returns non-zero.

Here are solutions I think:

  1. Don't check file descriptor for `0`, `1` and `2`.
  2. Open an internal pipe for `0`, `1` and `2` when they are invalid.

Which solution do you like?

I think that 1. is enough.

FYI: CRuby uses 2.: `fill_standard_fds()` https://github.com/ruby/ruby/blob/master/ruby.c#L2556-L2584

matz:
I agree that (1) is sufficient. Could you prepare a pull-request? Or shall I work on it?

kou:
OK. I'll do it.

matz:
Hmm. Are you going to make a pull-request from the above commit? Or shall I merge it?

kou:
Sorry for not sharing the status of the commit.
I don't have an environment that can reproduce this problem. I need to ask our customer to confirm the commit fixes this problem.
It'll be done in a few days.

But the patch is simple and it'll fix this problem. So we can merge it without waiting the confirmation. (I'll share the result of the confirmation later.)
I've opened #5119.

kou:
We've confirmed that #5119 this problem.

matz:
Does this fix migw crash issue?

kou:
I think that this is unrelated to the MinGW crash issue.
But I don't know why the issue was occurred and fixed...

PR #5119, opened 2020-11-09
Don't check FD for STDIN/STDOUT/STDERR; #5114

kou:
We don't need to require valid STDIN/STDOUT/STDERR. If we require it, we
can't use mruby on environment that doesn't have valid
STDIN/STDOUT/STDERR such as Windows service process. Windows service
process doesn't have valid STDIN.


kou:
The MinGW build is failed but it doesn't show any details.
It's also failed on master: https://github.com/mruby/mruby/runs/1369511422
But it's succeeded on my fork: https://github.com/kou/mruby/runs/1371799791

PR #5144, opened 2020-11-18
Check if irep->reps is NULL

hifoolno:


matz:
Thanks for your contribution.

PR #5146, opened 2020-11-18
Fix integer overflow error in mrdb test

shuujii:


mrbgems/mruby-bin-debugger/bintest/mrdb.rb
```diff
@@ -63,10 +63,7 @@ def self.test(rubysource, testcase)
   # ruby source
   src = "foo = 'foo'\n"
 
-  str = ""
-  103.times {
-    str += "1234567890"
-  }
+  str = "'#{'1234567890' * 103}'"
   cmd = "p a=#{str}"
```
matz, line 67: The original `cmd` was `p a=12345....` and the updated `cmd` is `p a='1234...'`. Is this difference intentional?

shuujii: Yes. We must use non-integer to avoid integer overflow. Because this test is a confirmation that long line results in an error, this change seems to satisfy the purpose.

However, I think it is better not to get a syntax error when truncating line (e.g. `cmd[0...1023]`), so will fix it.

Thank you for your comment.

PR #5147, opened 2020-11-18
Check if irep->reps is NULL in lv_defined_p

hifoolno:


src/dump.c
```diff
@@ -773,8 +773,10 @@ lv_defined_p(const mrb_irep *irep)
 
   if (irep->lv) { return TRUE; }
 
-  for (i = 0; i < irep->rlen; ++i) {
```
matz, line 776: Ah, I merged this. But in reallity, `irep->rlen == 0` when `irep->reps == NULL`. So this original code was no problem.

hifoolno: Thank you for your reply~ I found there is a patch to add a check on  `irep->reps` in `mrb_irep_free` in https://github.com/mruby/mruby/commit/f1523d24042ca3416dc5b9be7b3fc220ddaed896, so I submitted this PR.

PR #5148, opened 2020-11-18
Check if irep->reps is NULL in get_irep_record_size

hifoolno:


src/dump.c
```diff
@@ -307,8 +307,10 @@ get_irep_record_size(mrb_state *mrb, const mrb_irep *irep)
   int irep_no;
 
   size = get_irep_record_size_1(mrb, irep);
-  for (irep_no = 0; irep_no < irep->rlen; irep_no++) {
-    size += get_irep_record_size(mrb, irep->reps[irep_no]);
+  if (irep->reps) {
```
matz, line 310: `irep->rlen == 0` when `irep->reps == NULL`. So the original code was no problem.

PR #5157, opened 2020-11-21
Allow to mixed and specify `*.rb` and `*.mrb` in `bin/mruby`

dearblue:
It is not decides by the extension.
The first 64 bytes of the file are read, and if there is a `NUL` symbol in it, it is recognized as a binary file and read as a `mrb` file.
If there is no `NUL` symbol, it is judged as a text file and it is processed as a Ruby script.

The `bin/mruby -b` switch is still available which treats the given file as a `.mrb` file.

New `MRB_API` function:
- `include/mruby/dump.h` and `src/load.c`
  - `mrb_load_detect_file_cxt()` (remove with `MRB_DISABLE_STDIO`)

New internal function (not static):
- `include/mruby/compile.h` and `mrbgems/mruby-compiler/core/parse.y`
  - `mrb_parse_file_continue()` (remove with `MRB_DISABLE_STDIO`)
  - `mrb_load_file_continue_cxt()` (remove with `MRB_DISABLE_STDIO`)

NOTE:
- Even script files now always open in binary mode for `bin/mruby`.
  The `\r\n` is also handled by the `nextc()` function already, so there is no problem even on Windows.
- The `nextc0()` function in `mrbgems/mruby-compiler/core/parse.y` can now specify a string buffer and a file pointer at the same time.
  In this case, get it from the string buffer first.

### DIFF
--- a/include/mruby/compile.h
+++ b/include/mruby/compile.h
@@ -116,6 +116,7 @@ struct mrb_parser_state {
   mrb_ast_node *cells;
   const char *s, *send;
 #ifndef MRB_DISABLE_STDIO
+  /* If both f and s are non-null, it will be taken preferentially from s until s < send. */
   FILE *f;
 #endif
   mrbc_context *cxt;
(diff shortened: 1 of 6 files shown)


matz:
Interesting idea. A few comments before merge.
* `mrb` files always start with `RITE` header mark. Probably you can use this to detect `mrb` files as well
* You don't have to remove `NUL` check to avoid accidental detection
*  I think `mrb_parse_file_continue()` and `mrb_load_file_continue_cxt()` should be private (static)
* `mrb_load_detect_file_cxt()` depends on `mruby-compiler` gem so it should not be in the core (`load.c`)
* I am not fully satisfied with the name `mrb_load_detect_file_cxt()` but I have no other name candidate now

dearblue:
indeed. I'll try.

dearblue:
Thank you for marge.

I didn't specifically state what I was able to achieve with this PR.
Although it is a brief explanation, I was able to achieve the following two things:

  - It can be specified in `bin/mruby` without distinguishing between ".rb" files and ".mrb" files.
  - It can be specified as a mixture of ".rb" and ".mrb" files, such as `bin/mruby -r lib1.mrb -r lib2.rb < app.mrb`.

matz:
Could you please add brief explanation to `doc/mruby3 
.md`?

dearblue:
Yes, it will be the result of Google Translate.
I will make a PR later.

PR #5159, opened 2020-11-21
Avoid loop initial declarations

dearblue:


matz:
Thank you. But I am thinking that it might be OK to allow loop initial declarations. Most compilers allow them. For example, micropython uses them already.

dearblue:
I welcome a change of policy!
In that case, can you revise this document?
[CONTRIBUTING.md#comply-with-c99-isoiec-98991999](https://github.com/mruby/mruby/blob/00751ccbcd14a6681f2a9ff1f6dce84708309bc0/CONTRIBUTING.md#comply-with-c99-isoiec-98991999)

matz:
In that case, we should. If anyone knows any platform with issues from this change, let us know.

shuujii:
> But I am thinking that it might be OK to allow loop initial declarations

This is good news for me too!

matz:
I checked `micropython` and `qmk_firmware` and several others. All of them use new style declarations.
I consider using them is OK now. Maybe we should update other targets as well (e.g. VC 2013).

PR #5171, opened 2020-11-22
Reduce the minimum string buffer size from 128 to 63 bytes

shuujii:
Similar change has been made in CRuby.

* https://github.com/ruby/ruby/pull/2151
* https://bugs.ruby-lang.org/issues/15802

Because mruby is more resource-saving oriented than CRuby, I think it's
better to reduce that.

In addition, the reason why 63 is used instead of 64 seems to be helpful as
follows.

* https://bugs.ruby-lang.org/issues/12025

### DIFF
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -134,7 +134,7 @@
 //#define POOL_PAGE_SIZE 16000
 
 /* initial minimum size for string buffer */
-//#define MRB_STR_BUF_MIN_SIZE 128
+//#define MRB_STR_BUF_MIN_SIZE 63
 
 /* arena size */
 //#define MRB_GC_ARENA_SIZE 100
@@ -204,7 +204,7 @@
 # endif
 
 # ifndef MRB_STR_BUF_MIN_SIZE
-#  define MRB_STR_BUF_MIN_SIZE 32
+#  define MRB_STR_BUF_MIN_SIZE 31
 # endif
 
 # ifndef MRB_HEAP_PAGE_SIZE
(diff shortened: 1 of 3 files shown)


include/mrbconf.h
```diff
@@ -204,7 +204,7 @@
 # endif
 
 # ifndef MRB_STR_BUF_MIN_SIZE
-#  define MRB_STR_BUF_MIN_SIZE 32
+#  define MRB_STR_BUF_MIN_SIZE 31
```
matz, line 207: This makes the default size 31 bytes. So it should be officially `31` or change here `63`. I vote for making the default value to `31`.

src/string.c
```diff
@@ -176,7 +176,7 @@ mrb_str_new_capa(mrb_state *mrb, size_t capa)
 }
 
 #ifndef MRB_STR_BUF_MIN_SIZE
-# define MRB_STR_BUF_MIN_SIZE 128
+# define MRB_STR_BUF_MIN_SIZE 63
```
matz, line 179: Since the value is already set in `mrbconf.h` lines above are meaningless (my fault).

dearblue: It is set in `mrbconf.h` if` MRB_CONSTRAINED_BASELINE_PROFILE` is defined. This is defined otherwise.
Therefore, it is better to change this when defining it to 31 bytes.

shuujii:
After investigation again, `MRB_STR_BUF_MIN_SIZE` is only used when `capa` passed to `mrb_str_buf_new` is less than `MRB_STR_BUF_MIN_SIZE`.

This behavior itself doesn't seem to make much sense, and this function doesn't seem to be used much either. Therefore, I feel that `mrb_str_buf_new` can be an alias for `mrb_str_new_capa` and `MRB_STR_BUF_MIN_SIZE` can be removed.

matz:
Agreed. Let's close this issue, and remove `MRB_STR_BUF_MIN_SIZE`.

matz:
Thank you for the suggestion.

PR #5194, opened 2020-11-26
Create presym files for each build target

shuujii:
Previously, presym files were always created in `build/{presym,presym.inc}`.
However, this constraint is inconvenient because it is common to use
multiple build configurations and build targets in a single mruby tree.
Therefore, change to create presym file for each build target.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,9 +25,9 @@ end
 # load custom rules
 load "#{MRUBY_ROOT}/tasks/core.rake"
 load "#{MRUBY_ROOT}/tasks/mrblib.rake"
-
 load "#{MRUBY_ROOT}/tasks/mrbgems.rake"
 load "#{MRUBY_ROOT}/tasks/libmruby.rake"
+load "#{MRUBY_ROOT}/tasks/presym.rake"
 
 load "#{MRUBY_ROOT}/tasks/benchmark.rake"
 
@@ -55,11 +55,8 @@ else
   depfiles = []
 end
 
-cfiles  = Dir.glob("#{MRUBY_ROOT}/src/*.c")
-rbfiles = Dir.glob("#{MRUBY_ROOT}/mrblib/**/*.rb")
-psfiles = []
 MRuby.each_target do |target|
-  gems.map do |gem|
+  gems.each do |gem|
     current_dir = gem.dir.relative_path_from(Dir.pwd)
     relative_from_root = gem.dir.relative_path_from(MRUBY_ROOT)
     current_build_dir = File.expand_path "#{build_dir}/#{relative_from_root}"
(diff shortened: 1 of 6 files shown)


shuujii:
The presym files will be output below.

* `build/NAME/presym`
* `build/NAME/include/mruby/presym.inc`

There were descriptions about the presym files in `doc/guides/compile.md`, but I deleted them for the following reasons.

* Originally only directories were described.
* Output destination, etc. is likely to change in the future.

matz:
I was surprised the change was far smaller than I thought.

dearblue:
I ran into a failure of `rake MRUBY_CONFIG=cross-32bit`.
When specifying a directory directly in `conf.gem`, it seems that the directory where `mruby/presym.inc` is located is not added.

```console
% rake30 -v MRUBY_CONFIG=cross-32bit
mkdir -p /var/tmp/mruby/build/cross-32bit/mrbgems/c_and_ruby_extension_example/src
CC    examples/mrbgems/c_and_ruby_extension_example/src/example.c -> build/cross-32bit/mrbgems/c_and_ruby_extension_example/src/example.o
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -m32 -DMRBGEM_C_AND_RUBY_EXTENSION_EXAMPLE_VERSION=0.0.0 -I"/var/tmp/mruby/include" -MMD -o "/var/tmp/mruby/build/cross-32bit/mrbgems/c_and_ruby_extension_example/src/example.o" -c "/var/tmp/mruby/examples/mrbgems/c_and_ruby_extension_example/src/example.c"
In file included from /var/tmp/mruby/include/mruby.h:96,
                 from /var/tmp/mruby/examples/mrbgems/c_and_ruby_extension_example/src/example.c:1:
/var/tmp/mruby/include/mruby/presym.h:19:10: fatal error: mruby/presym.inc: No such file or directory
   19 | #include <mruby/presym.inc>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
rake aborted!
Command failed with status (1): [gcc -std=gnu99 -g -O3 -Wall -Wundef -Werro...]
/var/tmp/mruby/lib/mruby/build/command.rb:37:in `_run'
/var/tmp/mruby/lib/mruby/build/command.rb:83:in `run'
/var/tmp/mruby/lib/mruby/build/command.rb:106:in `block (2 levels) in define_rules'
/var/tmp/mruby/Rakefile:118:in `block (2 levels) in <top (required)>'
/var/tmp/mruby/Rakefile:118:in `each'
/var/tmp/mruby/Rakefile:118:in `block in <top (required)>'
/usr/local/bin/rake30:23:in `load'
/usr/local/bin/rake30:23:in `<main>'
Tasks: TOP => /var/tmp/mruby/build/cross-32bit/bin/mruby => /var/tmp/mruby/build/cross-32bit/lib/libmruby.a => /var/tmp/mruby/build/cross-32bit/mrbgems/c_and_ruby_extension_example/src/example.o
```

dearblue:
Also, it will probably cause problems as it will use `host/bin/mrbc` with incompatible `presym` in `MRuby::CrossBuild`.
In that case `mrbgems/mruby-bin-mrbc` needs the ability to read any sorted `presym` files.

matz:
Hmm, considering cross-compiling, we might need to revert part of this pull-request.

matz:
Besides that, I faced multiple errors after this pull-request, such as
```
NoMethodError: undefined method `*' for nil:NilClass
/home/matz/work/mruby/tasks/presym.rake:81:in `block (4 levels) in <top (required)>'
```
Or
```
mruby/include/mruby/presym.h:14:48: error: redeclaration of enumerator 'MRB_SYM__pton'
   14 | #define MRB_PRESYM_NAMED(lit, num, type, name) MRB_##type##__##name = (num<<1),
      |                                                ^~~~
```

shuujii:
I will investigate at build errors later.

I would like to do the following for presym related, and I hope that if these are realized, most of the problems will be solved.

1. Make it buildable without presym.
2. Symbol output to the structured dump by `mrbc` does not output the ID directly, but outputs a macro such as `MRB_SYM()`. I think this will keep the output constant regardless of presym state at the time of `mrbc` build.
3. Make presym parsing from C source more strict using C preprocessor. I feel that all static symbols are output to C source by 2, so I think that all symbols can be extracted without parsing Ruby files.

shuujii:
@matz Could you please show me the build configuration?

shuujii:
@matz It seems that it does not depend on build configuration and it may be reproduced in CI environment, so I will investigate it.

shuujii:
BTW, in the current `cross-32bit.rb`, the cross environment is for 32-bit, but it looks like files output by `mrbc` created in the 64-bit host environment are compiled and linked. Will this work fine?

I feel that at least some compile options had to be unified in the host and cross environments.

matz:
The generated binary should be independent of the host architecture. If not, it must be a bug.

shuujii:
For example, the example below does not seem to work fine, although it is not an architectural difference but a compile option difference.

```ruby
# cross.rb
MRuby::Lockfile.disable

MRuby::Build.new do |conf|
  conf.toolchain :gcc
  conf.enable_debug
end

MRuby::CrossBuild.new("cross") do |conf|
  conf.toolchain :gcc
  conf.cc.defines << "MRB_UTF8_STRING"
  conf.enable_debug
  conf.enable_test
end
```

```console
$ rake CONFIG=cross.rb -m
$ build/cross/bin/mrbtest

.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F.........................................................................................
Fail: String#inspect [15.2.10.5.46] (core)
 - Assertion[3]
    Expected: "\"\\xe3\\x82\\x8b\""
      Actual: "\"る\""
  Total: 659
     OK: 658
     KO: 1
  Crash: 0
Warning: 0
   Skip: 0
```

shuujii:
Well, the expected value is correct, so it's not an issue of binary representation.

https://github.com/mruby/mruby/blob/6059834244a53241ce6591cd6248497b0c495563/test/t/string.rb#L771-L775

When compiling Ruby files, it is not `UTF8STRING`, so it seems that the latter is used.

I think it is better to unify compile options in the host and cross environments because such an issue is likely to occur.

dearblue:
> I will investigate at build errors later.
> 
> I would like to do the following for presym related, and I hope that if these are realized, most of the problems will be solved.
> 
>  1. Make it buildable without presym.
>
>  2. Symbol output to the structured dump by `mrbc` does not output the ID directly, but outputs a macro such as `MRB_SYM()`. I think this will keep the output constant regardless of presym state at the time of `mrbc` build.
>
>  3. Make presym parsing from C source more strict using C preprocessor. I feel that all static symbols are output to C source by 2, so I think that all symbols can be extracted without parsing Ruby files.

It seems that (2) and (3) can be realized immediately. Especially, I think (3) is a great idea.

However, I am concerned about (1).
At this point, `Class # new` is premised on" presym ".
https://github.com/mruby/mruby/blob/6d07d9b3d7a64834bda8644ab6c4ed1fabe217a4/src/class.c#L2739-L2744
Also, if static method tables are implemented in the future, I think this will also assume "presym".
If (1) is realized, what will happen to `static const mrb_sym mid = ...`?

shuujii:
> If (1) is realized, what will happen to `static const mrb_sym mid = ...`?

Without presym, I think we can initialize it in `mrb_init_class`.

dearblue:
:thinking: Hmm,
  - Presym (`static const mrb_sym`) cannot be used in static method table definition
  - If it wants to use presym for static method table definition as well, it requires code redundancy with `#if ~ #else ~ #endif`

I could only think of that...

PR #5202, opened 2020-11-29
Allow build without presym

shuujii:
### Usage

* You can build without presym by `MRuby::Build#disable_presym`.
* Macros such as `MRB_SYM` are expanded to `mrb_intern_lit` if presym is
  disabled, therefore the mruby state variable is required. The macros can
  be used when the variable name is `mrb`. If you want to use other variable
  names, you need to use macros with `_2` suffix, such as `MRB_SYM_2`.

### Note

Ruby files will be dumped in mrb format if presym is disabled. C struct
format may also be available if symbol array is initialized in init
function, but it is not supported.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -94,7 +94,9 @@ end
 
 desc "preallocated symbols"
 task :gensym do
-  MRuby.each_target{|build| Rake::Task[build.presym_file].invoke}
+  MRuby.each_target do |build|
+    Rake::Task[build.presym_file].invoke if build.presym_enabled?
+  end
 end
 
 depfiles += MRuby.targets.map { |n, t|
--- a/include/mruby/presym.h
+++ b/include/mruby/presym.h
@@ -7,42 +7,127 @@
 #ifndef MRUBY_PRESYM_H
 #define MRUBY_PRESYM_H
 
-#undef MRB_PRESYM_MAX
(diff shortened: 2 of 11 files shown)


shuujii:
This change causes incompatibilities that in some cases it will not work unless `MRB_SYM_2` etc. is used instead of` MRB_SYM` etc. Therefore, I would like this change to be included in the release version of mruby3. I also think it is necessary as a way to easily avoid issues related to presym.

An issue related to current presym, for example, is that the state of presym must match at the time of compiling `mrbc` and at the time of compiling the C struct dumped file by` mrbc`.

This can be an issue, especially in cross-compilation. In fact, the current `cross-32bit.rb` doesn't work.

```console
$ rake CONFIG=cross-32bit -m
$ build/cross-32bit/bin/mruby -e ''
stack level too deep (SystemStackError)
```

To work around this issue, gems must match in the host and cross environments, including external gems (now scans C source even with external gems due to aae82373). However, some gems may only be built in a cross environment. There may be a way to create `presym` file manually, but it seems to be hard.

shuujii:
The above cross-compilation example is due in part to the fact that `presym` files have been separated for each build target at #5194. Should #5194 be reverted once?

matz:
Since (part of) #5194 is reverted, we need to retry this PR. I close this at the moment.

PR #5218, opened 2020-12-06
Suppress `-Wtautological-compare` warning

shuujii:
```console
/mruby/src/debug.c:69:18: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
  if (irep && pc >= 0 && pc < irep->ilen) {
              ~~ ^  ~
```

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -66,7 +66,7 @@ mrb_debug_get_filename(mrb_state *mrb, const mrb_irep *irep, ptrdiff_t pc)
 MRB_API int32_t
 mrb_debug_get_line(mrb_state *mrb, const mrb_irep *irep, size_t pc)
 {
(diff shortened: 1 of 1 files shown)


shuujii:
BTW, I also feel that it is not good that the type of `pc` are different in the following functions.

* `get_file(mrb_irep_debug_info *info, uint32_t pc)`
* `mrb_debug_get_filename(mrb_state *mrb, const mrb_irep *irep, ptrdiff_t pc)`
* `mrb_debug_get_line(mrb_state *mrb, const mrb_irep *irep, size_t pc)`

matz:
@shuujii Indeed. All of them made to be `uint32_t` by ed29d74.

shuujii:
The warning doesn't seem to suppress.

matz:
Really. Let me see.

Issue #5151, opened 2020-11-20
`presym` files under gem directories


shuujii:
Currently, `presym` file is only output to `build` directory. However, `psfiles` in `Rakefile` also includes `presym` files under gem directories.

https://github.com/mruby/mruby/blob/5dc9a8249fc92de089627cad17b5d5b7a5f0a505/Rakefile#L104

Are these `presym` files supposed to be created manually, for example for dynamically interned symbols?

matz:
It is supposed to be created manually. It's to specify symbols that cannot be found by the scanner in `Rakefile`. If symbol scanning using `mrbc`, those `presym` files will be obsolete.

PR #5220, opened 2020-12-13
Improve source scanning for presym

shuujii:
The accuracy is greatly improved by using the C preprocessor to scan C
sources for presym. C preprocessor can perfectly interpret all comments and
preprocessor directives, so it can detect all symbols defined, for example
`mrbgems/mruby-socket/src/const.cstub`.

Also, as described later, this change will greatly improve the accuracy of
presym detection from Ruby sources.

## Result

The number of lines in the `presym` file for all gems is as follows:

  ```console
  Previous:   999 (false positive = 89, undetected = 297)
  New:       1207
  ```

## Build process

The new build process (with presym) is as follows:

1. Build `mrbc` without presym (more on building without presym later).
2. Compile Ruby sources to C struct format with the `mrbc` created in
   step 1, and create` mrblib.c` and `gem_init.c`. Note that the symbols
   in the created files are output as `MRB_SYM` family macros or
   `mrb_intern_lit` instead of IDs (details will be described later).
3. C preprocessor processes C sources including the created files of
   step 2 and outputs them as `.i` files. In these files, for example,
   `MRB_IVSYM(foo)` is converted to `<@! "@" "foo" !@>` and
   `mrb_define_module(mrb, "Foo")` is converted to `<@! "Foo" !@>`.
4. Scan the files created in step 3 and create `presym` and` presym.inc`
   files.

The files created in step 2 should output all static symbols defined in Ruby
sources, including local variables, so we can detect all presyms by just
scanning C sources without scanning Ruby sources directly.

Further, by this process, the files to be scanned becomes the same as the
files to be compiled, so that there is no excess or deficiency.

## Related changes

The following changes have been made in relation to realizing this feature.

### Allow build without presym

It enables build without presym to achieve the "Build process: 1". This
incorporates #5202, see its issue for details.

Note that when presym is enabled, even adding a local variable to a Ruby
source may change contents of presym and require recompilation of almost
all C sources. This is inconvenient, especially during trial and error in
development, but this feature is also useful because it does not cause
this problem if presym is disabled.

### Automatically create build target for `mrbc` without presym

The `mrbc` used in the "Build process: 1" will be built by automatically
creating a build target for it. The build name is `SOURCE_BUILD_NAME/mrbc`.

### Constantize output of C struct format by `mrbc`

To realizing the "Build process: 2", as mentioned above, symbol IDs are not
output directly in C struct format output by `mrbc`. As a result, the output
becomes constant regardless of the state of presym at the time of `mrbc`
build, and it is possible to detect symbols of Ruby sources in the same way
as other C sources.

Note that `mrb_intern_lit` is used for symbols that do not become presym,
but in this state, the corresponding element in the symbol array cannot be
statically initialized, so it is initialized at run time (therefore, in this
case, the `const` qualifier is not added to the symbol array).

### Specify arbitrary `mrbc` file

To realizing the "Build process: 2", enabled to specify `mrbc` created by
another build target or pre-built` mrbc`. Use `MRuby::Build#mrbcfile =` to
specify it explicitly. You can omit the "Build process: 1" by specifying
pre-built `mrbc`, and you can always use an optimized build to compile Ruby
sources faster. I think changes that affect the output of `mrbc` are rare,
so in many cases it helps to improve efficiency.

With presym, the build will be a little slower due to more build steps, but
this feature will improve it a bit.

### Create presym files for each build target

This feature was proposed at #5194 and merged once, but was reverted in
5c205e6e due to problems especially with cross-compilation. It has been
introduced again because this change solves the problem.

The presym files will be created below.

* `build/NAME/presym`
* `build/NAME/include/mruby/presym.inc`

### Other changes

* Because presym detection accuracy is greatly improved as mentioned above,
  `MRuby::Gem::Specification#cdump?` is set to true by default, and
  `disable_cdump` is added instead of `enable_cdump`. Also, support for gem
  specific presym files has been discontinued (https://github.com/mruby/mruby/issues/5151#issuecomment-730967232).
* Previously, `mrbc` was automatically created for the `host` build, but it
  will not be created if the build target for `mrbc` mentioned above is
  automatically created. At this time, `mrbc` file of the `mrbc` build is
  copied to` bin/`.
* Two types of `.d` files will be created, `.o.d` and `.i.d`. oThis is
  because if `.i` depends on `presym.inc`, the dependency will circulate, so
  the `.d` file cannot be shared.
* Changed file created with `enable_cxx_exception` to `X-cxx.cxx` from
  `X.cxx` to use the mruby standard Rake rule.

### Note

Almost all C sources will need to be recompiled if there are any changes to
`persym.inc` (if not recompiled properly, it will often result in run-time
error). If `gcc` toolchain is used, dependencies are resolved by the `.d`
file, so it become automatically recompile target, but if not (e.g. MSVC),
it is necessary to manually make it recompile target.

Also, even if `gcc` toolchain is used, it may not become recompile target if
external gems does not use the mruby standard Rake rule. In particular, if
the standard rule is overwritten, such as
https://github.com/mruby/mruby/pull/5112/files, `.d` file will not be read,
so be careful.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -25,12 +25,11 @@ end
 # load custom rules
 load "#{MRUBY_ROOT}/tasks/core.rake"
 load "#{MRUBY_ROOT}/tasks/mrblib.rake"
-
 load "#{MRUBY_ROOT}/tasks/mrbgems.rake"
 load "#{MRUBY_ROOT}/tasks/libmruby.rake"
-
+load "#{MRUBY_ROOT}/tasks/bin.rake"
+load "#{MRUBY_ROOT}/tasks/presym.rake"
 load "#{MRUBY_ROOT}/tasks/benchmark.rake"
-
 load "#{MRUBY_ROOT}/tasks/gitlab.rake"
 load "#{MRUBY_ROOT}/tasks/doc.rake"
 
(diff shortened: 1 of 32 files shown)


shuujii:
All tests with presym for the following gems reported so far succeed, except for the dependency `appPlant/mruby-process`.

* iij/mruby-digest
* matsumotory/mruby-mrbgem-template
* matsumotory/mruby-simplehttpserver
* mattn/mruby-http
* mattn/mruby-onig-regexp

`appPlant/mruby-process` test fails even without this PR, and the cause is probably on the gem side.

shuujii:
> Note that `mrb_intern_lit` is used for symbols that do not become presym,
> but in this state, the corresponding element in the symbol array cannot be
> statically initialized, so it is initialized at run time (therefore, in this
> case, the `const` qualifier is not added to the symbol array).

This situation is usually only for global variables and special variables such as `$0`, so adding corresponding` MRB_SYM` family macros will almost avoid it.

Even though these variables are deprecated, they are still used in the core gems for compatibility with Ruby, so it may be good to support them.

matz:
I am reviewing this pull-request. It takes a little more time.

matz:
After applying the patch, we have to add `-I <path-to-build-dir>/include` to the compiler options when we compile a program linking to `libmruby.a`. Is there any way to work around it?

shuujii:
Certainly that is incompatible. But I'm sorry, I can't think of any good ideas.

`mruby-config --cflags` prints include paths, so compatibility may not be broken if this output is used, but I feel like it's not being used very often.

shuujii:
If we copy `presym.inc` in the `host` build to `include/mruby/presym.inc` when building mruby, we may be able to build without adding to include path in most cases.

However, when linking `libmruby.a` built other than `host`, `build/NAME/include` must be prioritized over `include/`. Otherwise, we may get a runtime error instead of a build error, which may obscure the cause of the error.

matz:
CFuzz test fails for exactly this reason, FYI.

shuujii:
Existing code may be able to build if `presym.inc` is not included in the existing header files. I'll try to fix it with this solution.

dearblue:
There are two cross-build issues.

(1) If the `host` build target does not exist, it seems that` mrbc` cannot be built.

```console
% rake30 MRUBY_CONFIG=cross-32bit
rake aborted!
Don't know how to build task '/var/tmp/mruby/build/host/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.i' (See the list of available tasks with `rake --tasks`)
Did you mean?  /var/tmp/mruby/build/host/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.o
               /var/tmp/mruby/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c
               /var/tmp/mruby/build/cross-32bit/mrbgems/mruby-bin-mruby/tools/mruby/mruby.o
/usr/local/bin/rake30:23:in `load'
/usr/local/bin/rake30:23:in `<main>'
Tasks: TOP => default => all => gensym => /var/tmp/mruby/build/cross-32bit/presym
```

dearblue:
(2) When I add `mruby-bin-mrbc` to the cross build target, it seems that the` <cross-build-dir> /include/mruby/presym.inc` file is not generated.

```console
% cat cross1.rb
MRuby::CrossBuild.new("cross") do |conf|
  toolchain "clang"
  gem core: "mruby-bin-mrbc"
end

% rake30 MRUBY_CONFIG=cross1.rb `pwd`/build/cross/include/mruby/presym.inc
rake aborted!
Don't know how to build task '/var/tmp/mruby/build/cross/include/mruby/presym.inc' (See the list of available tasks with `rake --tasks`)
Did you mean?  /var/tmp/mruby/build/cross/src/numeric.o
/usr/local/bin/rake30:23:in `load'
/usr/local/bin/rake30:23:in `<main>'
```

As a result, the build will fail.

```console
% rake30 -v MRUBY_CONFIG=cross1.rb
mkdir -p /var/tmp/mruby/build/cross/mrbgems/mruby-bin-mrbc/tools/mrbc
CC    mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c -> build/cross/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.o
clang -c -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -Wzero-length-array -DMRBGEM_MRUBY_BIN_MRBC_VERSION=0.0.0 -I"/var/tmp/mruby/include" -I"/var/tmp/mruby/build/cross/include" -MMD -MF "/var/tmp/mruby/build/cross/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.o.d" -o "/var/tmp/mruby/build/cross/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.o" "/var/tmp/mruby/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c"
In file included from /var/tmp/mruby/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.c:1:
In file included from /var/tmp/mruby/include/mruby.h:1412:
In file included from /var/tmp/mruby/include/mruby/presym.h:15:
/var/tmp/mruby/include/mruby/presym/enable.h:19:11: fatal error: 'mruby/presym.inc' file not found
# include <mruby/presym.inc>
          ^~~~~~~~~~~~~~~~~~
1 error generated.
rake aborted!
Command failed with status (1): [clang -c -std=gnu99 -g -O3 -Wall -Wundef -...]
/var/tmp/mruby/lib/mruby/build/command.rb:37:in `_run'
/var/tmp/mruby/lib/mruby/build/command.rb:99:in `run'
/var/tmp/mruby/lib/mruby/build/command.rb:120:in `block (2 levels) in define_rules'
/usr/local/bin/rake30:23:in `load'
/usr/local/bin/rake30:23:in `<main>'
Tasks: TOP => default => all => gensym => /var/tmp/mruby/build/cross/presym => /var/tmp/mruby/build/cross/mrblib/mrblib.i => /var/tmp/mruby/build/cross/mrblib/mrblib.c => /var/tmp/mruby/build/cross/bin/mrbc => /var/tmp/mruby/build/cross/mrbgems/mruby-bin-mrbc/tools/mrbc/mrbc.o
```

shuujii:
@dearblue Thank you for the reports.  I will investigate.

PR #5224, opened 2020-12-15
feat(CI): add a GitHub Action to lint the YAML

jbampton:
- Run only on pull request
- Add a `.yamllint` config file
- Lint YAML

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,11 +8,11 @@ jobs:
     env:
       MRUBY_CONFIG: travis
     steps:
-    - uses: actions/checkout@v1
-    - name: apt
-      run: sudo apt install ruby gperf
-    - name: build and test
-      run: rake -m -j4 gensym all test
+      - uses: actions/checkout@v1
+      - name: apt
+        run: sudo apt install ruby gperf
+      - name: build and test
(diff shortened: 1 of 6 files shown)


matz:
Thank you for running Lint for YAML files. But do we really need to run YAML lint on every pull-request?
Considering YAML files are not change often, I am afraid it's too much.

jbampton:
Hey @matz I think it's good to have the YAML lint check on the CI to stop regressions.

Plus so far it takes less than 10 seconds to run.

shuujii:
`.gitlab-ci.yml` is generated by `rake gitlab_config` using `YAML.dump`. So if you want to change the YAML format, you need to modify `tasks/gitlab.rake`. However can it be output in the following format?

https://github.com/mruby/mruby/blob/fbe5033f80cda70d4e835b7c99748df116d3be4c/.gitlab-ci.yml#L2-L4

By the way, currently `rake gitlab_config` gives an error. I feel that no one is using GitLab CI, so I personally think that it is okay to delete the related part.

jbampton:
You can always ignore files from YAML linting as seen here:

https://yamllint.readthedocs.io/en/stable/configuration.html#ignoring-paths

Just needs an edit to the `.yamllint` config file.

PR #5236, opened 2020-12-20
Reorganize GitHub Actions configuration [skip travis][skip appveyor]

shuujii:
* Skip if commit message contains `ci skip`, `skip ci`, or `skip gha`
  ENCLOSED WITH BRACKETS (excluding lint job).
* Separate build and test. This is because builds can be run in parallel,
  but running tests in parallel can cause logs to get mixed up or not
  finished.
* Don't use Chocolatey because it seems to take 1-2 minutes to start up.
* Use the cache better.
* Use `actions/checkout@v2` instead of `actions/checkout@v1`.
* Remove unnecessary package installation.
* Remove unnecessary flag settings.
* Remove `-j` for rake because it doesn't seem to have any effect.
* Rename `main.yml` to `oss-fuzz.yml` to clarify.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,105 +3,151 @@ name: Build & Test
 on: [push, pull_request]
 
 jobs:
+  Check-Skip:
+    if: |
+      !contains(github.event.head_commit.message, '[ci skip]') &&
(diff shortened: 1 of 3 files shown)


shuujii:
The skip feature seems to be effective when push, but not when pull-request.

PR #5237, opened 2020-12-20
Use `git ls-files` instead of `find` in `spell-checker.yml` [ci skip]

shuujii:
To avoid including `.git` directory.


shuujii:
BTW, IMHO, it may be better not to use emoji because it may not be read and written well depending on the environment.

PR #5241, opened 2020-12-21
Take advantage of gembox

dearblue:
I think that it is easy to use if it is divided according to the type of library and the general purpose.
It is a subdivision from the previous `default.gembox`.

By type gembox:
- `stdlib`: Add some standard Ruby methods not provided by core.
- `stdlib-ext`: Add some standard Ruby methods that are not provided by `stdlib`.
- `stdlib-io`: Provides `IO`, `File`, `Socket`, `print`, etc. Conflict with `MRB_NO_STDIO`
- `math`: Add a class related to math. Conflict with `MRB_NO_FLOAT`
- `metaprog`: Adds features related to metaprogramming.

Purpose gembox:
- `default.gembox`: Import the same gems as before
- `default-no-stdio.gembox`: Import the` stdlib` and `math`
- `default-no-fpu.gembox`: Import the` stdlib` only

---

I think there is room for debate about the gembox name and classification.

### DIFF
--- a/mrbgems/default-no-fpu.gembox
+++ b/mrbgems/default-no-fpu.gembox
@@ -1,0 +1,3 @@
+MRuby::GemBox.new do |conf|
+  gembox "stdlib"
+end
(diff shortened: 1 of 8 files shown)


matz:
Done.

matz:
Oops, I made a mistake. Reopen.

PR #5243, opened 2020-12-23
Add example for cross-compiling to Nintendo GameBoyAdvance

boatrite:
Hey everyone!

I've gotten mruby running on a Nintendo GameBoyAdvance and wanted to contribute my build config to the project.

It's pretty simple. I modeled it after the Dreamcast one, but in this case I'm using the the devkitPro toolchain to do the cross-compiling.

The biggest drawback here is that I've only tested it on Linux. I don't have access to a Mac and would have no idea how to even start going about doing any development on a Windows machine, but since the build configs here seem to mostly act as examples, I think there's still plenty of value in adding it as-is.

Happy Holidays!


matz:
Thank you!

PR #5246, opened 2020-12-27
Add support for squiggly heredocs

SeekingMeaning:
Hello, I have added support for [squiggly heredocs](https://ruby-doc.org/core-2.7.2/doc/syntax/literals_rdoc.html#label-Here+Documents+-28heredocs-29) to mruby. Please tell me if I did anything wrong! (Should I commit `y.tab.c`?)

PR #5247, opened 2020-12-27
Fix for empty lines in squiggly heredocs

SeekingMeaning:
Sorry, I have made a mistake in my previous PR https://github.com/mruby/mruby/pull/5246 and forgot to check for empty lines in squiggly heredocs.

### DIFF
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -1362,26 +1362,29 @@ heredoc_end(parser_state *p)
         node *pair = list->car->cdr;
         const char *str = (char*)pair->car;
         size_t len = (size_t)pair->cdr;
-        if (counting) {
-          list2 = push(list2, pair);
-        }
+        mrb_bool check = counting;
+        mrb_bool empty = TRUE;
+        mrb_bool newline = FALSE;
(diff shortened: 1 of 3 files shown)


shuujii:
The following result seems to be different from Ruby.

```ruby
# example.rb
p <<~EOS
\t
 a
EOS
```

```console
$ ruby example.rb
"\t\n a\n"
```

```console
$ mruby example.rb
"\na\n"
```

shuujii:
BTW, in Ruby, if the tab is a hard tab (not `\t`), the result seems to change and I'm not sure if it is intentional or not.

https://wandbox.org/permlink/po9PrAiZKOOkZWnn

SeekingMeaning:
Thank you for bringing this up @shuujii, I shall look into this further

SeekingMeaning:
The following input gives me an unexpected output:

```
<<~EOS
⇥This line is prefixed by a tab character.
··⇥This line is prefixed by two spaces and a tab character.
⇥··This line is prefixed by a tab character and two spaces.
EOS
```

Output:

```
This line is prefixed by a tab character.
This line is prefixed by two spaces and a tab character.
··This line is prefixed by a tab character and two spaces.
```

PR #5252, opened 2020-12-31
Removed unusable `GC.test`

dearblue:
The substance of the method was removed in commit 15ceb35e058a078f632a1ff7d0d424c59d48cd80.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -1614,11 +1614,6 @@ mrb_objspace_page_slot_size(void)
   return sizeof(RVALUE);
 }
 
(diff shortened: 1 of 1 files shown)


dearblue:
Maybe this can also be deleted?
https://github.com/mruby/mruby/blob/aabfe31d98794b2af238735e66b0ca31d195e8e7/src/gc.c#L185-L189
https://github.com/mruby/mruby/blob/aabfe31d98794b2af238735e66b0ca31d195e8e7/src/gc.c#L795

PR #5255, opened 2021-01-03
Avoid 64-bit operations in `src/hash.c`; close #5201

### LINKED ISSUE #5201: crash at class instance value access
I use mruby on realtek mips soc by bare metal.

I found problem at 2e30c68562e6c653191a57fa396e0591b878936f.

That is crash on modified mruby-simplehttp class  at line of @uri[:path] = path in request method.

https://github.com/yamori813/mruby-simplehttp

I do workaround for this problem by not use @uri[:path].

I don't have crash but @uri[:port] is half value. I send double value in script then work is fine.

shuujii:
The idea of using `size_t` in `ea_next_capa_for` is by @dearblue.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -251,7 +251,6 @@ HT_ASSERT_SAFE_READ(ea_capa);
 } while (0)
 
 #define U32(v) ((uint32_t)(v))
-#define U64(v) ((uint64_t)(v))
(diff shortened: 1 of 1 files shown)


matz:
LGTM

matz:
@yamori813 This PR seems to fix #5201. Can you confirm?

PR #5277, opened 2021-01-11
Avoid including `presym.inc` in existing header files

shuujii:
Addressed an issue where existing programs linking `libmruby.a` could only
be built by adding `<build-dir>/include` to compiler's include path.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1303,26 +1303,26 @@ MRB_API mrb_value mrb_vformat(mrb_state *mrb, const char *format, va_list ap);
    + those E_* macros requires mrb_state* variable named mrb.
    + exception objects obtained from those macros are local to mrb
 */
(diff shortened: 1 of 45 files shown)


shuujii:
I think [this issue](https://github.com/mruby/mruby/pull/5220#issuecomment-756724600) has been fixed, but CIFuzz still fails randomly.

CIFuzz seems to build mruby with `make clean all -j2`, so `clean` and `all` run in parallel, and `clean` may be executed after `all` starts.

I don't think `clean` is needed in the first place, so I'll create a pull request to https://github.com/google/oss-fuzz.

shuujii:
Is this way not good?

shuujii:
Thank you for merging. The pull request to oss-fuzz have also been merged (
https://github.com/google/oss-fuzz/commit/8f99fa0170fcd29a5f66a21290f44cc308cab70e).

PR #5290, opened 2021-01-15
Avoid syntax error when EXPR_MID appears in conditional operator

hasumikin:
Unexpected syntax error happens in a code for example,

```
mruby -e'true ? return : false'
-e:1:15: syntax error, unexpected tSYMBEG, expecting tLABEL_TAG or ':'
```

As long as I examined, it seems that the error raises when EXPR_MID
token (return, next, break and rescue) appears in the second clause of
conditional operator sentence.

```
true ? return : false
       ^^^^^^ here
```

I'm afraid my correction might be too simple to solve the problem
though, very glad if you check it.

### DIFF
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -5882,7 +5882,7 @@ parser_yylex(parser_state *p)
       p->lstate = EXPR_BEG;
       return tLABEL_TAG;
     }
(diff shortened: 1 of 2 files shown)


matz:
Thank you for the report. I fixed it in a different way to address the issue like the following:

```
mruby -e 'true ? return :# false
false'
```

hasumikin:
@matz I've learned one thing, thank you!!

PR #5272, opened 2021-01-10
Reorganize mrb_callinfo

dearblue:
Overview

  - `mrb_callinfo::stack` will now indicate the current call level.
    Previously, `stackent` saved the previous stack.
  - `mrb_callinfo::pc` will now indicate the current call level.
    Previously, this saved the previous one.
    This change reverts commit a0c1e075e35c358d21934c28ff1bec4153502409.
  - ` mrb_callinfo::argc` and` mrb_callinfo::acc` are represented by `int16_t`.
  - Represents `mrb_callinfo::env` and `mrb_callinfo::target_class` with union.

PR #5291, opened 2021-01-16
Fixed stack position of return value; ref #5272

dearblue:
When I `#call` the "proc" object created by the `mrb_proc_new_cfunc()` function from Ruby space, the return value did not go into the correct stack position.
This can destroy the calling variable.

This issue is now caused by #5272. sorry.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1515,7 +1515,7 @@ RETRY_TRY_BLOCK:
         /* pop stackpos */
         ci = cipop(mrb);
         pc = ci->pc;
-        regs[ci->acc] = recv;
+        regs[ci[1].acc] = recv;
         irep = mrb->c->ci->proc->body.irep;
         pool = irep->pool;
         syms = irep->syms;


dearblue:
Demonstration code

```console
% cat opcall.c
#include <mruby.h>
#include <mruby/compile.h>
#include <mruby/variable.h>
#include <mruby/proc.h>

static mrb_value
cfunc(mrb_state *mrb, mrb_value self)
{
  return mrb_fixnum_value(99);
}

int
main(int argc, char *argv[])
{
  mrb_state *mrb = mrb_open();
  mrb_gv_set(mrb, mrb_intern_lit(mrb, "$x"), mrb_obj_value(mrb_proc_new_cfunc(mrb, cfunc)));
  mrb_p(mrb, mrb_load_string(mrb, "def vertex; x, y, z = 1, 2, 3; w = $x.call; [x, y, z, w]; end; a = 1; vertex"));
  mrb_close(mrb);
  return 0;
}

% cc `bin/mruby-config --cflags --ldflags` opcall.c `bin/mruby-config --libs` && ./a.out
[99, 2, 3, #<Proc:0x262130>]
```

What I expect at this time is `[1, 2, 3, 99]`.

PR #5298, opened 2021-01-26
Refactor `E_xxx_ERROR` macros; ref 2ddfd50df

shuujii:


include/mruby.h
```diff
@@ -1305,32 +1305,29 @@ MRB_API mrb_value mrb_vformat(mrb_state *mrb, const char *format, va_list ap);
 */
 #ifdef MRB_SYM
```
matz, line 1306: If `mruby/presym.h` is not included **before** `mruby.h`, `MRB_SYM` is not defined here, so `mrb_intern_lit()` version seems to be used everytime. Am I missing something?

include/mruby/presym/enable.h
```diff
@@ -38,8 +38,4 @@ enum mruby_presym {
   static const mrb_sym name[] = {__VA_ARGS__};                                \
   static void init_##name(mrb_state *mrb) {}
 
-/* use MRB_SYM() for E_RUNTIME_ERROR etc. */
-# undef MRB_E_SYM
-# define MRB_E_SYM(sym) MRB_SYM(sym)
```
matz, line 43: Even when `mruby/presym/enable.h` is included **after** `mruby.h` above lines do the trick (replace `MRB_E_SYM` with `MRB_SYM` version).

mrbgems/mruby-error/test/exception.c
```diff
@@ -1,6 +1,7 @@
 #include <mruby.h>
 #include <mruby/error.h>
 #include <mruby/array.h>
+#include <mruby/presym.h>
```
matz, line 4: The only difference between the two versions of `MRB_E_SYM()` is slight performance. So unless the source file uses `MRB_SYM()` in it (as described in `doc/guides/symbol.md`), I don't think we need to include `mruby/presym.h` (to avoid unnecessary recompilation).

shuujii:
Sorry at the lack of explanation. I tried to fix it for the following reasons.

* `MRB_E_SYM` is confusing with `MRB_SYM_E` thus I would like to avoid it.
* `MRB_E_SYM` is not used in the first place.

However, this PR was wrong because I misunderstood the behavior of `MRB_E_SYM`.

> If `mruby/presym.h` is not included **before** `mruby.h`, `MRB_SYM` is not defined here, so `mrb_intern_lit()` version seems to be used everytime.

Indeed. I didn't realize it because it was originally the case. Should we just change it to something like `#define MRB_E_SYM(sym) mrb_intern_lit (mrb, #sym)` here?

> The only difference between the two versions of `MRB_E_SYM()` is slight performance.

Yes. So at first I thought we didn't have to use `MRB_SYM` in `E_xxx_ERROR`.

> So unless the source file uses `MRB_SYM()` in it (as described in `doc/guides/symbol.md`), I don't think we need to include `mruby/presym.h` (to avoid unnecessary recompilation).

If we use `MRB_SYM` we need `mruby/presym.h`, and if we use `E_xxx_ERROR` `MRB_SYM` is also used, so I think we can't compile without including `mruby/presym.h`.

matz:
Shall I update this PR, or would you update by yourself (to avoid `MRB_E_SYM` and `MRB_SYM_E` confusion)?
I prefer using `MRB_SYM` if available.

shuujii:
I change `MRB_E_SYM` to` MRB_ERROR_SYM` and recreate the PR.

shuujii:
Sorry, `MRB_SYM` is not used if `mruby/presym.h` is not included, so it could be compiled. Therefor, don't add include of `mruby/presym.h`.

PR #5299, opened 2021-01-27
Fix "Use `MRB_SYM()` for error class retrieval"; ref 2ddfd50df

shuujii:
The new macro (`MRB_E_SYM`) was not being used, so it is being used. Also
`MRB_E_SYM` is confusing with `MRB_SYM_E`, so change it to `MRB_ERROR_SYM`.


matz:
Thank you!

PR #5303, opened 2021-01-27
Split `presym_table` for reduced program size

shuujii:
Because a structure that is an element of `presym_table` has padding, split
it into individual arrays for name and length.

#### Result (64-bit CPU with full-core gembox)

|        |   mruby    | libmruby.a |
|--------|------------|------------|
| Before | 1,087,444B | 1,476,872B |
| After  | 1,079,340B | 1,469,784B |

### DIFF
--- a/src/symbol.c
+++ b/src/symbol.c
@@ -15,31 +15,24 @@
 
 #ifndef MRB_NO_PRESYM
 
-# undef MRB_PRESYM_MAX
-# define MRB_PRESYM_NAMED(lit, num, type, name) {lit, sizeof(lit)-1},
-# define MRB_PRESYM_UNNAMED(lit, num) {lit, sizeof(lit)-1},
-
-static const struct {
-  const char *name;
-  uint16_t len;
-} presym_table[] = {
 #ifndef MRB_PRESYM_SCANNING
-# include <mruby/presym.inc>
+/* const uint16_t presym_length_table[]   */
+/* const char * const presym_name_table[] */
+# include <mruby/presym/table.h>
 #endif
-};
 
 static mrb_sym
 presym_find(const char *name, size_t len)
 {
-  if (presym_table[MRB_PRESYM_MAX-1].len < len) return 0;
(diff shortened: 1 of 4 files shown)


src/symbol.c
```diff
@@ -15,31 +15,24 @@
...
-
-static const struct {
-  const char *name;
-  uint16_t len;
-} presym_table[] = {
 #ifndef MRB_PRESYM_SCANNING
-# include <mruby/presym.inc>
+/* const uint16_t presym_length_table[]   */
+/* const char * const presym_name_table[] */
+# include <mruby/presym/table.h>
```
matz, line 21: I like the idea. How about redefining `MRB_PRESYM_NAMED` and `MRB_PRESYM_UNNAMED` and include `presym.inc` twice? E.g.

```c
const char * const presym_name_table[] = {
#undef MRB_PRESYM_MAX
#define MRB_PRESYM_NAMED(lit, num, type, name) lit,
#define MRB_PRESYM_UNNAMED(lit, num) lit,
#include <mruby/presym.inc>
};
const uint16_t presym_length_table[] = {
#undef MRB_PRESYM_MAX
#define MRB_PRESYM_NAMED(lit, num, type, name) sizeof(lit)-1,
#define MRB_PRESYM_UNNAMED(lit, num) sizeof(lit)-1,
#include <mruby/presym.inc>
};
```
Pros
 * Changes are minimal and clear
 * Less complex build process

Cons

 *  Trickier C code
 *  `mruby/presym/id.h` and `mruby/presym/table.h` might be more readable.

shuujii:
IMHO, macro tricks are complex, obscure, and redundant, so I'd like to avoid them if possible. Also, I think build process is about the same complexity because it just splits what is generated as `presym.inc` into` id.h` and `table.h`.

shuujii:
BTW, I forgot to add `static` to the variables.

shuujii:
Thank you for the merge and a some fixes!

PR #5306, opened 2021-01-28
Fixed `String#unpack` to handle the highest range of integer values

dearblue:
Previously, problems occurred when the `fixnum` was exceeded.

- 32-bit cpu mode with `MRB_WORD_BOXING` and `MRB_INT32`:

  ```console
  % bin/mruby -e 'p [0x7fffffff].pack("N").unpack("N")'
  -e:1: cannot unpack to Integer: 2147483647 (RangeError)
  ```

- 64-bit cpu mode with `MRB_WORD_BOXING` and `MRB_INT64`:

  ```console
  % bin/mruby -e 'p [0x7fffffff_ffffffff].pack("q").unpack("q")'
  -e:1: cannot unpack to Integer: 9223372036854775807 (RangeError)
  ```

### DIFF
--- a/mrbgems/mruby-pack/src/pack.c
+++ b/mrbgems/mruby-pack/src/pack.c
@@ -216,26 +216,6 @@ u32tostr(char *buf, size_t len, uint32_t n)
   snprintf(buf, len, "%" PRIu32, n);
 #endif /* MRB_NO_STDIO */
 }
-
-static void
-i32tostr(char *buf, size_t len, int32_t n)
-{
-#ifdef MRB_NO_STDIO
-  if (len < 1) {
-    return;
-  }
-
-  if (n < 0) {
-    *buf ++ = '-';
-    len --;
-    n = -n;
-  }
-
-  u32tostr(buf, len, (uint32_t)n);
-#else
-  snprintf(buf, len, "%" PRId32, n);
-#endif /* MRB_NO_STDIO */
-}
 #endif /* MRB_INT64 */
 
 static int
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-pack/src/pack.c
```diff
@@ -258,22 +261,22 @@ unpack_l(mrb_state *mrb, const unsigned char *src, int srclen, mrb_value ary, un
   if (flags & PACK_FLAG_SIGNED) {
     int32_t sl = ul;
 #ifndef MRB_INT64
-    if (!FIXABLE(sl)) {
+    if (INT_OVERFLOW_P(sl)) {
```
matz, line 264: I admit `FIXABLE` is no longer a sufficient check. In the past, `fixnum` integers need to be fit in 31bits under word-boxing mode. But I fixed it recently.

But I have noticed that we are in `ifndef MRB_INT64` (`mrb_int` is `int32_t`) and `sl` is `int32_t` so that `INT_OVERFLOW_P` is always false. Probably we have to check if `sl` is negative (too big unsigned integer converted to signed) or call `UINT_OVERFLOW_P()` with `ul`.

dearblue: Thank you for pointing out.
When I think about it again, confirmation is not necessary for `int32_t` (Because it is processing unsigned integers to signed integers for convenience of processing).

In summary:

|                |  `MRB_INT32`  |  `MRB_INT64`  |
| -------------: | ------------: | ------------: |
|   signed int32 | no-need check | no-need check |
| unsigned int32 |    need check | no-need check |
|   signed int64 |    need check | no-need check |
| unsigned int64 |    need check |    need check |

I'm sorry, I'll fix it tomorrow (maybe)...

dearblue:
I corrected it.

PR #5317, opened 2021-02-03
Reduce memory usage of instance variable table

shuujii:
## Implementation Summary

* Only keys and only values of hash table are contiguous to eliminate
  structure padding.
* Change upper limit of `iv_tbl` size to `UINT16_MAX` (it seems to be
  acceptable in mruby because the total number of classes/modules
  immediately after starting Redmine is 20,000 or less).
* `iv_tbl*` point hash buckets directly.

## Benchmark Summary

Only the results of typical situations on 64-bit Word-boxing are present
here. For more detailed information, including consideration, see below
report (although most of the body is written in Japanese).

* https://shuujii.github.io/mruby-iv-benchmark

### Memory Usage

Lower value is better.

| iv_tbl Size |    Baseline    |       New      |   Factor   |
|------------:|---------------:|---------------:|-----------:|
|           4 |            88B |            52B |   0.59091x |
|          30 |           536B |           388B |   0.72388x |
|         100 |          2072B |          1540B |   0.74324x |
|         200 |          4120B |          3076B |   0.74660x |

Although not mentioned in the above report, the memory usage of `mrbtest`
(full-core gembox) is as follows in the result by Valgrind.

* Baseline: 108,086 allocs, 16,313,122 bytes allocated
* New:       94,273 allocs, 15,875,214 bytes allocated

### Performance

Higher value is better.

#### `mrb_obj_iv_set`

| iv_tbl Size |    Baseline    |       New      |   Factor   |
|------------:|---------------:|---------------:|-----------:|
|           4 |  88.63003M i/s |  92.60611M i/s |   1.04486x |
|          30 |  32.97066M i/s |  25.25095M i/s |   0.76586x |
|         100 |  16.33224M i/s |  22.74998M i/s |   1.39295x |
|         200 |   5.64484M i/s |   6.79949M i/s |   1.20455x |

#### `mrb_obj_iv_get`

| iv_tbl Size |    Baseline    |      New       |   Factor   |
|------------:|---------------:|---------------:|-----------:|
|           4 | 217.58391M i/s | 237.59912M i/s |   1.09199x |
|          30 | 139.56195M i/s | 160.49470M i/s |   1.14999x |
|         100 | 143.09716M i/s | 190.95047M i/s |   1.33441x |
|         200 |  89.75291M i/s | 134.78717M i/s |   1.50176x |

### Binary Size

Lower value is better.

|    File     |    Baseline    |      New       |   Factor   |
|:------------|---------------:|---------------:|-----------:|
| mruby       |       697,520B |       697,520B |   1.00000x |
| libmruby.a  |     1,046,570B |     1,046,682B |   0.99989x |

## Note

The address in `struct RObject::iv` may change after initialization because
`iv_tbl*` points directly to hash buckets. Therefore, the address cannot be
copied and shared when include/prepend. So, when sharing `iv_tbl`, refer to
it via the sharing source class. As a result, the following bug have also
been fixed.

* [An `iv_tbl` is not shared when a class includes or prepends an empty module](https://gist.github.com/shuujii/0ac23fa24b0c55b2c602b534d81e4a95)

### DIFF
--- a/src/variable.c
+++ b/src/variable.c
@@ -12,224 +12,315 @@
 #include <mruby/variable.h>
 #include <mruby/presym.h>
 
-struct iv_elem {
-  mrb_sym key;
-  mrb_value val;
-};
-
-/* Instance variable table structure */
+/*
+ * Instance variable table structure
+ *
+ * Layout:
+ *                                           iv_tbl* --.
+ *                                                      \
+ *           +---------+---------+---+-------+-------+---o--------+--------+
+ *   Type    |mrb_value|mrb_value|...|mrb_sym|mrb_sym|...|uint16_t|uint16_t|
+ *           +---------+---------+---+-------+-------+---+--------+--------+
+ *   Content |  val 1  |  val 2  |...| sym 1 | sym 2 |...|  size  | capa_1 |
+ *           +---------+---------+---+-------+-------+---+--------+--------+
+ *
+ *   - Only `val` and only `sym` are contiguous to eliminate structure padding.
+ *   - `val` is placed at the beginning to align it at 8-byte boundary.
+ *   - `capa_1` means "capacity - 1".
+ */
 typedef struct iv_tbl {
-  size_t size;
-  size_t alloc;
-  struct iv_elem *table;
+  char dummy1;
+  char dummy2[];
 } iv_tbl;
 
-/* Creates the instance variable table. */
-static iv_tbl*
-iv_new(mrb_state *mrb)
+typedef struct iv_tbl_iter {
+  mrb_sym *syms;
+  mrb_value *vals;
+  uint16_t idx;
+  uint16_t mask;
+} iv_tbl_iter;
+
+#define IV_INIT_CAPA 1
+#define IV_MAX_CAPA (U32(UINT16_MAX) + 1)
+#define IV_MAX_SIZE UINT16_MAX
+#define IV_EMPTY_SYM UINT32_MAX
+#define IV_DELETED_SYM (IV_EMPTY_SYM - 1)
+
+#define U16(v) ((uint16_t)(v))
+#define U16P(p) ((uint16_t*)(p))
+#define U32(v) ((uint32_t)(v))
+
+#define sym_hash_code(sym) U32((sym) ^ ((sym) << 2) ^ ((sym) >> 2))
+#define iv_size_ptr(t) U16P(t)
+#define iv_capa_1_ptr(t) U16P((t) + sizeof(uint16_t))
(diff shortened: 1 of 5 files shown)


matz:
I like the idea. But I need time to review it. Changes in `variable.c` is a little bit complex.

src/variable.c
```diff
@@ -12,224 +12,315 @@
...
+ *   - Only `val` and only `sym` are contiguous to eliminate structure padding.
+ *   - `val` is placed at the beginning to align it at 8-byte boundary.
+ *   - `capa_1` means "capacity - 1".
+ */
 typedef struct iv_tbl {
-  size_t size;
-  size_t alloc;
-  struct iv_elem *table;
+  char dummy1;
+  char dummy2[];
```
matz, line 33: I am not sure why we should do this kind of tricky memory allocation. Naive `struct` definition like:

```c
struct iv_tbl {
  uint16_t size;
  unit16_t capa_1;
};
```

and locating `values` and `syms` **after** `iv_tbl` seem to work. So we can access `size` and `capa_1` directly without tricky typecasting. Is there any reason?

shuujii: There are two reasons:

* To align `val` to 8-byte boundary (we had [a similar problem](https://github.com/mruby/mruby/pull/1353#issuecomment-20719140) in khash). I don't think `val` can aligned to 8-byte boundary if `size` and` capa_1` are placed at the beginning.
* It is easy to calculate address if `sizeof(iv_tbl)` is 1.

dearblue: @shuujii,
If the structure is the one shown by @matz, then `val` will be placed on an 8-byte boundary.
If you are concerned about padding and want to keep this PR layout, you should be fine with a structure that defines only `size` and `capa_1`.

```c
struct iv_tbl {
  /* mrb_value val[]; */
  /* mrb_sym sym[]; */
  uint16_t size;
  uint16_capa_1;
};
```

Example of an actual change (diff):
https://github.com/shuujii/mruby/compare/reduce-memory-usage-of-instance-variable-table...dearblue:shuujii/reduce-memory-usage-of-instance-variable-table+

shuujii: @dearblue, thank you for your comment.

> If the structure is the one shown by @matz, then `val` will be placed on an 8-byte boundary.

Ah, sorry, I misread the layout. `size -> capa_1 -> syms -> vals` has a boundary problem, but `size -> capa_1 -> vals -> syms` has no problem. However, the layout make 4-byte padding before `vals` on 64-bit CPU, so I would like to avoid it.

> If you are concerned about padding and want to keep this PR layout, you should be fine with a structure that defines only `size` and `capa_1`.

I agree with your example. Thank you!

PR #5320, opened 2021-02-05
Define `MRB_NO_GEMS` if `disable_libmruby` is specified

shuujii:


matz:
Merged, resolving conflict with #5387.

PR #5321, opened 2021-02-05
Fix name, super class and message for error class used in `mruby-catch`

shuujii:
### Example

```ruby
begin
  throw 1
rescue Exception => e
  puts "#{e.class}: #{e.message}"
  puts e.class.superclass
end
```

#### Before this patch:

```console
$ bin/mruby example.rb
ThrowCatchJump: uncaught throw :1
Exception
```

```console
$ bin/mruby example.rb
UncaughtThrowError: uncaught throw 1
ArgumentError
```

### DIFF
--- a/mrbgems/mruby-catch/mrblib/catch.rb
+++ b/mrbgems/mruby-catch/mrblib/catch.rb
@@ -1,8 +1,8 @@
-class ThrowCatchJump < Exception
+class UncaughtThrowError < ArgumentError
   def initialize(tag, val)
     @tag = tag
     @val = val
-    super("uncaught throw :#{tag}")
+    super("uncaught throw #{tag.inspect}")
   end
   def _tag
     @tag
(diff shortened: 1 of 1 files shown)


dearblue:
The current situation is probably due to the implementation.

CRuby throws an exception from `throw` only when there is no corresponding` catch`.
So in CRuby, if you have a corresponding `catch`, you won't get caught in a`rescue` in the middle.

I think the reason mruby derives from `Exception` is that it doesn't get caught easily with` rescue => e`.

We can do the same with mruby by throwing a `RBreak` object, but before that, the challenge is how to register / unregister a symbol with` catch` and how to find it with `throw`.

matz:
As you guessed, it's an implementation limitation. In CRuby, global jumps from `throw` cannot be caught by `rescue`, for example:

```ruby
catch(:a) do
  begin
    throw :a
  rescue Exception => e
    p e
  end
end
```
does not print anything in CRuby, but mruby does. We implemented only one global jump using exceptions. So I don't think this PR is a good idea. It could cause more problems than its benefit. 

Options are:

* Leave this limitation as it is.
* Implement different global jump system (probably using `RBreak`)

I am not sure whether it's worth implementing it.

shuujii:
@dearblue and @matz, thank you for the explanation. I understood it well. Well then, I will close this PR and create PR that only fix the message.

PR #5327, opened 2021-02-06
Ignore if `conf.enable_test` does not exist

dearblue:
Because if the configuration file didn't contain any `conf.enable_test`, `rake test` would report an exception and exit.

```console
% cat my_config.rb
MRuby::Build.new { toolchain }

% rake MRUBY_CONFIG=my_config.rb test

  ...SNIP...

rake aborted!
NoMethodError: undefined method `invoke' for nil:NilClass
/var/tmp/mruby/tasks/test.rake:24:in `block (3 levels) in <top (required)>'
Tasks: TOP => test => test:build => test:build:lib
```

### DIFF
--- a/tasks/test.rake
+++ b/tasks/test.rake
@@ -21,7 +21,8 @@ namespace :test do |test_ns|
     desc "build library tests"
     task :lib => :all do
       MRuby.each_target{|build| build.gem(core: 'mruby-test')}
-      test_build_ns["lib_without_loading_gem"].invoke
+      test = test_build_ns["lib_without_loading_gem"]
+      test.invoke if test
     end
   end
 


dearblue:
I wanted to use the Bocchi operator from Ruby-2.3, but I couldn't use it due to compilation requirements...

[doc/guides/compile.md#prerequisites](https://github.com/mruby/mruby/blob/88113df0981142be989e868689bf549cb95f0e6c/doc/guides/compile.md#prerequisites)

matz:
Shall we upgrade to `2.5` or later?

dearblue:
I would be happy if you could do so.

shuujii:
@dearblue, Thank you so much for fixing many bugs in build system.

PR #5328, opened 2021-02-06
Reimplement mruby-catch; ref #5321

dearblue:
When there is a corresponding tag, the `RBreak` object is used to make a global jump.

Like CRuby, it can't be caught by `rescue`.
It is also the same as CRuby that it can be canceled in the middle by `ensure`.

 ### How to find the corresponding tag with `throw`

The called `catch` method remains in the call stack, and the tag also remains in the stack at that time.
So it is possible to find the called location by searching the two.

Note that no method can be given to the `proc` object specified in `RBreak`.
Therefore, inside the `catch` method, the argument block is called in a seemingly meaningless closure.

Also, as a countermeasure against `alias` etc., the `proc` object, which is the body of the `catch` method, is saved when mrbgem is initialized.

### DIFF
--- a/mrbgems/mruby-catch/mrblib/catch.rb
+++ b/mrbgems/mruby-catch/mrblib/catch.rb
@@ -1,4 +1,4 @@
-class ThrowCatchJump < Exception
+class UncaughtThrowError < ArgumentError
   attr_reader :_tag, :_val
   def initialize(tag, val)
     @_tag = tag
@@ -9,14 +9,14 @@ end
 
 module Kernel
   def catch(tag=Object.new, &block)
-    block.call(tag)
-  rescue ThrowCatchJump => e
-    unless e._tag.equal?(tag)
-      raise e
-    end
-    return e._val
+    # A double closure is required to make the nested `catch` distinguishable
+    # and because `break` goes back to `proc->upper`.
+    -> { -> { block.call(tag) }.call }.call
   end
   def throw(tag, val=nil)
-    raise ThrowCatchJump.new(tag, val)
+    __throw(tag, val)
+    raise UncaughtThrowError.new(tag, val)
   end
+
+  __preserve_catch_method
 end
--- a/mrbgems/mruby-catch/src/mruby-catch.c
+++ b/mrbgems/mruby-catch/src/mruby-catch.c
@@ -1,0 +1,70 @@
+#include <mruby.h>
+#include <mruby/class.h>
+#include <mruby/variable.h>
+#include <mruby/error.h>
+#include <mruby/proc.h>
+#include <mruby/presym.h>
+
+#define ID_PRESERVED_CATCH MRB_SYM(__preserved_catch_proc)
+
+static const mrb_callinfo *
+find_catcher(mrb_state *mrb, mrb_value tag)
+{
+  mrb_value pval = mrb_obj_iv_get(mrb, (struct RObject *)mrb->kernel_module, ID_PRESERVED_CATCH);
+  mrb_assert(mrb_proc_p(pval));
+  const struct RProc *proc = mrb_proc_ptr(pval);
+
+  const mrb_callinfo *ci = mrb->c->ci;
+  size_t n = ci - mrb->c->cibase;
+  ci--;
+  for (; n > 0; n--, ci--) {
+    const mrb_value *arg1 = ci->stack + 1;
+    if (ci->proc == proc && mrb_obj_eq(mrb, *arg1, tag)) {
+      return ci;
+    }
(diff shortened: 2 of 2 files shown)


dearblue:
I came up with a way to find the corresponding tag in the call stack and tried it.

The way to throw a `RBreak` object through a closure from within a method is the one I came up with in the `mruby-timeout` mrbgem in the `timeout` branch of my mruby repository.

mrbgems/mruby-catch/mrblib/catch.rb
```diff
@@ -9,14 +9,14 @@ def initialize(tag, val)
...
   def catch(tag=Object.new, &block)
-    block.call(tag)
-  rescue ThrowCatchJump => e
-    unless e._tag.equal?(tag)
-      raise e
-    end
-    return e._val
+    # A double closure is required to make the nested `catch` distinguishable
+    # and because `break` goes back to `proc->upper`.
+    -> { -> { block.call(tag) }.call }.call
```
matz, line 14: I am not sure I understand why double lambda is needed. Could you show me the case we need double lambda?

dearblue: My initial idea was a single lambda, but testing the nested `catch` added in #5323 revealed a problem.

After considering the cause,
 1. The `RBreak` object specifies the `proc` that issues the `break`.
 2. In the call stack unwind process, `proc-> upper` is targeted in comparison with `mrb_callinfo::proc`.  
    ref https://github.com/mruby/mruby/blob/5735d7c2fd18c7d92bfe7a7d1d68547e5765c60d/src/vm.c#L2145
 3. The `proc` object as the `catch` method does not change from the time of initialization.
 4. If `proc->upper` points to the `catch` method, it reacts to the innermost `catch`.

Therefore, `proc` and `proc->upper` need to be changed each time, which requires the generation of double lambdas.

Extra:
I mentioned that the unwind process of `break` will compare `proc->upper`, but to do this directly by comparing with `RBreak::proc`, it need to change the VM.
I think that probably leads to simplification of the exception handling part, but I haven't succeeded.
If all goes well, I think `RBREAK_TAG` will be combined into `RBREAK_TAG_BREAK`, `RBREAK_TAG_JUMP` and `RBREAK_TAG_STOP`.
It's not working well so far, so it's a mochi that I drew in a picture.

mrbgems/mruby-catch/mrblib/catch.rb
```diff
@@ -9,14 +9,14 @@ def initialize(tag, val)
...
   end
   def throw(tag, val=nil)
-    raise ThrowCatchJump.new(tag, val)
+    __throw(tag, val)
+    raise UncaughtThrowError.new(tag, val)
```
dearblue, line 18: `UncaughtThrowError#initialize` is defined in `catch.rb`, so it was separated to avoid reentry to the VM when raising an exception.

matz: This reentrance should not cause a new problem because every exception class generation from `mrb_raise()` and `mrb_raisef()` calls `mrb_obj_new()` that calls VM recursively. Avoiding one call here does not improve overall situation.

mrbgems/mruby-catch/mrblib/catch.rb
matz: OK. When I tried single lambda worked for nested catches as well. I don't know what caused the difference.
In any way I updated `mruby-catch` but I left double lambda unchanged.

mrbgems/mruby-catch/mrblib/catch.rb
dearblue: Thank you for your reply. I understood that.

mrbgems/mruby-catch/mrblib/catch.rb
dearblue: I tried single lambda again, but it still fails `rake test`.

After thinking about it, I realized that you might have kept `ci[2].proc` as is.
https://github.com/mruby/mruby/blob/d178e147f5cd86af247fc4b6cfafbde1308ede33/mrbgems/mruby-catch/src/mruby-catch.c#L98

If that guess is correct, it can pass the `rake test`, but it seems to cause another problem.

```console
% cat catch1.rb
def xxx
  proc { throw 1 }
end

catch(1, &xxx)

% bin/mruby catch1.rb
catch1.rb:5: break from proc-closure (LocalJumpError)
```

matz: OK, I will keep double lambda. Thank you!

PR #5343, opened 2021-02-13
feat: add Dependabot config file

jbampton:
Set for GitHub Actions to check for update everyday.

When you add or update the dependabot.yml file, this triggers an immediate check for version updates. Any options that also affect security updates are used the next time a security alert triggers a pull request for a security update.

https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Basic set up
+
+version: 2
+updates:
+
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


matz:
We don't have any package dependency from within `mruby/mruby`. What does this PR do for?
Probably we should add this to `mruby/mruby.github.io`?

jbampton:
This update to is to keep the GitHub Actions up to date.  All the linters, spell checkers, whitespace checker and code-QL all run on GitHub actions and they have versions as they as just software libraries tooo.  We need to keep them up to date too.

Example below but we have ones for `actions/checkout@v2`, `actions/setup-python@v2`, `actions/setup-node@v2.1.4`, `github/codeql-action/analyze@v1`, `github/codeql-action/autobuild@v1`, `github/codeql-action/init@v1`

https://github.com/mruby/mruby/blob/d04c651405189e132bd19985435d4c1c5fa1762b/.github/workflows/lint.yml#L29

For example `setup-python` is a verified action:

https://github.com/actions/setup-python

Seen in the GitHub Marketplace here:

https://github.com/marketplace/actions/setup-python

PR #5345, opened 2021-02-13
chore: fix missing HTTPS on links

jbampton:


mrbgems/mruby-compiler/core/y.tab.c
```diff
@@ -16,7 +16,7 @@
    GNU General Public License for more details.
 
    You should have received a copy of the GNU General Public License
-   along with this program.  If not, see <http://www.gnu.org/licenses/>.  */
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.  */
```
shuujii, line 19: The same change was made before, but I don't think this change makes much sense because this comment is generated by Bison.

matz: Ah, it will be overwritten next time. Never mind.

PR #3468, opened 2017-02-25
Try Binding class

ksss:
I'm trying to implement CRuby's `Binding` class. 
`Binding#local_variables` is easy.
But I couldn't implement `Binding#eval` because it's too difficult.
It failed in this case.

```rb
class A
  class << self
    def b
      @a = 42
      b = 'local_variable'
      binding
    end
  end
end
p A.b.eval("[@a, b]")
# Expect: [42, "local_variable"]
# Actual(This branch): [42, []]
```

Does anyone have a good idea?

PR #5362, opened 2021-02-22
Binding

dearblue:
I have implemented `Binding`. This is a continuation from #3468.

### Implementations of new classes and methods (compared to Ruby-3.0)

  - `mrbgems/mruby-binding` (depending `mrbgems/mruby-binding-core` and `mrbgems/mruby-eval`)
      - `Binding#eval` method
  - `mrbgems/mruby-binding-core` (no dependence)
      - `Binding` class
      - `Binding#local_variable_defined?` method
      - `Binding#local_variable_get` method
      - `Binding#local_variable_set` method
      - `Binding#local_variables` method
      - `Binding#receiver` method
      - `Binding#source_location` method
      - `Kernel#binding` method
  - `mrbgems/mruby-proc-binding` (depending `mrbgems/mruby-binding`)
      - `Proc#binding` method
  - ***Not implemented***
      - `Binding#irb` (I'm interested in it, but I'm not sure if I'll implement it in the future)

### Reasons for separating everything but the `Binding#eval` method as `mrbgems/mruby-binding-core`

To separate the dependency on `mruby-compiler`.
It would be nice to have a `binding` for remote control, but I don't think the `mruby-compiler` is always needed on the receiving end in that case.

I'd like to try something like kishima-san's remote mirb at some point.
[(youtube) Remote mirb on M5Stack using mruby/c](https://www.youtube.com/watch?v=fTyNUX802pI)

### Reasons for separating the `Proc#binding` method as `mrbgems/mruby-proc-binding`

Because there are methods in core mrbgems that return proc/lambda objects, we want to minimize the problems that can arise from being able to manipulate internal variables.
By keeping them separate as different mrbgem, each user can decide if they want to include them or not.

ref: [Feature #7702: Remove Proc#binding - Ruby master - Ruby Issue Tracking System](https://bugs.ruby-lang.org/issues/7702)

But black magic is fascinating.......

## Limitations

  - The maximum number of local variables (`bind.eval("var = 1")` or `bind.local_variable_set(:var, 1)`) that can be added is 50 (net is 49 because `self` is included).
    This is defined as the `MRB_IREP_LVAR_MERGE_LIMIT` constant (can be specified in build_config in the range 8..240).
    It is precomputed before variables are taken in, and if the result exceeds the limit, all local variables before the addition are lost and an exception is raised.
    See also: `mrb_proc_merge_lvar()`

## Implementation Overview

  - Extend the private global function `mrb_exec_irep()` defined in `src/vm.c` to allow specifying a hook function to be called when returning from a given `proc`.
  - If a `binding` object is specified in `Kernel.#eval`, the return from `eval` is now hooked and the defined local variables are incorporated into the `binding` object.
  - The `binding` object will hold a `proc` object that will be used as a space for dynamically added local variables. From this variable space, it references the calling `proc`.

## Finally

When I commented on #3468 I thought it would not be possible to add local variables to binding objects, but I was able to work it out.

I hope there are no bugs left that lead to `SIGSEGV`.......


dearblue:
:scream_cat: I'm sorry, there was a bug in this implementation.
The variable defined in `binding.eval` cannot be referenced well from within the `proc` object.

```ruby
bx = binding
block = bx.eval("a = 1; proc { a }")
bx.eval("a = 2")
p bx.local_variable_get(:a)
p block.call  # Expect 2 but return 1 due to a bug
```

The cause is that even after `eval` ends, the` proc` inside keeps the old proc and keeps referencing that variable.
This conceals the `a` in the variable space that the` binding` object holds directly.

As a fix, I'm currently thinking of pre-parsed and merging variables before doing `eval` inside` binding.eval`.

matz:
I created a new issue #5491 for you. Shall I investigate it? Or do you wish to fix it by yourself?

dearblue:
Thank you for creating a new issue.
I am working on a fix. Please give me some time.

matz:
OK.

PR #5363, opened 2021-02-22
Specify tasks as dependencies rather then manual invocation

katzer:
Fixes failures when you wrapped all mruby tasks under a namespace

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -39,7 +39,8 @@ load "#{MRUBY_ROOT}/tasks/doc.rake"
 task :default => :all
 
 desc "build all targets, install (locally) in-repo"
-task :all => %w[gensym build] do
+task :all => :gensym do |t|
+  Rake::Task[t.scope.path_with_task_name('build')].invoke
   puts
   puts "Build summary:"
   puts
--- a/Rakefile
+++ b/Rakefile
@@ -39,8 +39,7 @@ load "#{MRUBY_ROOT}/tasks/doc.rake"
 task :default => :all
 
 desc "build all targets, install (locally) in-repo"
-task :all => :gensym do
-  Rake::Task[:build].invoke
+task :all => %w[gensym build] do
   puts
   puts "Build summary:"
   puts
--- a/tasks/test.rake
+++ b/tasks/test.rake
@@ -1,7 +1,5 @@
 desc "build and run all mruby tests"
-task :test => "test:build" do
-  Rake::Task["test:run"].invoke
-end
+task :test => %w[test:build test:run]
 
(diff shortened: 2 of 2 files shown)


shuujii:
This change does not work with `rake -m`.

`rake -m` will execute prerequisites in parallel. For example, `rake all` will execute `gensym` and `build` in parallel, but build may fail because `build` depends on `gensym` (in fact, CI has failed).

> Fixes failures when you wrapped all mruby tasks under a namespace

Could you please show us the code that reproduces this issue?

matz:
Thank you for the pull-request. But I don't merge a PR that fails CI.

katzer:
I tried without `-m` :see_no_evil:. Would be nice if you update the contribution guidelines.

katzer:
> Could you please show us the code that reproduces this issue?

Imagine you have your own `rake build`. To avoid naming conflicts I load the mruby's Rakefile within an namespace. Then all mruby tasks have a prefix like `mruby:build`.

That was working until 3.0.0-beta. Now it fails. Either the rake tasks does not exist because of the added namespace or it will invoke the wrong task.

shuujii:
Do you mean to edit `Rakefile` directly and add namespace? If so, I do not recommend it as such things are generally unsupported. For example, I think it is better to use another way such as the following.

```ruby
# your_rakefile
namespace :mruby do
  %w[all test].each do |task_name|
    task task_name do
      sh *%W[rake -f mruby/Rakefile #{task_name}]
    end
  end
end
namespace :your do
  task :task do |t|
    puts t.name
  end
end
```

katzer:
I did not edit the external Rakefile, instead something like that:

```ruby
namespace :mruby do
  Dir.chdir('mruby') { load 'Rakefile' }
end

task build: 'mruby:build' do
  ...
end
```

katzer:
I've updated my PR already to be compatible with `rake -m`

matz:
Thank you for the update. I will review.

shuujii:
@katzer, thank you for creating the example and the new patch. However, at least `rake mruby:test` doesn't seem to work, because some explicitly specify top-level tasks.

https://github.com/mruby/mruby/blob/a9ebd4703f8261057ae8d8929dcfe441e4f76069/tasks/test.rake#L13

I think this can be fixed as well, but I'm not sure if we can guarantee that all tasks will work. Even if it does, I'm a little worried that it won't be overly complex or hinder future improvements.

BTW, `build` task (tasks without description) are internal tasks and are not assumed to be specified directly.

katzer:
@Shokuji Good point. I tested against the stable branch. Now I see the issue...

PR #5383, opened 2021-03-18
`Float::NAN/0` should be `Float::NAN`; ref a0b3378b3

shuujii:
#### Before this patch:

```console
$ bin/mruby -e 'p(Float::NAN/0)'
Infinity
```

```console
$ bin/mruby -e 'p(Float::NAN/0)'
NaN
```


matz:
Thank you, especially for the tests.

PR #5389, opened 2021-03-26
Add IO#getbyte

take-cheeze:


mrbgems/mruby-io/mrblib/io.rb
```diff
@@ -184,6 +184,11 @@ def ungetc(substr)
     nil
   end
 
+  def ungetbyte(substr)
+    substr = substr.chr if substr.is_a? Integer
```
dearblue, line 188: `Integer#chr` is defined in `mruby-string-ext`.
So it needs to add a dependency.

mrbgems/mruby-io/mrblib/io.rb
```diff
@@ -296,6 +301,12 @@ def getc
     end
   end
 
+  def getbyte
+    ret = getc
+    return ret.getbyte 0 if ret
```
dearblue, line 306: If it is built with `MRB_UTF8_STRING` added, only the first byte of the character seems to be retrieved.

```console
% echo '🌠🌟💫' > test.txt

% bin/mruby -e 'File.open("test.txt") { |f| p f.size.times.map { f.getbyte } }'
[240, 240, 240, 10, nil, nil, nil, nil, nil, nil, nil, nil, nil]

% ruby30 -e 'File.open("test.txt") { |f| p f.size.times.map { f.getbyte } }'
[240, 159, 140, 160, 240, 159, 140, 159, 240, 159, 146, 171, 10]
```

If performance issues don't concern you, it may be better to use `IO#read`.

matz: `getbyte` should work with `MRB_UTF8_STRING`.

PR #5391, opened 2021-03-27
Fix `SIGSEGV` with mruby-method + mruby-catch

dearblue:
Previously, the following code would cause a `SIGSEGV`.

```ruby
mm = method(:throw)
define_method(:throw, ->(*args) { mm.call(*args) })
catch { |tag| throw tag }
```

I think the reason is in the `mrb_yield_with_class()` function:
- Even if a C function is called, `CI_ACC_SKIP` is used
- `cipop()` is not done if globally jumping from a C function


matz:
I just remembered `CI_ACC_DIRECT` is provided for this purpose. Reverted and fixed.

PR #5394, opened 2021-03-30
Add support for partial backtraces

fundamental:
Previously backtraces involving code built without debug information just silently hid stack frames from backtraces.

Consider the code:
```
[1,2].reverse_each do |val|
  val.invalid_method
end
```

Before this commit the output without debug symbols was:
```
        [1] test.rb:1
test.rb:2: undefined method 'invalid_method' (NoMethodError)
```

After this commit the output is:
```
        [2] test.rb:1
        [1] (unknown):missing-lineno:in reverse_each
test.rb:2: undefined method 'invalid_method' (NoMethodError)
```

With full debug information the second frame in the backtrace refers to the source in the mruby-array-ext gem.

### DIFF
--- a/mrbgems/mruby-kernel-ext/test/kernel.rb
+++ b/mrbgems/mruby-kernel-ext/test/kernel.rb
@@ -20,6 +20,9 @@ assert('Kernel.caller, Kernel#caller') do
       bar(*args)
     end
   end
+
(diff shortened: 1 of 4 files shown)


fundamental:
Looks like this commit is causing some test failures. I'll take a look into updating the test cases.

fundamental:
Looks like this should be good to go now.

PR #5415, opened 2021-04-18
Introducing the `mrb_protect_raw()` API function

dearblue:
The purpose is two-fold:

 1. to be able to specify a pointer directly when user data is used

    When using `mrb_protect()`, it is necessary to allocate objects by `mrb_obj_cptr()` function when using user data.

    Adding `mrb_protect_raw()` will make it simpler to reimplement `mrbgems/mruby-error`.

 2. to correctly unwind callinfo when an exception is raised from a C function defined as a method (the main topic)

    If a method call is made directly under `mrb_protect()` and a C function is called, control is returned from `mrb_protect()` if an exception occurs there.
    In this case, callinfo is not restored, so it is out of sync.
    Moreover, returning to mruby VM (`mrb_vm_exec()` function) in this state will indicate `ci->pc` of C function which is equal to `NULL`, and subsequent `JUMP` will cause `SIGSEGV`.

    Following is an example that actually causes `SIGSEGV`:

    - `crash.c`

      ```c
      #include <mruby.h>
      #include <mruby/compile.h>
      #include <mruby/error.h>

      static mrb_value
      level1_body(mrb_state *mrb, mrb_value self)
      {
        return mrb_funcall(mrb, self, "level2", 0);
      }

      static mrb_value
      level1(mrb_state *mrb, mrb_value self)
      {
        return mrb_protect(mrb, level1_body, self, NULL);
      }

      static mrb_value
      level2(mrb_state *mrb, mrb_value self)
      {
        mrb_raise(mrb, E_RUNTIME_ERROR, "error!");
        return mrb_nil_value();
      }

      int
      main(int argc, char *argv[])
      {
        mrb_state *mrb = mrb_open();
        mrb_define_method(mrb, mrb->object_class, "level1", level1, MRB_ARGS_NONE());
        mrb_define_method(mrb, mrb->object_class, "level2", level2, MRB_ARGS_NONE());
        mrb_p(mrb, mrb_load_string(mrb, "p level1"));
        mrb_close(mrb);
        return 0;
      }
      ```

    - compile & run

      ```console
      % `bin/mruby-config --cc --cflags --ldflags` crash.c `bin/mruby-config --libs`

      % ./a.out
      zsh: segmentation fault (core dumped)  ./a.out
      ```

    After applying this patch, it will print exception object and exit normally.

The `mrb_protect()`, `mrb_ensure()` and `mrb_rescue_exceptions()` in `mrbgems/mruby-error` have been rewritten using `mrb_protect_raw()`.

### DIFF
--- a/include/mruby/error.h
+++ b/include/mruby/error.h
@@ -132,6 +132,14 @@ MRB_API mrb_value mrb_rescue_exceptions(mrb_state *mrb, mrb_func_t body, mrb_val
                                         mrb_func_t rescue, mrb_value r_data,
                                         mrb_int len, struct RClass **classes);
 
+typedef mrb_value mrb_protect_raw_func(mrb_state *mrb, void *userdata);
+
+/**
+ * This API function behaves like `mrb_protect()`.
+ * The advantage is that it avoids objectifying the user data.
+ */
+MRB_API mrb_value mrb_protect_raw(mrb_state *mrb, mrb_protect_raw_func *body, void *userdata, mrb_bool *error);
+
 MRB_END_DECL
 
 #endif  /* MRUBY_ERROR_H */
--- a/mrbgems/mruby-error/mrbgem.rake
+++ b/mrbgems/mruby-error/mrbgem.rake
@@ -2,9 +2,4 @@ MRuby::Gem::Specification.new('mruby-error') do |spec|
   spec.license = 'MIT'
   spec.author  = 'mruby developers'
   spec.summary = 'extensional error handling'
-
-  if build.cxx_exception_enabled?
-    objs << build.compile_as_cxx("#{spec.dir}/src/exception.c")
-    objs.delete_if { |v| v == objfile("#{spec.build_dir}/src/exception") }
-  end
 end
--- a/mrbgems/mruby-error/src/exception.c
+++ b/mrbgems/mruby-error/src/exception.c
@@ -1,57 +1,38 @@
 #include <mruby.h>
-#include <mruby/throw.h>
 #include <mruby/error.h>
+
+struct protect_data {
+  mrb_func_t body;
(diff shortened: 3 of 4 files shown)


include/mruby/error.h
```diff
@@ -132,6 +132,14 @@ MRB_API mrb_value mrb_rescue_exceptions(mrb_state *mrb, mrb_func_t body, mrb_val
                                         mrb_func_t rescue, mrb_value r_data,
                                         mrb_int len, struct RClass **classes);
 
+typedef mrb_value mrb_vm_protect_func(mrb_state *mrb, void *userdata);
+
+/**
+ * This API function behaves like `mrb_protect()`.
+ * The advantage is that it avoids objectifying the user data.
+ */
+MRB_API mrb_value mrb_vm_protect(mrb_state *mrb, mrb_vm_protect_func *body, void *userdata, mrb_bool *error);
```
matz, line 141: The `_vm` functions are to be called from VM, but this function is not. We should rename the function.

dearblue: Oh, I see. I will think of another name.

dearblue:
Renamed from `mrb_vm_protect()` to `mrb_protect_raw()`.

And some minor tweaks in this PR in `mrbgems/mruby-error/src/exception.c`:
  - Removed prototype declaration of `protect_body()`
  - Removed unnecessary `mrb_gc_arena_restore()` immediately after `mrb_protect_raw()` in `mrb_ensure()`
  - Matching the type of `i` variable to `len` variable in `mrb_rescue_exceptions()`

include/mruby/error.h
matz: Thank you. I am sorry but I am not 100% satisfied with the new name either. But I will merge this anyway, and rename the function later.

PR #5433, opened 2021-04-29
Remove "parse.y" directive

zzak:
To be honest, I was really wondering why this is here, and I can't find any use for it.

It seems that compiling `y.tab.c` is all that is needed.

Am I missing something?

### DIFF
--- a/mrbgems/mruby-compiler/mrbgem.rake
+++ b/mrbgems/mruby-compiler/mrbgem.rake
@@ -15,12 +15,6 @@ MRuby::Gem::Specification.new 'mruby-compiler' do |spec|
 
   lex_def = "#{dir}/core/lex.def"
 
-  # Parser
-  file "#{dir}/core/y.tab.c" => ["#{dir}/core/parse.y", lex_def] do |t|
-    yacc.run t.name, t.prerequisites.first
-    replace_line_directive(t.name)
-  end
-
   # Lexical analyzer
   file lex_def => "#{dir}/core/keywords" do |t|
     gperf.run t.name, t.prerequisites.first


shuujii:
We will need it when we modify `parse.y`.

matz:
We put `y.tab.c` to the repository to avoid `bison` installation and version compatibility problems.

zzak:
@matz @shuujii Can you explain to me how it's used? I can't imagine how this code block is ever executed. 🤔

shuujii:
Without this code block, editing `parse.y` will not recreate `y.tab.c`.

zzak:
@shuujii Ok, I didn't know that `y.tab.c` was generated from `parse.y`. Can you point me to the code where we call this function? Or is there a separate command we run to regenerate the parser.

matz:
`bison` generates `y.tab.c` from `parse.y`. `mrb_load_exec()` and other functions around are entry points. They are called from `mruby`, `mirb` and other commands.

zzak:
Thanks @matz. I think I saw in #5437 that we were running this rake task manually to generate `y.tab.c`? It doesn't seem like `mruby` or `mirb` command would need to call `bison` directly since they are compiled with `y.tab.c` during build time.

matz:
When we touch `parse.y`, `rake` will automatically generate `y.tab.c` using `bison`. In theory, `y.tab.c` should not need to be kept in the repository (like other generated files such as `*.o`), but we have some serious versioning/installation issues on some platforms (Windows and macOS) in the past so that we had to put it in the repository. See #5007

dearblue:
- `parse.y` :: Suitable for humans. It is this file that we actually modify and extend to parse the Ruby syntax.
  - `y.tab.c` :: A file mechanically generated by `bison` based on `parse.y`. It does not consider humans reading or making changes to the code.

If `parse.y` were deleted and we had to maintain` y.tab.c` directly, we would be in tremendous pain.

`bison` is one of the programs called" parser generator ".
Information is also available on wikipedia:
  - https://en.wikipedia.org/wiki/GNU_Bison
  - https://en.wikipedia.org/wiki/Compiler-compiler

shuujii:
> Can you point me to the code where we call this function? Or is there a separate command we run to regenerate the parser.

No, there is no separate command to generate `y.tab.c`. The usual entry point is `rake` command execution. The following is a brief description of how `rake` execution generates `y.tab.c` from `parse.y` .

* `default` task depends on `all` task: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/Rakefile#L38
* `all` task executes `build` task: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/Rakefile#L41-L42
* `build` task depends on `MRuby::Build#products`: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/Rakefile#L52
* `MRuby::Build#products` includes `MRuby::Build#libmruby_core_static`: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/tasks/libmruby.rake#L6
* `MRuby::Build#libmruby_core_static` depends on `MRuby::Build#libmruby_core_objs`: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/tasks/libmruby.rake#L2
* `MRuby::Build#libmruby_core_objs` includes `y.tab.o`: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/mrbgems/mruby-compiler/mrbgem.rake#L6-L14
* `y.tab.o` depends on `y.tab.c` (Rule): https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/lib/mruby/build/command.rb#L106-L115
* `y.tab.c` depends on `parse.y`: https://github.com/mruby/mruby/blob/24dcd252f83e97bd1dbcb7abec9de3a77c28709b/mrbgems/mruby-compiler/mrbgem.rake#L19-L22

zzak:
@shuujii @dearblue @matz Thank you for the extra context, this was very helpful. Is it safe to assume then that no MRuby user will ever execute this code path, but it's used by MRuby developers when changing `parse.y`? For example: `update parse.y > rake y.tab.c > git add y.tab.c > git commit mruby/mruby`?

matz:
@zzak yes

PR #5436, opened 2021-04-29
feat(ci): add GNU Aspell as a GitHub Action

jbampton:
Prints an ordered list of spelling mistakes to the terminal -> `check-spelling-single.sh`

Custom English dictionary -> `.aspell.en.pws`

Full Shell script so far -> `check-spelling.sh` -> runs against the dictionary

This script runs with Actions and prints out all the spelling mistakes grouped by file

https://en.wikipedia.org/wiki/GNU_Aspell
https://github.com/GNUAspell/aspell
http://aspell.net/


zzak (approved):

matz:
Spell checking sounds good, but it would be a burden for us to keep maintaining the dictionary for abbreviations and technical terms.
I feel it's not worth the burden. Any idea?

jbampton:
Hey @matz I agree now too.  We already have `codespell` and `misspell` 👍 

https://github.com/mruby/mruby/blob/f678620a436aa015f9252858aa00112256ab0668/.github/workflows/lint.yml#L6

https://github.com/mruby/mruby/blob/f678620a436aa015f9252858aa00112256ab0668/.pre-commit-config.yaml#L32

PR #5437, opened 2021-05-02
Run the task only once when `parse.y` is updated

dearblue:
Previously, the task was executed as many times as there were targets, including `mruby-compiler`.

```console
% rm mrbgems/mruby-compiler/core/y.tab.c
% rake `pwd`/mrbgems/mruby-compiler/core/y.tab.c
YACC  mrbgems/mruby-compiler/core/parse.y -> mrbgems/mruby-compiler/core/y.tab.c
YACC  mrbgems/mruby-compiler/core/parse.y -> mrbgems/mruby-compiler/core/y.tab.c
```

### DIFF
--- a/mrbgems/mruby-compiler/mrbgem.rake
+++ b/mrbgems/mruby-compiler/mrbgem.rake
@@ -12,18 +12,21 @@ MRuby::Gem::Specification.new 'mruby-compiler' do |spec|
     end
   end
   build.libmruby_core_objs << objs
+end
 
+if MRuby::Build.current.name == "host"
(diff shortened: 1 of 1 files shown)


zzak:
@dearblue Is this task just something we run manually when we want to regenerate `y.tab.c`? It doesn't seem to be executed during normal rake tasks.

dearblue:
No. I am manually performing the task to clearly reproduce the problem.
Normally it does not need to perform the task manually as it will be executed automatically if needed.

zzak (approved):

PR #5460, opened 2021-05-23
Separate non-mruby licenses into files; ref #5448

dearblue:
If it is included in the source code, it will be copy instead of move.

@tsahara @matsumotory,
If you think this PR is a problem, please comment.


zzak (approved):

tsahara:
it looks okay to me.

LICENSE.mrb_float_read
```diff
@@ -0,0 +1,20 @@
+This license terms applies to the mrb_read_float() function in
```
matz, line 1: I am going to replace `mrb_float_read()` with the public domain `strtod` implementation soon (probably today). So we don't need this file.

dearblue: I updated this PR.

matsumotory:
@dearblue Looks good!

PR #5480, opened 2021-06-17
Add two more pre-commit hooks

jbampton:


matz:
Could you explain what added two lines do?

jbampton:
Both hooks are from the official `pre-commit` repo:

https://github.com/pre-commit/pre-commit-hooks

So we have:

https://github.com/pre-commit/pre-commit-hooks#check-vcs-permalinks

https://github.com/pre-commit/pre-commit-hooks#detect-private-key

matz:
Thank you.

PR #5485, opened 2021-06-19
Added `MRB_API` function to get only block arguments

dearblue:
`MRB_API` function added:
- ` mrb_get_arg_block()`
- ` mrb_block_given_p()` -- The name comes from CRuby's `rb_block_given_p ()`

At the same time, it applies to `f_instance_eval()` and `f_class_eval()` of `mruby-eval`.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -1004,6 +1004,16 @@ MRB_API const mrb_value *mrb_get_argv(mrb_state *mrb);
  */
 MRB_API mrb_value mrb_get_arg1(mrb_state *mrb);
 
+/**
+ * Retrieve block argument from mrb_state.
+ */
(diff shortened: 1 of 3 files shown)


matz:
I am not fully satisfied with `mrb_get_arg_block` (`mrb_block_given_p` is OK). Why do we need this function where we can retrieve the block argument with `mrb_get_args`. Thus I don't understand the benefit of the changes for `eval.c` neither.

dearblue:
There is no strong reason, but:
  - Because there is `mrb_get_arg{c,v,1}()` functions.
  - Because it is heavy to use `mrb_get_args()` inside the `mrb_block_given_p()` function.

matz:
OK, I will modify and merge the patch.

matz:
merged by f678620a436aa015f9252858aa00112256ab0668

PR #5488, opened 2021-06-22
Standardize Ruby file names

jbampton:


matz:
Underscores (`_`) are used to combine words in a description, and hyphens (`-`) are used to combine descriptions. There are the following two files that don't follow the rule. We will rename them.

* `android_arm64-v8a.rb` → `android_arm64_v8a.rb`
* `no-float.rb` → `host-nofloat.rb`

Additional two files, `clang-asan.rb` and `ci/gcc-clang.rb` are bit controversial. For the time being, we will keep the names as they are.

PR #5490, opened 2021-06-22
Add prettier pre-commit hook

jbampton:
Add prettier config file .prettierrc

https://prettier.io/docs/en/configuration.html

Add .prettierignore

https://prettier.io/docs/en/ignore.html

https://prettier.io/docs/en/precommit.html


README.md
```diff
@@ -6,12 +6,12 @@
 ## What is mruby
 
 mruby is the lightweight implementation of the Ruby language complying to (part
-of) the [ISO standard][ISO-standard]. Its syntax is Ruby 2.x compatible.
+of) the [ISO standard][iso-standard]. Its syntax is Ruby 2.x compatible.
```
matz, line 9: ISO is an acronym, so it should still be ISO.

matz: Or is there a rule to force labels to be lower case?

PR #5493, opened 2021-06-26
Fixed finding variables from `proc` in `binding.eval` failed

### LINKED ISSUE #5491: Binding fails to update local variables
As @dearblue reported in #5362 `Binding` fails to update local variables in procs.

```ruby
bx = binding
block = bx.eval("a = 1; proc { a }")
bx.eval("a = 2")
p bx.local_variable_get(:a)
p block.call  # Expect 2 but return 1 due to a bug
```

dearblue:
Previously the following code did not produce the expected results:
```ruby
bx = binding
block = bx.eval("a = 1; proc { a }")
bx.eval("a = 2")
p block.call  # Expect 2 but return 1 due to a bug
```

The previous implementation of `Binding#eval` evaluated the code and then merged the top layer variables.
This patch will parse and expand the variable space before making a call to `eval`.
This means that the call to `Binding#eval` will do the parsing twice.

In addition, the following changes will be made:
- Make `mrb_parser_foreach_top_variable()`, `mrb_binding_extract_proc()` and `mrb_binding_extract_env()` functions private global functions.
- Remove the `posthook` argument from `mrb_exec_irep()`.
  The `posthook` argument was introduced to implement the `binding` method.
  This patch is unnecessary because it uses a different implementation method.

ref #5362
fixed #5491

### DIFF
--- a/mrbgems/mruby-binding/src/binding.c
+++ b/mrbgems/mruby-binding/src/binding.c
@@ -1,0 +1,173 @@
+#include <mruby.h>
+#include <mruby/array.h>
+#include <mruby/class.h>
+#include <mruby/compile.h>
+#include <mruby/error.h>
+#include <mruby/proc.h>
+#include <mruby/presym.h>
+#include <mruby/string.h>
+
+mrb_noreturn void mrb_method_missing(mrb_state *mrb, mrb_sym name, mrb_value self, mrb_value args);
+void mrb_proc_merge_lvar(mrb_state *mrb, mrb_irep *irep, struct REnv *env, int num, const mrb_sym *lv, const mrb_value *stack);
+mrb_value mrb_exec_irep(mrb_state *mrb, mrb_value self, struct RProc *p);
+const struct RProc *mrb_binding_extract_proc(mrb_state *mrb, mrb_value binding);
+struct REnv *mrb_binding_extract_env(mrb_state *mrb, mrb_value binding);
+typedef mrb_bool mrb_parser_foreach_top_variable_func(mrb_state *mrb, mrb_sym sym, void *user);
+void mrb_parser_foreach_top_variable(mrb_state *mrb, struct mrb_parser_state *p, mrb_parser_foreach_top_variable_func *func, void *user);
+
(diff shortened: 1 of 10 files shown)


mrbgems/mruby-binding/src/binding.c
```diff
@@ -0,0 +1,174 @@
...
+}
+
+static mrb_value
+mrb_binding_eval(mrb_state *mrb, mrb_value binding)
+{
+  binding_eval_prepare(mrb, binding);
+
+  struct RClass *c = mrb->kernel_module;
+  mrb_method_t m = mrb_method_search_vm(mrb, &c, MRB_SYM(eval));
+  if (MRB_METHOD_UNDEF_P(m)) {
```
matz, line 149: This is an exceptional case (`Kernel#eval` made undefined explicitly) so that you can just call `assert(!MRB_METHOD_UNDEF(m))` to check.

mrbgems/mruby-binding/src/binding.c
```diff
@@ -0,0 +1,174 @@
...
+    // call `method_missing` method
+    mrb_int argc;
+    mrb_value *argv, block;
+    mrb_get_args(mrb, "*!&", &argv, &argc, &block);
+    return mrb_funcall_with_block(mrb, mrb_obj_value(mrb->kernel_module), MRB_SYM(eval), argc, argv, block);
```
matz, line 154: If you really want to call `method_missing` here, use `mrb_method_missing()` function. See `mruby-method` gem, line 9 and 57.

dearblue: Okay, I've updated it to call the `mrb_method_missing ()` function directly.
I'd like to confirm, is it okay to not call the `method_missing` method, which may be redefined?

matz: I think it's OK for mruby, since undefining `eval` is a very rare case.

dearblue: I agree with your opinion. Thank you.

Issue #1564, opened 2013-11-07
Segfault when a line starting a heredoc is the last line


utkarshkukreti:
`<<`, `<<-`, and `<<hi` all segfault; `<<\n`, `<<-\n`, and `<<hi\n` show proper error messages.

Some examples:

```
➜  bin/mruby -e '<<'
[1]    4713 segmentation fault  bin/mruby -e '<<'

➜  ruby -e '<<'
-e:1: syntax error, unexpected <<

➜  bin/mruby -e '<<-'
[1]    4738 segmentation fault  bin/mruby -e '<<-'

➜  ruby -e '<<-'
-e:1: syntax error, unexpected <<

➜  bin/mruby -e '<<--'
-e:1:2: syntax error, unexpected tLSHFT

➜  bin/mruby -e 'p <<hi
hello
hi'
"hello\n"

➜  bin/mruby -e '<<hi
quote> '
-e:3:0: can't find heredoc delimiter "hi" anywhere before EOF

➜  bin/mruby -e '<<hi'
[1]    4854 segmentation fault  bin/mruby -e '<<hi'
```

PR #5496, opened 2021-06-26
Make `mrb_load_exec()` return `nil` if there is a syntax error

dearblue:
Previously it returned `undef`.
So for example, if there was a syntax error, the following code would cause `SIGSEGV`.

```c
const char *code = "syntax ? error ?";
mrb_p(mrb, mrb_load_string(mrb, code));
```

This affects the behavior of the `MRB_API` function below:
- `mrb_load_file()`
- `mrb_load_file_cxt()`
- `mrb_load_detect_file_cxt()`
- `mrb_load_string()`
- `mrb_load_nstring()`
- `mrb_load_string_cxt()`
- `mrb_load_nstring_cxt()`

I mentioned earlier why I want to change from `undef` to `nil`, but here's why I decided to change it:
- Returns `nil` if an error occurred when returning from `mrb_top_run()`.
- The specific error can be found by confirming that `mrb->exc` is non-`NULL`.
- Even if `mrbc_context::no_exec` is true, it is sufficient to check if it is either `proc` or `nil`.

However, this change can cause compatibility issues if it has already used `mrb_undef_p()` to distinguish between syntax errors.
This is the reason for making changes to the `mrdb.c` file (although the behavior does not change with or without changes).

---

Just to be sure.
I think the current behavior of returning `undef` is non-bug.

### DIFF
--- a/mrbgems/mruby-bin-debugger/tools/mrdb/mrdb.c
+++ b/mrbgems/mruby-bin-debugger/tools/mrdb/mrdb.c
@@ -708,7 +708,7 @@ main(int argc, char **argv)
     v = mrb_load_file_cxt(mrb, args.rfp, cc);
     mrbc_context_free(mrb, cc);
   }
-  if (mrdb->dbg->xm == DBG_QUIT && !mrb_undef_p(v) && mrb->exc) {
+  if (mrdb->dbg->xm == DBG_QUIT && mrb->exc) {
     const char *classname = mrb_obj_classname(mrb, mrb_obj_value(mrb->exc));
     if (!strcmp(classname, "DebuggerExit")) {
       cleanup(mrb, &args);
@@ -732,14 +732,12 @@ main(int argc, char **argv)
   }
   puts("mruby application exited.");
   mrdb->dbg->xphase = DBG_PHASE_AFTER_RUN;
-  if (!mrb_undef_p(v)) {
-    if (mrb->exc) {
-      mrb_print_error(mrb);
-    }
-    else {
-      printf(" => ");
-      mrb_p(mrb, v);
-    }
+  if (mrb->exc) {
+    mrb_print_error(mrb);
+  }
+  else {
+    printf(" => ");
+    mrb_p(mrb, v);
   }
 
   mrdb->dbg->prvfile = "-";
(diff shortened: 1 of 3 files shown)


matz:
Thank you for this PR. Re-reviewing it now after 5 years, I attempted to apply it (extending the same change to the three other internal tools — `mruby`, `mrbc`, `mrb` — that have the same `mrb_undef_p(v)` pattern as `mrdb.c`). The bintest suite then exposed a deeper issue that this PR alone can't address.

**The `undef` return is doing more than signaling "load failed."** It's distinguishing **"parser already printed the error to stderr"** (the `capture_errors == FALSE` case used by the `mruby` command) from **"runtime error needs `mrb_print_error` to be called now."** With `undef`, the binary tools use:

```c
if (mrb->exc) {
  if (!mrb_undef_p(v)) {          // false for parser errors
    mrb_print_error(mrb);         //   → skip, parser already printed
  }                                // true for runtime errors
  n = EXIT_FAILURE;                //   → call mrb_print_error
}
```

Replacing `undef` with `nil` collapses this distinction. After the change, `mrb_print_error` is called for both cases, producing **double output** for syntax errors:

```
-e:1:2: syntax error, unexpected <<            (parser's direct output)
(unknown):0: syntax error (SyntaxError)        (mrb_print_error's output, now redundant)
```

The bintest in `mrbgems/mruby-bin-mruby/bintest/mruby.rb` ("regression for #1564", "mruby -c option", "codegen error") fails on this. Three failed test cases, seven assertions.

**Fixing this properly would require introducing a new mechanism** — for example, a flag on `mrb_state` or the exception object marking "already printed" — that lets `mrb_print_error` distinguish the two cases. That's a larger change than the PR contemplates.

**The original concern is still valid** (a naive caller doing `mrb_p(mrb_load_string(mrb, "syntax error"))` segfaults today because `mrb_p` doesn't handle `MRB_TT_UNDEF`). A different approach to that footgun — such as making `mrb_p` and other public APIs safely handle `undef` values rather than crashing — could be a cleaner path. But this PR's specific approach (changing the load functions' return value) doesn't work without the additional mechanism.

Closing in favor of a future approach. Thanks again for the work, and apologies for the long silence.

PR #5505, opened 2021-07-09
Do not include `stdint.h` before `mruby.h`

dearblue:


matz:
Recent changes in my Emacs configuration added the header lines automagically. I will investigate and fix it.

PR #5516, opened 2021-07-25
Fixed that sizeof(pointer) was used for buf size

dearblue:
In a 32-bit environment, it was fixed to 4 bytes and no writing was performed.


matz:
Thank you. I have already fixed this issue in c2c025006386a164478af4776b158b9d3f4d51df and 26340a8818e0d54b6fde7a3a1c5f38c9137065bf

PR #5533, opened 2021-08-20
Add SerenityOS crossbuild

dantecatalfamo:
Cross compile mruby for SerenityOS (https://github.com/SerenityOS/serenity)

This goes along with the [port](https://github.com/SerenityOS/serenity/pull/9525) I wrote for serenityOS


build_config/serenity-32bit.rb
```diff
@@ -0,0 +1,12 @@
+MRuby::CrossBuild.new('serenity-32bit') do |conf|
```
matz, line 1: SerenityOS is a relatively less known operating system yet. So it seems better to have a short description of the OS at the top of the configuration file.

Besides that, the configuration name is `serenity-32bit`. Do you (or they) have any plan to support non-32bit configuration on SerenityOS (maybe in the future)? If not, it should be just `serenity` (or `cross-serenity` if you plan to build on the serenity host).

dantecatalfamo: I've added a short description to the top of the file.

I also replaced the 32bit specific version with one that can compile on 64bit systems. 64bit is new in Serenity so I had to check with the developers to see how to get the correct platform.

Thank you! :)

matz: Thank you. Could you tell me if `SERENITY_ARCH` is a common environment variable with Serenity development?
If so, it's OK. Otherwise, could you add a description of the variable?

dantecatalfamo: `SERENITY_ARCH` is defined by the SerenityOS [ports system](https://github.com/SerenityOS/serenity/blob/master/Ports/.port_include.sh#L5), so it should always be there. I can add an extra comment specifying the default value just in case.

matz:
Thank you! I will merge the PR.

PR #5587, opened 2021-11-28
Fix `args_unshift()` in `mrbgems/mruby-method`

dearblue:
Both keyword arguments and block arguments were being destroyed when there were no arguments.
The cause of this is #5585. I' m sorry.


matz:
How should we treat #5585 then? Should we revert it?

dearblue:
I don't think there is a need for revert.

However, it wasn't good to mention only the keyword arguments in the #5585 commit message.
#5585's `args_unshift()` was to fix the missing block arguments.
Perhaps I should have kept the commit itself separate.

PR #5600, opened 2021-12-14
Fixing Typing Error and some Punctuation Error

Rakesh9100:
## Fixed some punctuation errors and typos !!


README.md
```diff
@@ -16,7 +16,7 @@ also able to generate compiled byte code in a C source file, see the "mrbtest"
 program under the "test" directory for an example.
 
 This achievement was sponsored by the Regional Innovation Creation R&D Programs
-of the Ministry of Economy, Trade and Industry of Japan.
+of the Ministry of Economy, Trade, and Industry of Japan.
```
matz, line 19: "the Ministry of Economy, Trade and Industry of Japan" is the official translation of the name of the [ministry](https://www.meti.go.jp/english/index.html). So we have to keep it.

README.md
```diff
@@ -49,13 +49,13 @@ See the [compile.md](doc/guides/compile.md) file.
 
 There are two sets of documentation in mruby: the mruby API (generated by yard) and C API (Doxygen)
 
-To build both of them, simply go
+To build both of them, simply navigate:
```
matz, line 52: I am not an English native, but the word "navigate" makes me feel a bit awkward. Is it natural for command-line interaction?

README.md
```diff
@@ -49,13 +49,13 @@ See the [compile.md](doc/guides/compile.md) file.
...
 
 ```
 rake doc
 ```
 
-You can also view them in your browser
+You can also view them in your browser by navigating:
```
matz, line 58: Same above regarding "navigate".

README.md
Rakesh9100: I was bit confused and did it by mistake i think, Indeed you are right.

fundamental:
I would say "To build both of them, simply run:". I wouldn't add the "by navigating:" to browser command as it's redundant. Also I don't know that "complying to STANDARD" should be altered to "complying with STANDARD" as I think the former is equally correct.

Rakesh9100:
> I would say "To build both of them, simply run:". I wouldn't add the "by navigating:" to browser command as it's redundant. Also I don't know that "complying to STANDARD" should be altered to "complying with STANDARD" as I think the former is equally correct.

Yes, 'simply run' will also be good to say. You can change it to that. But talking about complying, it is good to have “complying with”. It is formally correct.

matz:
According to the dictionary 'comply with' is the idiom. I am OK with the change. Or maybe we can use 'conform to' instead.

Issue #5613, opened 2021-12-27
Untrusted Pointer Dereference in mrb_vm_exec()


AiDaiP:
# Untrusted Pointer Dereference in mrb_vm_exec()

## Description

An Untrusted Pointer Dereference was discovered in mrb_vm_exec(). The vulnerability causes a segmentation fault and application crash.

**version**

6de0fcb

```
./mruby -v
mruby 3.0.0 (2021-03-05)
```

**System information**
Ubuntu 20.04 focal, AMD EPYC 7742 64-Core @ 16x 2.25GHz

## Proof of Concept

**poc**

```
base64 poc
Y2xhc3MgVmNsYXNzDQoJQEB2YXJyID0gWzE8MiwzLDQsNSw2LDcsOCw5LDEwLDExLDEyLDEzLDE0
LDE1LDE2LDE3XQ0KCWRlZiB2YXJyDQoJCUBAdmFycg0KCWVuZA0KCWRlZiB0b19pJnQNCgkJQEB2
YXJyLmNsZWFyDQoJCTExDQoJZW5kDQplbmQNCg0Kb2JqID0gVmNsYXNzLm5ldw0KDQpwcmludCBv
YmoudmFyci5zaGlmdChvYmop
```

**command:**

```
./mruby ./poc
```

**Result**

```
./mruby ./poc
[1]    283553 segmentation fault  ./mruby ./poc
```

**gdb**

```
Program received signal SIGSEGV, Segmentation fault.
mrb_vm_exec (mrb=<optimized out>, proc=<optimized out>, pc=0x55555561f7c3 <mrblib_proc_iseq_115+99> "/\006\003\001=\005\001%\377\331Q\006\002\001\a\004Q\b\003/\a\004\001<\006Q\a\004<\006\070\006") at /home/aidai/fuzzing/mruby/mruby-master/include/mruby/boxing_word.h:139
139       return x;
LEGEND: STACK | HEAP | CODE | DATA | RWX | RODATA
───────────────────────────────[ REGISTERS ]────────────────────────────────
 RAX  0x10
 RBX  0x38
 RCX  0x7
 RDX  0x0
 RDI  0x1
 RSI  0xffffffff
 R8   0x6
 R9   0x0
 R10  0x1
 R11  0x55555566fb08 ◂— 0x10
 R12  0x2eb
 R13  0x10
 R14  0x7
 R15  0x2f
 RBP  0x0
[...]

PR #5620, opened 2021-12-30
Adjusting the stack for after it enters the virtual machine

dearblue:
ref. #5613.

I mentioned in #5540 that there was no reentrant to the virtual machine, but in fact it was still a possibility at that point.
Also, the variable `ci` needs to be recalculated at the same time.

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -930,6 +930,13 @@ mrb_block_given_p(mrb_state *mrb)
 MRB_API mrb_int
 mrb_get_args(mrb_state *mrb, const char *format, ...)
 {
+#define ADJUST_STACK() do { \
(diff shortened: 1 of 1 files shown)


matz:
Since b6283978c586702e1b2ff182b83ceb3e2bf691b3, `mrb_as_int` no longer call `mrb_vm_exec` recursively. But `mrb_as_float` still does (due to float conversion from Rational and Complex numbers).

dearblue:
#5622 has been merged, so I think the changes contained in this PR are unnecessary.
Or am I missing something?
Just in case?

matz:
Correct. I am going to cancel this change.

PR #5626, opened 2022-01-04
Revert the order of error backtrace display

dearblue:
This follows the change in CRuby-3.0.

### DIFF
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -99,27 +99,26 @@ each_backtrace(mrb_state *mrb, ptrdiff_t ciidx, each_backtrace_func func, void *
 static void
 print_backtrace(mrb_state *mrb, struct RObject *exc, mrb_value backtrace)
 {
-  mrb_int i;
+  mrb_value mesg = mrb_exc_inspect(mrb, mrb_obj_value(exc));
   mrb_int n = RARRAY_LEN(backtrace);
-  mrb_value *loc, mesg;
+  mrb_value *loc;
 
   if (n != 0) {
-    if (n > 1) {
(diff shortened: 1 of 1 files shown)


src/backtrace.c
```diff
@@ -99,27 +99,26 @@ each_backtrace(mrb_state *mrb, ptrdiff_t ciidx, each_backtrace_func func, void *
...
+    loc++;
+    n--;
   }
   mesg = mrb_exc_inspect(mrb, mrb_obj_value(exc));
   fwrite(RSTRING_PTR(mesg), RSTRING_LEN(mesg), 1, stderr);
   fputc('\n', stderr);
+  for (; n > 0; n--, loc++) {
+    if (mrb_string_p(*loc)) {
+      fprintf(stderr, "\tfrom %.*s\n",
+              (int)RSTRING_LEN(*loc), RSTRING_PTR(*loc));
```
dearblue, line 119: `mrb_exc_inspect ()` can call `mrb_vm_exec ()`, and the variables `n` and` loc` can have incorrect values.

dearblue: Fixed it.

matz:
Could you tell me the reason behind this change? I did the current way to print the error messages last (Python way).
I believe this is better. And I once tried to change the CRuby behavior but we had a compatibility issue.

dearblue:
There is no reason other than that it is the way it is in CRuby.
I think it's more likely to be compared to CRuby than Python.

Personally, I prefer the current behavior of CRuby.

However, I don't have a strong reason, so I'm withdrawing the proposal.

matz:
Thank you

Issue #5574, opened 2021-11-11
Name clashes in mgem-list entries


suetanvil:
When checking out the git repository associated with an item in `mgem-list`, the checked-out repository takes its name from the source repository, **not** the mgem itself and the build system does not enforce that the names are consistent.

In my specific example, `mgem-process.gem` points to repository `https://github.com/iij/mruby-process.git` while `mgem-process2.gem` points to `https://github.com/katzer/mruby-process.git`.  If I include them in the same `buld_config.rb` file:

```ruby
MRuby::Build.new do |conf|
  conf.toolchain

  conf.gem :mgem => 'mruby-process'
  conf.gem :mgem => 'mruby-process2'  

  conf.enable_bintest
  conf.enable_test
end
```

`rake` will only check out one of them. 

(In this specific case, this often passes its tests because `mruby-process2` is derived from `mruby-process`.  However, the build will sometimes fail after a `rake clean` because the `.lock` file points to a commit ID that is only in the other repository.  However, the potential for hard-to-diagnose bugs in this case is pretty high.)

My initial idea for a fix is to name the checkout after the `.gem` file instead of the originating repository but I haven't yet looked into what that entails.

katzer:
Hi, I am the author of `mruby-process2`. I wonder why you're adding both gems to your build. Even if the build system would pick the right repo, both mgems override each other in terms of functionality.

suetanvil:
Hi @katzer,

The above example is contrived.  However, I've run into the case where
I explicitly use `mruby-process2` but several dependencies use
`mruby-process` (specifically `mruby-dir-glob` and `mruby-open3` in my
case.)  This results in the whole build using either `mruby-process`
or `mruby-process2` with no clear way to specify which one wins.
Currently, my dependency order seems to get yours first so everything
works as expected (**and** none of the dependencies are incompatible
with it), but there doesn't seem to be a way to guarantee that.

(As an aside, I seem to have run into a related issue in the
`mruby-process2` gem as well, which I'll file with you later.  I
mention it because it's possible that I've mistaken that for this
issue in some cases.)

PS: Thanks for your ongoing maintenance of `mruby-process2`; it's a
really useful gem.

suetanvil:
(@katzer: it turns out I was adding the dependency incorrectly, so the apparent other issues is my mistake.  I still maintain that the mruby rakefiles should handle duplicate repository names correctly, though.)

dearblue:
[...]

PR #5638, opened 2022-01-28
Gracefully handle clashing git checkout names (plus load_gems.rb refactor)

suetanvil:
This is a fix for https://github.com/mruby/mruby/issues/5574 .

The problem is that when retrieving gem dependencies from git repositories, if two or more repositories have the same basename, the retrieval system will get confused and treat it as one or another, depending on declaration order.  This is also the case when using different branches or commits from the same repository.

This change fixes this as follows:

1. It is now an error to attempt to clone more than one gem to the same directory **unless** they are identical (same url, branch, and commit ID).
2. It is possible to declare the first import of a gem as being "canonical". This will cause the system to ignore duplicate imports with a message.

In addition, I have refactored `load_gems.rb` to improve maintainability.  Specifically, the function `load_special_path_gem` has been turned into a class with the functionality spread across multiple descriptively-named private methods.

### DIFF
--- a/doc/guides/mrbgems.md
+++ b/doc/guides/mrbgems.md
@@ -33,6 +33,9 @@ conf.gem :github => 'masuidrive/mrbgems-example', :branch => 'master'
 conf.gem :bitbucket => 'mruby/mrbgems-example', :branch => 'master'
 ```
 
+NOTE: `:bitbucket` option supports only git. Hg is unsupported in this
+version.
+
 You can specify the subdirectory of the repository with `:path` option:
 
 ```ruby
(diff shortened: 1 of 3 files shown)


matz:
Could you clean up the pull-request?
- cancelling rebuild.sh
- separate unrelated trailing space removal of README.md in mruby-errno

suetanvil:
I'll try to do that within the next to or 3 days.

suetanvil:
The requested changes have been made. Please let me know if you need anything else.

matz:
Merged.

PR #5640, opened 2022-02-05
Check the type of the keyword rest argument

dearblue:
If a non-hash object is passed in `OP_SEND` and `OP_SUPER`, an error should occur.

```ruby
kwd = "string"
p(**kwd)  #  => expect raised TypeError
```

This PR also integrates the code that was copied and pasted in #5628.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1163,6 +1163,35 @@ hash_new_from_values(mrb_state *mrb, mrb_int argc, mrb_value *regs)
 void mrb_method_added(mrb_state *mrb, struct RClass *c, mrb_sym mid);
 mrb_value mrb_str_aref(mrb_state *mrb, mrb_value str, mrb_value idx, mrb_value len);
 
+#define ARGUMENT_NORMALIZE(arg_base, arg_info, insn) do { \
+  int n = *(arg_info)&0xf; \
+  int nk = (*(arg_info)>>4)&0xf; \
+  mrb_int bidx = (arg_base) + mrb_bidx(*(arg_info)); \
+  if (0 < nk && nk < CALL_MAXARGS) {  /* pack keyword arguments */ \
+    mrb_int kidx = (arg_base)+(n==CALL_MAXARGS?1:n)+1; \
+    mrb_value kdict = hash_new_from_values(mrb, nk, regs+kidx); \
+    regs[kidx] = kdict; \
+    nk = CALL_MAXARGS; \
+    *(arg_info) = n | (nk<<4); \
+  } \
+  \
+  mrb_assert(bidx < irep->nregs+(arg_base)); \
+  mrb_int new_bidx = (arg_base)+mrb_bidx(*(arg_info)); \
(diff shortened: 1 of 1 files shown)


dearblue:
It turns out that there is a problem that this patch does not fix.

```console
% bin/mruby -e 'p [**1]'
[1]

% ruby31 -e 'p [**1]'
-e:1:in `<main>': no implicit conversion of Integer into Hash (TypeError)
```

I also found a problem where the hash object is destroyed when it is expanded as a keyword argument at the beginning of the argument.

```console
% bin/mruby -e 'h = {}; p(**h, a: 1, b: 2); p h'
{:a=>1, :b=>2}
{:a=>1, :b=>2}

% ruby31 -e 'h = {}; p(**h, a: 1, b: 2); p h'
{:a=>1, :b=>2}
{}
```

The solution I came up with after viewing the irep for the above two problems is to add an implicit empty hash by `codegen.c` if the first element of `NODE_KW_HASH` is `NODE_KW_REST_ARGS`.
I've mimicked the behavior by hand, and it seems to be the result I'm expecting.

```console
% bin/mruby -e 'p [**{}, **1]'
-e:1: Integer cannot be converted to Hash (TypeError)

% bin/mruby -e 'h = {}; p(**{}, **h, a: 1, b: 2); p h'
{:a=>1, :b=>2}
{}
```

So far, I have not been able to write a patch to fix it (If it is a patch that crashes, I can prepare it right away).
Hmmm... I would like to leave it to @matz...

dearblue:
Is this correct?

```diff
diff --git a/mrbgems/mruby-compiler/core/codegen.c b/mrbgems/mruby-compiler/core/codegen.c
index b90eae3e8..20b7d5c89 100644
--- a/mrbgems/mruby-compiler/core/codegen.c
+++ b/mrbgems/mruby-compiler/core/codegen.c
@@ -1605,10 +1605,16 @@ gen_hash(codegen_scope *s, node *tree, int val, int limit)
   if (cursp() >= GEN_LIT_ARY_MAX) slimit = INT16_MAX;
   int len = 0;
   mrb_bool update = FALSE;
+  mrb_bool first = TRUE;

   while (tree) {
     if (nint(tree->car->car->car) == NODE_KW_REST_ARGS) {
-      if (val && len > 0) {
+      if (val && first) {
+        genop_2(s, OP_HASH, cursp(), 0);
+        push();
+        update = TRUE;
+      }
+      else if (val && len > 0) {
         pop_n(len*2);
         if (!update) {
           genop_2(s, OP_HASH, cursp(), len);
@@ -1647,6 +1653,7 @@ gen_hash(codegen_scope *s, node *tree, int val, int limit)
       update = TRUE;
       len = 0;
     }
+    first = FALSE;
   }
   if (val && len > limit) {
     pop_n(len*2);
```

matz:
I think the patch in [the above comment](https://github.com/mruby/mruby/pull/5640#issuecomment-1038104677) should work.
Could you make a pull request?

PR #5644, opened 2022-02-06
Support `disable_presym` for `mrbgems/mruby-errno`

dearblue:
The `mrbgems/mruby-errno/src/known_errors_def.cstub` file has been extended, so the `mrbgems/mruby-errno/src/known_errors_e2c.cstub` file is no longer needed.


matz:
Thank you for the patch. I am still wondering if I should remove `disable_presym` though.

dearblue:
I do not object to `disable_presym` being removed.
I was wondering while working on this if it was really necessary to make it `disable_presym` compatible.
However, since `disable_presym` exists and is included in `stdlib.gembox`, I just decided that this action is necessary.

matz:
Reasonable. Let me think for a while to determine if we should remove `disable_presym`.

PR #5667, opened 2022-03-13
Reorganize the `struct RException`

dearblue:
With this change, the lower 16 bits of `struct RBasic::flags` are unused.
If this PR is accepted, a patch will be prepared to store `errno` of `SystemCallError` in the lower 10 bits of `struct RBasic::flags` if possible (I'm currently working on it, though).

### DIFF
--- a/include/mruby/error.h
+++ b/include/mruby/error.h
@@ -18,6 +18,7 @@ struct RException {
   MRB_OBJECT_HEADER;
   struct iv_tbl *iv;
   struct RObject *mesg;         // NULL or probably RString
+  struct RObject *backtrace;    // NULL, RArray or RData
 };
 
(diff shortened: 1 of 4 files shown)


include/mruby/error.h
```diff
@@ -17,11 +17,14 @@ MRB_BEGIN_DECL
 struct RException {
   MRB_OBJECT_HEADER;
   struct iv_tbl *iv;
-  struct RString *mesg;
+#ifdef MRB_WORD_BOXING
```
matz, line 20: I think this is unnecessary complexity. Should we always use `struct RObject* mesg`?

dearblue: I understand. Then let's do so.

PR #5675, opened 2022-03-24
Add missing include stdio.h to mruby-print

jasl:
CLion tips me because "Implicit declaration of function 'fwrite' is invalid in C99"


matz:
`stdio.h` is included from `mrbconf.h` unless `MRB_NO_STDIO` is defined. You don't need to include it explicitly.
Or are we missing something?

PR #5687, opened 2022-04-10
Must the `mrb_sys_fail()` function raise an exception

dearblue:
Normal return of control from the `mrb_sys_fail()` function has unexpected results.
Example: double free, malloc with huge size used as error integer

The following is an example of an actual crash:

```console
% cat crash.rb
def SystemCallError._sys_fail(*a)
  nil
end

File.readlink("/404")

% bin/mruby crash.rb
zsh: segmentation fault (core dumped)  bin/mruby crash.rb
```

### DIFF
--- a/include/mruby/error.h
+++ b/include/mruby/error.h
@@ -23,7 +23,7 @@ struct RException {
 #define mrb_exc_ptr(v) ((struct RException*)mrb_ptr(v))
 #define MRB_EXC_MESG_STRING_FLAG 0x100
 
-MRB_API void mrb_sys_fail(mrb_state *mrb, const char *mesg);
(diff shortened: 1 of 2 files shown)


dearblue:
As a point of reference

As far as I can tell, the script shown does not crash mruby 3.1.0-rc2.
It crashes after commit a4f740ff9ab8f42fd242eb34290821624d0d8b94 + 1a9991f16575b922746e6e656d7e05e5812bbda9.

The reason for the crash in the example taken is that `RSTR_EMBEDDABLE_P(len = -1)` is true before `str_init_embed()`, resulting in an illegal write (buffer over run).

Last but not least, even mruby 3.1.0-rc2 can return control normally from a call to `mrb_sys_fail()`.
Therefore, it is still possible for subsequent processing to cause unexpected results.

PR #5696, opened 2022-04-23
Move bigint implementation files to `mruby-bigint`

dearblue:
Users can now switch to their own implementation of GEM.
However, we do not guarantee that this will not be a problem in the current situation.
This may need to be improved in the future.

And, a part of `bigint.h` has been removed because it duplicates `include/mruby/internal.h`.

 - - -

  - `mrbgems/mruby-bigint/core/bigint.h` ⬅ `include/mruby/bigint.h`
  - `mrbgems/mruby-bigint/core/bigint.c` ⬅ `src/bigint.c`

The files are placed in ***the `core/` directory***. Note that it is ***not the `src/` directory***.

I also plan to add tips for creating custom GEMs to `mrbgems/mruby-bigint/README.md`, using the GNU Multi-Precision Library as an example.

### DIFF
--- a/include/mruby/bigint.h
+++ b/include/mruby/bigint.h
@@ -85,33 +85,6 @@ struct RBigint {
 
 #define RBIGINT(v) ((struct RBigint*)mrb_ptr(v))
 
-mrb_value mrb_bint_new_int(mrb_state *mrb, mrb_int x);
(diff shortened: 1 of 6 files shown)


matz:
I am not sure how useful to allow switching implementation of big integers, but I merged this PR anyway.
At least it makes the core implementation a little bit cleaner.

PR #5703, opened 2022-05-02
Strict acquisition of the `Random` class given by `mrb_get_args()`

dearblue:
This is to prevent data corruption of the substituted `Random` instance.
If the `Object::Random` constants are redefined to a different `MRB_TT_ISTRUCT` object class, they can pass `mrb_get_args()` validation.

The following is an example of how the `Array#shuffle` method rewrites the data of a non `Random` instance.

```c
// code.c

#include <mruby.h>
#include <mruby/class.h>
#include <mruby/compile.h>
#include <mruby/istruct.h>

#define EVAL_LIT(MRB, ...) mrb_load_string((MRB), #__VA_ARGS__)

static mrb_value
fake_initialize(mrb_state *mrb, mrb_value self)
{
  char *p = ISTRUCT_PTR(self);

  for (int i = 0; i < ISTRUCT_DATA_SIZE; i++) {
    p[i] = 'A' + i;
  }

  return self;
}

static mrb_value
fake_to_bytes(mrb_state *mrb, mrb_value self)
{
  const char *p = ISTRUCT_PTR(self);
  return mrb_str_new(mrb, p, ISTRUCT_DATA_SIZE);
}

int
main(int argc, char *argv[])
{
  mrb_state *mrb = mrb_open();

  struct RClass *fake = mrb_define_class(mrb, "Fake", mrb->object_class);
  MRB_SET_INSTANCE_TT(fake, MRB_TT_ISTRUCT);
  mrb_define_method(mrb, fake, "initialize", fake_initialize, MRB_ARGS_NONE());
  mrb_define_method(mrb, fake, "to_bytes", fake_to_bytes, MRB_ARGS_NONE());

  EVAL_LIT(mrb,
    Random = Fake \n
    fake = Fake.new \n
    p [fake, fake.to_bytes] \n
    p a = (1..10).to_a \n
    p a.shuffle(fake) \n
    p [fake, fake.to_bytes]
  );

  mrb_close(mrb);

  return 0;
}
```

```console
% `bin/mruby-config --cc --cflags --ldflags` code.c `bin/mruby-config --libs`
% ./a.out
[#<Fake:0x8007d8f20>, "ABCDEFGHIJKLMNOPQRSTUVWX"]
[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
[1, 7, 8, 10, 2, 6, 9, 3, 4, 5]
[#<Fake:0x8007d8f20>, "\xe81{\v\fM\xb0D<\x99\xb1V+l\x84\x86QRSTUVWX"]
```

### DIFF
--- a/mrbgems/mruby-random/src/random.c
+++ b/mrbgems/mruby-random/src/random.c
@@ -167,6 +167,7 @@ random_default(mrb_state *mrb) {
 
 #define random_ptr(v) (rand_state*)mrb_istruct_ptr(v)
 #define random_default_state(mrb) random_ptr(random_default(mrb))
+#define ID_RANDOM_STRICT MRB_SYM(mruby_Random)
 
 static mrb_value
 random_m_init(mrb_state *mrb, mrb_value self)
@@ -245,7 +246,7 @@ mrb_ary_shuffle_bang(mrb_state *mrb, mrb_value ary)
   rand_state *random;
 
   if (RARRAY_LEN(ary) > 1) {
-    struct RClass *c = mrb_class_get_id(mrb, MRB_SYM(Random));
+    struct RClass *c = mrb_class_get_id(mrb, ID_RANDOM_STRICT);
     if (mrb_get_args(mrb, "|I", &random, c) == 0) {
       random = random_default_state(mrb);
     }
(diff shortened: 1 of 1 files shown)


dearblue:
I was not aware of this until recently.
I should add a note about this to `mrb_get_args()` in `include/mruby.h`.

dearblue:
Since `mrb_const_get()` can be used as is, I changed it to be defined as an invisible constant from Ruby.
I think this can be an example for users to use `MRB_TT_ISTRUCT`.
\# I will write a separate document.

PR #5706, opened 2022-05-04
Added note on `I` specifier of `mrb_get_args()`

dearblue:
I thought it was redundant, so I stopped writing the same content in `src/class.c`.


matz:
Thank you for the detailed note. But issues like #5703 were caused by the mistake of API design introduced by #3251. I should have noticed the issue beforehand. So I decided to remove `I` specifier altogether from `mrb_get_args` to reduce the risk. The `I` specifier behavior has been changed by #5530, so the compatibility was lost anyway.

dearblue:
I understand and agree with your decision.
I will also close this PR as it is no longer needed.

Issue #5709, opened 2022-05-17
Argument `path:` of `MRuby::Build` configuration method `gem` does not work as documented


suetanvil:
The mrbgem guide (`doc/guides/mrbgems.md`) documents the `path:`
argument for `conf.gem` as being used to point to the root directory
of a gem when that gem is retrieved via git but is located in a
subdirectory of the checkout.
    
The actual code does not do this.  Instead, it treats path as (mostly)
identical to the `gemdir:` argument; that is, it points to a local
directory containing the gem.

I have a fix for this that restores the documented behaviour that I will be submitting shortly.

suetanvil:
The pull request is [here](https://github.com/mruby/mruby/pull/5710).

PR #5710, opened 2022-05-17
Fix `conf.gem path:` build_config argument.

suetanvil:
The mrbgem guide (`doc/guides/mrbgems.md`) documents the `path:`
argument for `conf.gem` as being used to point to the root directory
of a gem when that gem is retrieved via git but is located in a
subdirectory of the checkout.

The actual code does not do this.  Instead, it treats path as (mostly)
identical to the `gemdir:` argument; that is, it points to a local
directory containing the gem.

This change fixes this by making `path:` behave as documented.

It also adds a rudimentary `build_config.rb` file that exercises most of (?) the `conf.gem` keyword arguments.


matz:
1. I don't want to add a file `conftest/test_build_config.rb`.
2. if you add `close #5709` the issue will be auto closed.

suetanvil:
Will do.

PR #5719, opened 2022-06-08
Updates`mrb->c->ci` after `mrb_env_unshare()`

dearblue:
If GC occurs in `mrb_malloc()` called by `mrb_env_unshare()`, `ci->u.env` may be destroyed.
If destroyed it, `ci->u.env->tt` becomes `MRB_TT_FREE`.
When control is returned to `mrb_env_unshare()`, `struct free_obj::next` in the same offset as `struct REnv::stack` is rewritten.
Unexpected results then occur when the object is reused.


dearblue:
This PR may reiterate the problem fixed in #3087.
I will close it.

PR #5720, opened 2022-06-08
Add `MRB_DEBUG_ALWAYS_GC` configuration

dearblue:
Many widely known modern desktop operating systems have a mechanism called memory overcommitment, which usually makes it difficult to make `malloc()` fail.
By introducing this configuration, `mrb_full_gc()` is forced by `mrb_realloc_simple()`.
This is intended for debugging mruby's internal implementation, but may be useful for general users as well.
However, there are still problems with mruby at this time.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -224,6 +224,9 @@ mrb_realloc_simple(mrb_state *mrb, void *p,  size_t len)
 {
   void *p2;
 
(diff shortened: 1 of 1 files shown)


dearblue:
Hmm, this should probably be merged with `MRB_GC_STRESS`.
It is used in several `build_config/*` files, so it may need to be rewritten so they don't use it.

PR #5726, opened 2022-06-23
fix typo otherBasis -> orthoBasis

aoki1980taichi:
The original function name in ao.c was orthoBasis.
I guess the function is generating orthonormal basis (https://en.wikipedia.org/wiki/Orthonormal_basis).


matz:
Thank you!

PR #5728, opened 2022-06-24
Fix mismatched braces in `extern "C"` blocks when compiling as C++

lopopolo:
When compiling mruby with `-DMRB_USE_CXX_EXCEPTION`, clang fails to
compile and emits these warnings:

    vendor/mruby/src/vm.c:3066:1: error: extraneous closing brace ('}')
    } /* end of extern "C" */
    ^
    vendor/mruby/src/vm.c:3072:7: error: expected '}'
    #endif
          ^
    vendor/mruby/src/vm.c:3070:12: note: to match this '{'
    extern "C" {
               ^
    2 errors generated.

Fixup the implementation of the `extern "C"` block in `vm.c`.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -27,9 +27,6 @@
 extern "C" {
 #endif
 void abort(void);
-#if defined(__cplusplus)
-}  /* extern "C" { */
-#endif
 #endif
 
 #define STACK_INIT_SIZE 128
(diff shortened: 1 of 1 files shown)


src/vm.c
```diff
@@ -27,9 +27,6 @@
 extern "C" {
 #endif
 void abort(void);
-#if defined(__cplusplus)
```
lopopolo, line 30: i don't think this is quite right and I'm not sure if this works if `MRB_NO_STDIO` is not defined.

These are all of the flags I'm passing to clang++:

```
            .arg("-DMRB_ARY_NO_EMBED")
            .arg("-DMRB_GC_TURN_OFF_GENERATIONAL")
            .arg("-DMRB_INT64")
            .arg("-DMRB_NO_BOXING")
            .arg("-DMRB_NO_PRESYM")
            .arg("-DMRB_NO_STDIO")
            .arg("-DMRB_USE_CXX_EXCEPTION")
            .arg("-DMRB_UTF8_STRING");
```

Issue #5725, opened 2022-06-22
weird `const_missing` exceptions in mrblib code (mruby 3.1.0 and above)


kazuho:
I've been trying to use mruby-require with the recent versions of mruby (see https://github.com/iij/mruby-require/pull/29) and have noticed one behavior that seems like an issue with mruby 3.1.0 and above.

When I run mruby-require of the PR linked above with mruby 3.0.0, everything works fine. But when I run it using mruby 3.1.0, a const_missing error is raised complaining that `File.exist?` cannot be found.

__Steps to reproduce:__
1. Checkout https://github.com/iij/mruby-require/pull/29.
2. Build mruby with the following configuration:
```ruby
MRuby::Build.new do |conf|
  conf.toolchain :gcc
  conf.gembox "default"
  conf.gem "mruby-require"
end
```
3. Run the following script using mruby being built:
```ruby
print "#{File.file?("/")}\n"
require "/nonexistent"
```
When I run this script using mruby 3.0.0 being built, the first line emits "false" (because `/` is a directory), then complains that `/nonexistent` cannot be loaded. Fine.
But when I run this script using mruby 3.1.0 being built, it raises a const_missing exception [here](https://github.com/iij/mruby-require/blob/5afbed21d2cf6e8f0c75fa9d4e0397a8ff893d32/mrblib/require.rb#L47).

__Why I think this is a bug of mruby:__

While this error is related to mruby-require, I'm reporting it here because I suspect this is a problem of mruby, due to the following reasons:
* the version of mruby-require being used does not have external dependency outside of mrbgems bundled by mruby
* the "gem_init" function of mruby-require just calls `mrb_define_method` and does nothing else, the problem that we see seems to happen before any logic of mruby-require written in C is being invoked

Therefore, I'm suspecting if there's an issue with the ruby code found in mrbgems having issues with accessing constants defined in other mrbgems.

kazuho:
Did a bisect, it seems that this is a regression caused by #5272.

kazuho:
@matz Of course. Thank you for looking into the issue.

When I run the ruby script above, I get:
```
% build/host/bin/mruby ~/tmp/defect.rb       
false
trace (most recent call last):
	[5] /Users/kazuho/tmp/defect.rb:2
	[4] /Users/kazuho/mydev/3p/mruby-require/mrblib/require.rb:40:in require
	[3] /Users/kazuho/mydev/3p/mruby/mrblib/enum.rb:78:in find
[...]

PR #5733, opened 2022-06-25
Do not expand "included class" in `mrb_proc_new()`.

dearblue:
If `target_class->tt` is `MRB_TT_ICLASS`, it may cause loss of class inheritance relationships.
This problem is caused by #5402.

ref. #5725

### DIFF
--- a/src/proc.c
+++ b/src/proc.c
@@ -52,15 +52,7 @@ mrb_proc_new(mrb_state *mrb, const mrb_irep *irep)
     struct RClass *tc = NULL;
 
     if (ci->proc) {
-      if (ci->proc->color != MRB_GC_RED) {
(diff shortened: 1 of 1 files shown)


matz:
This partially reverts caeacea. Is this really OK?

dearblue:
Sorry, I can't recall the reason that led to the change in #5402.
But, Yes, I think it is fine.

I actually ran `bin/mrbtest` in GDB with a "break point" on the current master https://github.com/mruby/mruby/blob/d3e8494f8ddc15b2a50dcc88432a4039300cb67e/src/proc.c#L61, and it only caught the `proc` object in [`catch_irep_2`](https://github.com/mruby/mruby/blob/d3e8494f8ddc15b2a50dcc88432a4039300cb67e/mrbgems/mruby-catch/src/catch.c#L33).

PR #5739, opened 2022-06-27
Fix C comments in `src/class.c`

jbampton:


dearblue:
The comment block beginning with `/*!` are collected by doxygen.
https://www.doxygen.nl/manual/docblocks.html#cppblock

In the case of mruby, the `rake doc` command can generate yard and doxygen documentation.
I don't think this PR is desirable, because it will not generate the document `mruby/doc/capi/html/mruby_8h.html#ae34c3bf2d5fc9cea641b7f4cf88db216`, for example.

PR #5750, opened 2022-07-17
Place `LocalJumpError` messages in ROM

dearblue:
Note that the meaning of the operand `a` in `OP_ERR` has changed.

---

Perhaps `RITE_BINARY_MAJOR_VER` needs to be updated.

### DIFF
--- a/include/mruby/internal.h
+++ b/include/mruby/internal.h
@@ -45,6 +45,23 @@ mrb_value mrb_get_backtrace(mrb_state *mrb);
 void mrb_exc_mesg_set(mrb_state *mrb, struct RException *exc, mrb_value mesg);
 mrb_value mrb_exc_mesg_get(mrb_state *mrb, struct RException *exc);
 
+#define MRB_LOCALJUMP_ERROR_FOREACH(def) \
+  def(MRB_LOCALJUMP_ERROR_BREAK,  "unexpected break") \
+  def(MRB_LOCALJUMP_ERROR_NEXT,   "unexpected next") \
+  def(MRB_LOCALJUMP_ERROR_REDO,   "unexpected redo") \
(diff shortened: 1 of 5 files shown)


matz:
Thank you for this PR. Closing it after re-reviewing — the situation has changed in master since 2022 in a way that makes most of this PR's motivation moot.

**The vm.c half is already addressed differently in master.** In commit 9d0617f27b ("Remove `localjump_error()`", 2024-08-15), the dynamic string concatenation in `localjump_error()` was eliminated by replacing the call sites with direct `RAISE_LIT(mrb, E_LOCALJUMP_ERROR, "unexpected return")` etc. — string literals avoid the heap allocation that the original `localjump_error()` was doing, without needing a shared message table. So the "Suppress heap allocation for strings" goal of the PR's vm.c portion is already achieved.

**What remains is the `OP_ERR` codegen-side optimization** — replacing the iseq pool string with an index into a global table. That part is still applicable in principle, but:

- The PR notes that `RITE_BINARY_MAJOR_VER` would need to be bumped, since the meaning of the `OP_ERR` `a` operand changes. Bumping the bytecode major version invalidates all existing `.mrb` files in the wild — that's a heavy cost.
- The benefit is bytecode-size only: ~50–100 bytes saved per `.mrb` file containing `LocalJumpError`-emitting paths (which are uncommon in real code; usually 0–1 such instructions per script).
- The cost/benefit doesn't justify a standalone major version bump.

If a future change bumps `RITE_BINARY_MAJOR_VER` for an unrelated reason, this `OP_ERR` optimization would be a natural opportunistic addition. Until then, the existing pool-based `OP_ERR` is fine.

Closing in favor of "fold into the next bytecode format bump." Thanks again for the work.

PR #5777, opened 2022-08-11
Protect prepared arguments when GC occurs in `cipush()`

dearblue:
When `cipush()` extends "callinfo", GC may occur.
In this case, there was a problem that prepared arguments were spoiled depending on the situation when `cipush()` is called.
Therefore, in the problematic part, `cipush()` is prepared first, and then the arguments are prepared.

The reason for passing `CINFO_DIRECT` to `cipush()` is that it is simply ignored in the `MRB_CATCH()` part of `mrb_vm_exec()`.

Also, the argument processing parts of `mrb_funcall_with_block()` and `mrb_yield_with_class()` are combined and made independent as `funcall_args_capture()`.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -586,6 +586,35 @@ prepare_missing(mrb_state *mrb, mrb_value recv, mrb_sym mid, struct RClass **cls
   return m;
 }
 
+static uint16_t
+funcall_args_capture(mrb_state *mrb, int stoff, mrb_int argc, const mrb_value *argv, mrb_value block)
+{
+  if (argc < 0 || argc > INT32_MAX) {
+    mrb_raisef(mrb, E_ARGUMENT_ERROR, "negative or too big argc for funcall (%i)", argc);
+  }
+
+  if (argc < CALL_MAXARGS) {
+    int extends = stoff + argc + 2 /* self + block */;
+    stack_extend_adjust(mrb, extends, &argv);
+
+    mrb_value *args = mrb->c->ci->stack + stoff + 1 /* self */;
+    stack_copy(args, argv, argc);
+    args[argc] = block;
+
+    return argc;
+  }
+  else {
+    int extends = stoff + 3 /* self + splat + block */;
+    stack_extend_adjust(mrb, extends, &argv);
+
+    mrb_value *args = mrb->c->ci->stack + stoff + 1 /* self */;
+    args[0] = mrb_ary_new_from_values(mrb, argc, argv);
+    args[1] = block;
+
+    return CALL_MAXARGS;
+  }
+}
+
 MRB_API mrb_value
 mrb_funcall_with_block(mrb_state *mrb, mrb_value self, mrb_sym mid, mrb_int argc, const mrb_value *argv, mrb_value blk)
 {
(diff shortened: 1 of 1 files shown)


dearblue:
Perhaps the problem occurs on embedded systems that do not have an MMU, for example.
To observe the problem on a desktop OS, apply a temporary patch to `cipush()` and `rake` it.

```patch
index 61286c2c5..6ff5cff6e 100644
--- a/src/vm.c
+++ b/src/vm.c
@@ -336,6 +336,7 @@ cipush(mrb_state *mrb, mrb_int push_stacks, uint8_t cci,
   struct mrb_context *c = mrb->c;
   mrb_callinfo *ci = c->ci;
 
+  mrb_full_gc(mrb);
   if (ci + 1 == c->ciend) {
     ptrdiff_t size = ci - c->cibase;
 
```

### (1) Unintended error in `#const_missing` method

```
% bin/mruby -e 'MISSING_CONST'
        [1] -e:1
-e:1:in const_missing: nil is not a symbol nor a string (TypeError)
```

This is because in `mrb_funcall_with_block()`, the prepared argument stack is not protected from GC and becomes `nil` by `mark_context_stack()`.

### (2) The `#method_missing` method causes the given block to be lost

```console
% cat cipush-gc1.rb
class C
  def method_missing(*args, &block)
    yield args, block
  end
end


C.new.missing! { |*args| p args }

% bin/mruby cipush-gc1.rb
        [1] cipush-gc1.rb:7
cipush-gc1.rb:3:in method_missing: unexpected yield (LocalJumpError)
```

This is because the splat argument is plugged by `prepare_missing()` in `OP_SENDB` of `mrb_vm_exec()`.
The extruded block exceeds `ci->proc->body.irep->nregs` and is set to `nil` by the GC of `cipush()`.

### (3) Giving more than `CALL_MAXARGS` arguments to `mrb_funcall_with_block()` causes `SIGSEGV`

```console
% cat cipush-gc2.c
#include <mruby.h>

int
main(int argc, char *argv[])
{
  mrb_state *mrb = mrb_open();

  const mrb_value args[16] = {
    mrb_nil_value(), mrb_nil_value(), mrb_nil_value(), mrb_nil_value(),
    mrb_nil_value(), mrb_nil_value(), mrb_nil_value(), mrb_nil_value(),
    mrb_nil_value(), mrb_nil_value(), mrb_nil_value(), mrb_nil_value(),
    mrb_nil_value(), mrb_nil_value(), mrb_nil_value(), mrb_nil_value(),
  };
  mrb_value str = mrb_str_new_lit(mrb, "");
  mrb_funcall_argv(mrb, str, mrb_intern_lit(mrb, "[]"), 16, args);

  mrb_close(mrb);

  return 0;
}

% `bin/mruby-config --cc --cflags --ldflags` cipush-gc2.c `bin/mruby-config --libs`
% ./a.out
zsh: segmentation fault (core dumped)  ./a.out
```

This is because the stack placed as a splat argument is not protected from GC and becomes `nil`.
Then, in `mrb_get_args()`, while `ci->n == 15`, a pointer is taken out in the form `ARY_PTR(nil)`, resulting in a null pointer dereference.

Issue #5781, opened 2022-08-12
A exception in fiber corrupses call-stack. Only when using mrb_funcall().


haramako:
Hi.

I encountered a problem in Fiber.

When call Fiber#resume by mrb_funcall() and fiber raise a exception.
Stack is broken.

But call Fiber#resume in mrb_load_string() is OK. (So in ruby scripting, no problem)

I wrote a reproduct code of this problem.

```c
// fiber_problem.c in mruby directory.
#include "mruby.h"
#include "mruby/compile.h"

int main(int argc, char **argv)
{
    int n = 0; // Change to n = 1, no error occured.

    struct mrb_state *mrb = mrb_open();

    mrb_value fiber = mrb_load_string(mrb, "$f = Fiber.new { raise 'err' }"); // raise in fiber

    if( n == 0 ){
        mrb_funcall(mrb, fiber, "resume", 0); // NG
    }else{
        mrb_load_string(mrb, "$f.resume"); // OK
    }

    mrb_load_string(mrb, "1"); // # segv in mrb_vm_ci_target_class() when n == 0

    mrb_close(mrb);

    return 0;
}
```

```
$ gcc -I include fiber_problem.c build/host/lib/libmruby.a -lm && ./a.out
zsh: segmentation fault (core dumped)  ./a.out
```

I read fiber's code and try to fix this problem, but not yet.

Regard,

dearblue:
If my understanding is correct, we cannot call the `Fiber#resume` method in `mrb_funcall()`.
It should lead to an exception (`can't cross C function boundary (FiberError)`).

May need to do `fiber_check_cfunc()` with `fiber_resume()`?

```diff
diff --git a/mrbgems/mruby-fiber/src/fiber.c b/mrbgems/mruby-fiber/src/fiber.c
index 4fca6cae0..d4d872d18 100644
--- a/mrbgems/mruby-fiber/src/fiber.c
+++ b/mrbgems/mruby-fiber/src/fiber.c
@@ -267,6 +267,7 @@ fiber_resume(mrb_state *mrb, mrb_value self)
   mrb_int len;
   mrb_bool vmexec = FALSE;
 
+  fiber_check_cfunc(mrb, mrb->c);
   mrb_get_args(mrb, "*!", &a, &len);
   if (mrb->c->ci->cci > 0) {
     vmexec = TRUE;
```

haramako:
Thank you for your advice!

I'll call `Fiber#resume` from ruby.

[...]

PR #5782, opened 2022-08-13
Protect when call Fiber#resume from C function.

haramako:
When call `Fiber#resume` from C function like `mrb_funcall()` and fiber raise a exception, mruby's stack is broken.

Add `fiber_check_cfunc` in fiber_resume() to make this problem as safe error.
Because fiber can't cross C function boundary.

This change has a breaking change, when fiber raises no exception.
Before this change, mrb_funcall() make fiber run normally.
After this change, mrb_funcall() sets mrb->exc.

Reproduction code:
```c
#include "mruby.h"
#include "mruby/compile.h"

int main(int argc, char **argv)
{
    struct mrb_state *mrb = mrb_open();
    mrb_value fiber = mrb_load_string(mrb, "Fiber.new { raise 'err' }"); // raise in fiber

    mrb_funcall(mrb, fiber, "resume", 0); // corrupse ruby's stack.

    mrb_load_string(mrb, "1"); // # SEGV in mrb_vm_ci_target_class() 

    mrb_close(mrb);
    return 0;
}
```

See: #5781

### DIFF
--- a/mrbgems/mruby-fiber/src/fiber.c
+++ b/mrbgems/mruby-fiber/src/fiber.c
@@ -267,6 +267,7 @@ fiber_resume(mrb_state *mrb, mrb_value self)
   mrb_int len;
   mrb_bool vmexec = FALSE;
 
+  fiber_check_cfunc(mrb, mrb->c);
   mrb_get_args(mrb, "*!", &a, &len);
   if (mrb->c->ci->cci > 0) {
     vmexec = TRUE;


haramako:
I'm not sure the error in CI... 

It seems no related to this change.

dearblue:
I don't know what the cause is, but I sometimes see this in CI against MSVC/MinGW.
My guess is that it is a mix of `rake` output and `mrbc` output to `stdout`.
I'm sure the mruby project admins will restart CI if they deem it necessary, but since other versions have been successful, it's probably not a problem.

PR #5783, opened 2022-08-16
Use Super-Linter Slim Image

jbampton:
Should improve workflow time.

https://github.com/github/super-linter#slim-image


jbampton:
Screen Shot 2022-08-17 at 9 00 04 am

Screen Shot 2022-08-17 at 8 59 54 am

PR #5790, opened 2022-09-15
Add `lint` target to the Makefile

jbampton:
Runs `pre-commit run --all-files`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,3 +13,7 @@ test : all
 clean :
 	$(RAKE) clean
 .PHONY : clean
+
+lint :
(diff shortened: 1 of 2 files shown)


matz:
I think the target should be added to Rakefile as well.
Besides that, pre-commit check is not really a lint. Should we rename it?

jbampton:
I could not think of a better name for the `task` since `pre-commit` does run some linters.

matz:
I have renamed `lint` to `check`. I want to keep the term `lint` for static **software** analysis.

PR #5793, opened 2022-09-17
Strict references to `SystemCallError` and `Errno` in C

dearblue:
For example, if `SystemCallError` is changed, it may invite undesirable behavior.

```console
% bin/mruby -e '
E = SystemCallError
SystemCallError = String
E._sys_fail(1)
'
-e:4: allocation failure of String (TypeError)
```


matz:
Hmm, it's users fault to see the failure when they redefine class names (if we don't see fatal failures).
All we have to do is improving errors to be more comprehensive, probably.

dearblue:
Indeed it is. Thanks for reviewing it.

PR #5803, opened 2022-10-09
Update year in LICENSE file

bovi:
It's nearly 2023 but better late than never 🙃


dearblue:
The same goes for the `include/mruby.h` file. I think it would be better to update this one as well.

bovi:
> The same goes for the `include/mruby.h` file. I think it would be better to update this one as well.

thanks for the hint. done

PR #5804, opened 2022-10-09
Using `mrb_funcall_argv()` instead of `mrb_funcall_id()`

dearblue:
`mrb_funcall_id()` is an attractive and easy-to-use function, but it requires more stack space. This is because the `mrb_funcall_id()` function takes variable-length arguments and also requires copying to a fixed-length array.

If the `mrb_funcall_argv()` function is called directly, stack consumption can be avoided. However, it requires an array to be defined, making it less usable.

Therefore, the `MRB_FUNCALL()` macro function is introduced to improve usability. This is for internal implementation at this time, because if the number of arguments is wrong, the error will be hard to understand at compile time.

Currently, only calls with up to 4 arguments are made. For reserve purposes, calls with up to 6 arguments are valid. Calls with 7 to 13 arguments will output a "function not defined" error at compile time. For 14 or more arguments, a seemingly inexplicable error is output.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -96,7 +96,6 @@
 
 #ifndef mrb_static_assert
 # define mrb_static_assert1(exp) mrb_static_assert2(exp, #exp)
-# define mrb_static_assert_expand(...) __VA_ARGS__ /* for MSVC behaviour - https://stackoverflow.com/q/5530505 */
 # define mrb_static_assert_selector(a, b, name, ...) name
 /**
  * The `mrb_static_assert()` macro function takes one or two arguments.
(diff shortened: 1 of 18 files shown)


dearblue:
Also, `mrb_static_assert_expand()` has been renamed to `MRB_VA_EXPAND()` and moved to `include/mruby/common.h`.

matz:
Closing without merge — the underlying motivation (avoiding the 16-element fixed argv in `mrb_funcall_id` for callers that pass far fewer args) is valid, but the variadic-macro-with-PP-counting machinery and the hard 6-arg cap have downsides: confusing diagnostics for >6 args, harder to gdb-step into, and the `MRB_VA_EXPAND`/`MRB_VA_ARGS_COUNT` infrastructure leaking into common.h.

The actual call sites in tree are all argc=1 or argc=2 (12 in total), so a much smaller surface — two `static inline` helpers `mrb_funcall_argv1` / `mrb_funcall_argv2` — covers the same need without any preprocessor counting, without an arbitrary cap, and visible as plain function names in the debugger. I'll add those and switch the in-tree callers in a separate change.

Thanks for the analysis and the prototype; the diff here was a useful map of where `mrb_funcall_id` is used.

PR #5811, opened 2022-10-21
ops.h: update `OP_ARYDUP` instruction and rename  to `OP_ARYSPLAT`

FnControlOption:
Transforms the value of a splat inside an array or a return statement. For example, `return *nil` should return `nil.to_a`, while `return *1` should return `[1]`

### DIFF
--- a/mrbgems/mruby-compiler/core/codegen.c
+++ b/mrbgems/mruby-compiler/core/codegen.c
@@ -2202,7 +2202,7 @@ gen_retval(codegen_scope *s, node *tree)
   if (nint(tree->car) == NODE_SPLAT) {
     codegen(s, tree, VAL);
     pop();
-    genop_1(s, OP_ARYDUP, cursp());
+    genop_1(s, OP_ARYSPLAT, cursp());
   }
   else {
     codegen(s, tree, VAL);
--- a/src/vm.c
+++ b/src/vm.c
@@ -2748,15 +2748,10 @@ RETRY_TRY_BLOCK:
       NEXT;
     }
 
-    CASE(OP_ARYDUP, B) {
-      mrb_value ary = regs[a];
-      if (mrb_array_p(ary)) {
-        ary = mrb_ary_new_from_values(mrb, RARRAY_LEN(ary), RARRAY_PTR(ary));
-      }
-      else {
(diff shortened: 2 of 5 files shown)


mrbgems/mruby-compiler/core/codegen.c
```diff
@@ -2202,7 +2202,7 @@ gen_retval(codegen_scope *s, node *tree)
   if (nint(tree->car) == NODE_SPLAT) {
     codegen(s, tree, VAL);
     pop();
-    genop_1(s, OP_ARYDUP, cursp());
+    genop_1(s, OP_ARYSPLAT, cursp());
   }
```
matz, line 2206: Since this is an only usage of `OP_ARYDUP`, we should rename (if needed) instead of adding a new instruction.

matz:
Adding new instructions can invalidate existing compiled binaries (plus other programs e.g. mruby/c). Thus, we want to avoid unless it's absolutely necessary.

matz:
@FnControlOption Yes.

src/vm.c
```diff
@@ -2748,15 +2748,9 @@ mrb_vm_exec(mrb_state *mrb, const struct RProc *proc, const mrb_code *pc)
...
-      mrb_value ary = regs[a];
-      if (mrb_array_p(ary)) {
-        ary = mrb_ary_new_from_values(mrb, RARRAY_LEN(ary), RARRAY_PTR(ary));
-      }
-      else {
-        ary = mrb_ary_new_from_values(mrb, 1, &ary);
-      }
-      regs[a] = ary;
+    CASE(OP_ARYSPLAT, B) {
+      regs[a] = mrb_ary_splat(mrb, regs[a]);
```
dearblue, line 2752: Can you please make assignments to `regs` via temporary variables?
See commit 8981b2fb28dbaed76078aff6c84f19b90832aaaf.
\# 🐈‍⬛  I tend to forget too.

dearblue: No, we have no problem using them if necessary.
However, we need to separate statements when we make assignments to `regs`, and function or method calls that involve GC occurrences in a single statement.

In this case, for example, it is better to do the following:

```c
mrb_value ary = mrb_ary_splat(mrb, regs[a]);
regs[a] = ary;
```

matz:
Thank you!

PR #5815, opened 2022-10-22
codegen.c: exit headless `case` early if a condition is always true

FnControlOption:
Without this, the following program would hang:

```rb
case
when true, false
  puts :success
end
```

Old VM instructions: (Notice the infinite loop created by `JMP 000`)

```
    2 000 JMP		000
    2 003 JMP		016
    3 006 LOADSYM	R2	:success	
    3 009 SSEND		R1	:puts	n=1
    3 013 JMP		018
    3 016 LOADNIL	R1		
    3 018 RETURN	R1		
    3 020 STOP
```

New VM instructions:

```
    3 000 LOADSYM	R2	:success	
    3 003 SSEND		R1	:puts	n=1
    3 007 RETURN	R1		
    3 009 STOP
```

### DIFF
--- a/mrbgems/mruby-compiler/core/codegen.c
+++ b/mrbgems/mruby-compiler/core/codegen.c
@@ -2606,6 +2606,25 @@ codegen(codegen_scope *s, node *tree, int val)
       }
       tree = tree->cdr;
       while (tree) {
+        if (!head) {
+          n = tree->car->car;
+          while (n) {
+            if (true_always(n->car)) {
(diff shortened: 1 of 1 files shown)


matz:
Since some problems remains for cases like:

```ruby
a = true
case
when a,false
  p "success"
end
```
We have reverted this pull-request. Instead, I stop the peephole optimization (for `OP_JMPIF`) that causes the problem.

PR #5816, opened 2022-10-23
Makefile/Rakefile: add checkupdate target

jbampton:
Runs `pre-commit autoupdate`


matz:
Hmm, `update` may confuse users what to update. Maybe `checkupdate` or `updatecheck`?

PR #5833, opened 2022-10-26
Remove `# coding: utf-8` from Ruby files

jbampton:
The default script encoding is Encoding::UTF-8 after v2.0.

https://ruby-doc.org/core-2.1.2/Encoding.html#class-Encoding-label-Script+encoding


matz:
My Emacs configuration automagically add the config line if the file contains non-ASCII characters.

PR #5834, opened 2022-10-27
kernel.c (mrb_cmp_m): use mrb_equal instead of mrb_obj_equal.

FnControlOption:
`Kernel#<=>` should also call `#==` to check if the objects are equal.

```rb
class Foo
  def ==(_)
    true
  end
end
p Foo.new <=> Object.new # => 0
```

~~Admittedly, I'm not sure if this is worth the performance hit. Perhaps it would be better to simply revert c1697a9.~~

### DIFF
--- a/src/kernel.c
+++ b/src/kernel.c
@@ -83,7 +83,7 @@ mrb_cmp_m(mrb_state *mrb, mrb_value self)
 {
   mrb_value arg = mrb_get_arg1(mrb);
 
-  if (mrb_obj_equal(mrb, self, arg))
+  if (mrb_equal(mrb, self, arg))
     return mrb_fixnum_value(0);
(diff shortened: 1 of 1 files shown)


mrblib/compar.rb
```diff
@@ -48,7 +48,11 @@ def <= other
   #
   # ISO 15.3.3.2.3
   def == other
+    @@eq_call_stack ||= []
+    return false if @@eq_call_stack.include?([self.object_id, other.object_id])
```
matz, line 52: I am afraid that this recursive comparison check may consume too much time and memory.
Do we really need this?

matz: Hmm, let me think about the issue. The cost is not preferable.

PR #5835, opened 2022-10-28
use WSAGetLastError() for errno on Windows

mattn:
The error of select(2) on Windows should be given from WSAGetLastError().

https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-select#return-value

### DIFF
--- a/mrbgems/mruby-io/src/io.c
+++ b/mrbgems/mruby-io/src/io.c
@@ -1258,8 +1258,13 @@ mrb_io_s_select(mrb_state *mrb, mrb_value klass)
 retry:
   n = select(max, rp, wp, ep, tp);
   if (n < 0) {
+#ifdef _WIN32
+    if (WSAGetLastError() != WSAEINTR)
+      mrb_sys_fail(mrb, "select failed");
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-io/src/io.c
```diff
@@ -1258,8 +1258,13 @@ mrb_io_s_select(mrb_state *mrb, mrb_value klass)
 retry:
   n = select(max, rp, wp, ep, tp);
   if (n < 0) {
+#ifdef _WIN32
+    if (WSAGetLastError() != WSAEINTR)
+      mrb_sys_fail(mrb, "select failed");
```
dearblue, line 1263: I am wondering if `WSAGetLastError()` sets `errno`.
If it does not, it might be preferable to have `errno = EINTR` before `mrb_sys_fail()`.

🐈‍⬛ I may PR it if I can confirm it (but I don't know when that will be).

mattn: Ah, right. I'll fix in later.

PR #5838, opened 2022-10-29
Remove `charset = latin1` from `.editorconfig` for batch files

jbampton:
The top `.editorconfig` rule `charset = utf-8` now applies to batch files

```
mruby % file -I mrbgems/mruby-bin-config/mruby-config.bat
mrbgems/mruby-bin-config/mruby-config.bat: text/x-msdos-batch; charset=us-ascii
```

US-ASCII (basic English) is a 7-bit, 128 characters code page, originally designed for telegraphy.

UTF-8 encoding uses the same encoding as 7-bit ASCII for its first 128 characters. So a text file that only contains characters from that range of the first 128 characters will be identical at a byte level whether encoded with UTF-8 or 7-bit ASCII.

https://stackoverflow.com/questions/11303405/force-encode-from-us-ascii-to-utf-8-iconv

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -29,8 +29,8 @@ indent_style = space
 trim_trailing_whitespace = true
 #max_line_length = 120
 
+# limitation to US-ASCII
 [*.bat]
-charset = latin1
 end_of_line = crlf
 #max_line_length = 80
 


jbampton:
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#charset

https://editorconfig.org/#supported-properties

dearblue:
Ah, I mistakenly equated "US-ASCII" with "Latin1".

If I remember correctly, batch files change encoding depending on the codepage used to run the file.
For example, if Windows is a Japanese environment, it is usually treated as ["CP932"](https://en.wikipedia.org/wiki/Code_page_932_%28Microsoft_Windows%29).

It was my intention to limit the character set to 7-bit characters to minimize this effect.

I do not oppose this PR.
Hmmm, even if I could, it would only be to put the limitation to US-ASCII as a comment in `.editorconfig`.......

PR #5843, opened 2022-10-30
Simplify the `L_RAISE` block

dearblue:
The current flow of the `L_RAISE` block is as follows:

```mermaid
flowchart LR

raise((raise))
raise --> citop{"top of call info?"}

citop -- YES --> chftop{"no catch handler?"}
citop -- NO --> chf{"no catch handler?"}

chftop -- YES --> fiber{"root of fiber?"}
chftop -- NO --> vmexec((vmexec))

fiber -- YES --> vmstop((vmstop))
fiber -- NO --> fswitch["switch fiber"]
fswitch --> citop

chf -- YES --> cipop["cipop"]
chf -- NO --> vmexec((vmexec))
cipop --> citop
```

This patch makes the catch handler search first:

```mermaid
flowchart LR

raise((raise))
raise --> chf{"no catch handler?"}

chf -- NO --> vmexec((vmexec))
chf -- YES --> citop{"top of call info?"}

citop -- YES --> fiber{"root of fiber?"}
citop -- NO --> cipop["cipop"]
cipop --> chf

fiber -- YES --> vmstop((vmstop))
fiber -- NO --> fswitch["switch fiber"]
fswitch --> chf
```

---

It is easier to understand the changes if you check with `git diff -w`.
<https://github.com/mruby/mruby/pull/5843/files?w=1>

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -2189,39 +2189,31 @@ RETRY_TRY_BLOCK:
       if (mrb->exc) {
       L_RAISE:
         ci = mrb->c->ci;
-        if (ci == mrb->c->cibase) {
-          ch = catch_handler_find(mrb, ci, pc, MRB_CATCH_FILTER_ALL);
-          if (ch == NULL) goto L_FTOP;
-          goto L_CATCH;
-        }
         while ((ch = catch_handler_find(mrb, ci, pc, MRB_CATCH_FILTER_ALL)) == NULL) {
-          ci = cipop(mrb);
-          if (ci[1].cci == CINFO_SKIP && prev_jmp) {
-            mrb->jmp = prev_jmp;
-            MRB_THROW(prev_jmp);
-          }
-          pc = ci[0].pc;
-          if (ci == mrb->c->cibase) {
-            ch = catch_handler_find(mrb, ci, pc, MRB_CATCH_FILTER_ALL);
-            if (ch == NULL) {
-            L_FTOP:             /* fiber top */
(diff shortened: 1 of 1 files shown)


dearblue:
After this PR, we are considering moving the `L_RAISE` block.
Current candidates will be either of the following:

https://github.com/mruby/mruby/blob/4371b736b7dadad7366c14b4cb5e53292d057713/src/vm.c#L1356-L1362
or
https://github.com/mruby/mruby/blob/4371b736b7dadad7366c14b4cb5e53292d057713/src/vm.c#L1705-L1716

dearblue:
PS: The move targets the `L_BREAK` block as well.

matz:
If you want to move `L_MOVE`, I prefer L1713.

dearblue:
This PR was motivated by an attempt to move the `L_RAISE` block out of the instruction dispatch.
So I have not yet benchmarked it when the move is to `OP_RAISEIF`.

For reference, when I moved it into L1356, there seemed to be little performance impact, for better or worse.

There is also an attempt to replace `struct RBreak::proc` with ci-index.
I will be moving the `L_RAISE` / `L_BREAK` blocks as I see how it goes with this one.

After those, I would like to reduce some `CHECKPOINT_MAIN()`.
\# :black_cat: Currently, this test does not succeed.......

If these work, the outlook for `OP_RETURN` should be a bit better.

matz:
So should I wait til you move `L_RAISE`, or just merge this PR?

PR #5844, opened 2022-10-31
ruby: remove unneeded parenthesis from methods

jbampton:


lib/mruby/build/load_gems.rb
```diff
@@ -71,11 +71,11 @@ def full_gemdir
         return File.join(@gemdir, @path)
       end
 
-      def canonical?()  return @canonical;                  end
-      def git?()        return !!@repo;                     end
-      def gemname()     return File.basename(@gemdir);      end
+      def canonical?;  return @canonical;                  end
+      def git?;        return !!@repo;                     end
```
matz, line 75: For single line `def`s, parentheses `()` look better than semicolons (`;`). At least, I prefer them.

PR #5846, opened 2022-11-01
Fixed `mrb_jmpbuf::jmpbuf_id` to be a C++ linkage

dearblue:
Using `conf.enable_cxx_exception` in GCC (g++) caused a compile error.

ref. #5728


matz:
I have fixed the issue with different approach.

PR #5848, opened 2022-11-04
Reduce memory copy in `mrb_str_inspect()` with UTF-8

dearblue:
:scream_cat: When I compiled optimally with gcc, I got a warning.
However, I could not understand why this warning was printed.

```console
% gcc12 -Iinclude -DMRB_NO_PRESYM -DMRB_UTF8_STRING -Wall -O3 -c src/string.c
src/string.c: In function 'str_escape':
src/string.c:1225:18: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
 1225 |           buf[i] = p[i];
      |           ~~~~~~~^~~~~~
src/string.c:1209:8: note: at offset 4 into destination object 'buf' of size 4
 1209 |   char buf[4];  /* `\x??` or UTF-8 character */
      |        ^~~
src/string.c:1225:18: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
 1225 |           buf[i] = p[i];
      |           ~~~~~~~^~~~~~
src/string.c:1209:8: note: at offset 5 into destination object 'buf' of size 4
 1209 |   char buf[4];  /* `\x??` or UTF-8 character */
      |        ^~~
```

By taking a larger buffer size, the warning went away.

```diff
diff --git a/src/string.c b/src/string.c
index 0e204118a..00a3b87b4 100644
--- a/src/string.c
+++ b/src/string.c
@@ -1206,7 +1206,7 @@ static mrb_value
 str_escape(mrb_state *mrb, mrb_value str, mrb_bool inspect)
 {
   const char *p, *pend;
-  char buf[4];  /* `\x??` or UTF-8 character */
+  char buf[7];  /* `\x??` or UTF-8 character */
   mrb_value result = mrb_str_new_lit(mrb, "\"");
 #ifdef MRB_UTF8_STRING
   uint32_t ascii_flag = MRB_STR_ASCII;
```

But I don't know what that means, so I suggest this PR method.

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -1219,12 +1219,7 @@ str_escape(mrb_state *mrb, mrb_value str, mrb_bool inspect)
     if (inspect) {
       mrb_int clen = mrb_utf8len(p, pend);
       if (clen > 1) {
-        mrb_int i;
-
-        for (i=0; i<clen; i++) {
-          buf[i] = p[i];
-        }
-        mrb_str_cat(mrb, result, buf, clen);
+        mrb_str_cat(mrb, result, p, clen);
         p += clen-1;
         ascii_flag = 0;
         continue;


matz:
Ah, maximum value from `mrb_utf8len()` (which is the biggest elements in `mrb_utf8len_table`) is `4`. So gcc detected and warned. I'm impressed how smart gcc is.
And copying a substring to `buf` is meaningless. I accept the PR. Thank you!

dearblue:
I think the warning is valid if `mrb_utf8len()` returns 5 or 6, but it doesn't seem so.
After a little more checking, changing the type of `clen` from `mrb_int` to `size_t` also stopped the warning.

At this point I wondered about gcc's behavior, so I checked for bugs and found a similar report.
<https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107476#c0>

Others can be traced through "Depends on" at <https://gcc.gnu.org/bugzilla/show_bug.cgi?id=Wstringop-overflow>.

This situation seems to be the same for gcc-13.0 as for gcc-12.2.
In addition, it seems that `-Wstringop-overflow` is included in `-Wall` since gcc-12.

:smile_cat: At any rate, I am happy to say that we have found areas for improvement.

PR #5849, opened 2022-11-05
Using `mrb_realloc()` with `stack_extend_alloc()`

dearblue:
Raising a `SystemStackError` exception in an out-of-memory situation will generate backtrace information. This can eventually lead to a `NoMemoryError` exception, and this process flow is completely undesirable.

In effect, this change means that it will include a revert of commit 0dbb9e6e411572b0c37e7be723d6dc3895387f96.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -177,10 +177,7 @@ stack_extend_alloc(mrb_state *mrb, mrb_int room)
     size += room;
 #endif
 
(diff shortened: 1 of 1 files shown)


matz:
So do you mean it's OK to cancel #5307?

dearblue:
> So do you mean it's OK to cancel #5307?

Yes, I will cancel.
#5307 was trying to take advantage of the original code `if (newstack == NULL)`.
However, in practice, I think it is preferable to raise the `NoMemoryError` exception as is.

PR #5854, opened 2022-11-13
gha: replace `Ubuntu 18.04` with `Ubuntu 22.04` in `build` workflow

jbampton:
https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,8 +6,8 @@ permissions:
   contents: read
 
 jobs:
-  Ubuntu-1804-gcc:
-    runs-on: ubuntu-18.04
+  Ubuntu-2204-gcc:
+    runs-on: ubuntu-22.04
     timeout-minutes: 10
     env:
       MRUBY_CONFIG: ci/gcc-clang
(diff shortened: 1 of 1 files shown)


matz:
Probably, we should replace ubuntu-2004 by ubuntu22-04.

dearblue:
@matz,
Isn't it Ubuntu 18.04 that replaces Ubuntu 22.04, not Ubuntu 20.04?
https://github.com/actions/runner-images/issues/6002

matz:
@dearblue I thought we picked ubuntu-1804 for some compatibility reason. But `git blame` told us that last time we updated, we have replaced 1604 by 1804. So I think it's OK to replace 1804.

@jbampton Sorry for the inconvenience, but could you replace 1804 instead of 2004, please?

dearblue:
🐈‍⬛ This is off-topic.

> I thought we picked ubuntu-1804 for some compatibility reason.

By "compatibility reasons" do you mean past GCC or Ruby (rake)?
If limited to GCC, gcc-4.8 may be available on Ubuntu 22.04.
\# I may do something after this PR trend is determined.

PR #5856, opened 2022-11-19
Updates documentation for `mrb_fiber_resume()` and `mrb_fiber_yield()`

dearblue:
- Added note on `mrb_fiber_resume()` and `mrb_fiber_yield()`. Removed comments from C files that were not documented by yard and doxygen instead.
- Leads to use of `mrb_fiber_resume()` instead of `Fiber#resume` in C.
- Leads to use of `mrb_fiber_yield()` instead of `Fiber.yield` in C.

### DIFF
--- a/mrbgems/mruby-fiber/src/fiber.c
+++ b/mrbgems/mruby-fiber/src/fiber.c
@@ -259,6 +259,9 @@ fiber_switch(mrb_state *mrb, mrb_value self, mrb_int len, const mrb_value *a, mr
  *  to the next <code>Fiber.yield</code> statement inside the fiber's block
  *  or to the block value if it runs to completion without any
  *  <code>Fiber.yield</code>
+ *
+ *  This method cannot be called from C using <code>mrb_funcall()</code>.
+ *  Use <code>mrb_fiber_resume()</code> function instead.
  */
 static mrb_value
 fiber_resume(mrb_state *mrb, mrb_value self)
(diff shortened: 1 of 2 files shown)


mrbgems/mruby-fiber/src/fiber.c
```diff
@@ -259,6 +259,9 @@ fiber_switch(mrb_state *mrb, mrb_value self, mrb_int len, const mrb_value *a, mr
  *  to the next <code>Fiber.yield</code> statement inside the fiber's block
  *  or to the block value if it runs to completion without any
  *  <code>Fiber.yield</code>
+ *
+ *  This method cannot be called from C.
```
matz, line 263: Is this comment needed/useful? It's a static function that never be able to be called from outside.

dearblue: This will be a comment on a Ruby method.
So I believe it will be useful.

matz: Probably better to mention `mrb_funcall` in the comment.

dearblue: Indeed it is. I will update it.

dearblue:
For reference, comments on `Fiber.yield` cannot be documented by `rake doc:api` (`mrbdoc`).
We need to wait for [yard-mruby](https://rubygems.org/gems/yard-mruby) to be updated.
This is because `mrb_define_class_method()` is not supported.

PR #5857, opened 2022-11-19
Avoid `abort()` when `FiberError` occurs in `mrb_fiber_resume()`

dearblue:
However, this may not be the case depending on where the `FiberError` raises.

### DIFF
--- a/mrbgems/mruby-fiber/src/fiber.c
+++ b/mrbgems/mruby-fiber/src/fiber.c
@@ -1,6 +1,7 @@
 #include <mruby.h>
 #include <mruby/array.h>
 #include <mruby/class.h>
+#include <mruby/error.h>
 #include <mruby/proc.h>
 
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-fiber/src/fiber.c
```diff
@@ -173,6 +174,22 @@ fiber_switch_context(mrb_state *mrb, struct mrb_context *c)
   mrb->c = c;
 }
 
+static mrb_value
+fiber_error_static(mrb_state *mrb, const char *mesg)
```
matz, line 178: I, as an insider, can imagine this `_static` means "calls mrb_str_new_static() inside", but I think it's too much to expect for others. Probably we should remove `_static` and add a comment that "`mesg` should be a literal string" or something similar.

dearblue: :black_cat: Surely, three months from now I will be doing something.
I will fix it.

PR #5858, opened 2022-11-20
Print errors by calling `mrb_print_backtrace()` without backtrace

dearblue:
- In case of `NoMemoryError` exceptions, the error message is now printed directly.
- Replaced `mrb_p()` used by #4250 with `mrb_print_error()`.

  ref. squashed commit f1523d24042ca3416dc5b9be7b3fc220ddaed896
  ref. subcommit da7d7f881bbbad9988a3a2b7bad8f2b72ff06bc6
  ref. subcommit d9c7b6be6eb54630b64eea5c35be241e551676e5

---

I don't remember for sure, but I think I used `mrb_p()` in #4250 because it sometimes didn't output and that was also the case before the change in commit da7d7f881bbbad9988a3a2b7bad8f2b72ff06bc6.

### DIFF
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -100,7 +100,7 @@ static void
 print_backtrace(mrb_state *mrb, struct RObject *exc, struct RArray *backtrace)
 {
   mrb_int i;
-  mrb_int n = ARY_LEN(backtrace);
+  mrb_int n = (backtrace ? ARY_LEN(backtrace) : 0);
   mrb_value *loc, mesg;
 
   if (n != 0) {
@@ -117,9 +117,19 @@ print_backtrace(mrb_state *mrb, struct RObject *exc, struct RArray *backtrace)
       fprintf(stderr, "%.*s: ", (int)RSTRING_LEN(*loc), RSTRING_PTR(*loc));
     }
   }
-  mesg = mrb_exc_inspect(mrb, mrb_obj_value(exc));
-  fwrite(RSTRING_PTR(mesg), RSTRING_LEN(mesg), 1, stderr);
-  fputc('\n', stderr);
+  else {
+    fprintf(stderr, "%s:0: ", "(unknown)");
+  }
+
+  if (exc == mrb->nomem_err) {
+    static const char nomem[] = "Out of memory (NoMemoryError)\n";
+    fwrite(nomem, sizeof(nomem) - 1, 1, stderr);
+  }
+  else {
+    mesg = mrb_exc_inspect(mrb, mrb_obj_value(exc));
+    fwrite(RSTRING_PTR(mesg), RSTRING_LEN(mesg), 1, stderr);
+    fputc('\n', stderr);
+  }
 }
 
 /* mrb_print_backtrace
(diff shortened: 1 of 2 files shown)


dearblue:
- example c file

    ```c
    #include <mruby.h>
    #include <mruby/compile.h>
    #include <stdio.h>
    #include <stdlib.h>

    #define LOAD_LIT(MRB, ...) mrb_load_string(MRB, #__VA_ARGS__)

    int
    main(int argc, char *argv[])
    {
      mrb_state *mrb = mrb_open();
      mrb_value fiber = LOAD_LIT(mrb,
        Fiber.new do \n
          5.times { |i| Fiber.yield(i) } \n
          0 / 0 \n
        end
      );

      for (int i = 0; i < 9; i++) {
        mrb_value args[] = { mrb_fixnum_value(i) };
        mrb_value ret = mrb_fiber_resume(mrb, fiber, 1, args);
        if (mrb->exc) {
          fprintf(stderr, "[%d:ERR] ", i);
          mrb_print_error(mrb);
        } else {
          fprintf(stderr, "[%d:RET] ", i);
          mrb_p(mrb, ret);
        }
      }

      mrb_close(mrb);
      fprintf(stderr, "[SHUTDOWN]\n");

      return 0;
    }
    ```

  - currently master (3a0e2255bd43fd46c72091dfbf197644df639853)

    ```
    [0:RET] 0
    [1:RET] 1
    [2:RET] 2
    [3:RET] 3
    [4:RET] 4
    [5:ERR] (unknown):0: divided by 0 (ZeroDivisionError)
    [6:ERR] [7:ERR] [8:ERR] [SHUTDOWN]
    ```

  - this PR

    ```
    [0:RET] 0
    [1:RET] 1
    [2:RET] 2
    [3:RET] 3
    [4:RET] 4
    [5:ERR] (unknown):0: divided by 0 (ZeroDivisionError)
    [6:ERR] (unknown):0: resuming dead fiber (FiberError)
    [7:ERR] (unknown):0: resuming dead fiber (FiberError)
    [8:ERR] (unknown):0: resuming dead fiber (FiberError)
    [SHUTDOWN]
    ```

src/backtrace.c
```diff
@@ -117,9 +117,19 @@ print_backtrace(mrb_state *mrb, struct RObject *exc, struct RArray *backtrace)
       fprintf(stderr, "%.*s: ", (int)RSTRING_LEN(*loc), RSTRING_PTR(*loc));
     }
   }
-  mesg = mrb_exc_inspect(mrb, mrb_obj_value(exc));
-  fwrite(RSTRING_PTR(mesg), RSTRING_LEN(mesg), 1, stderr);
-  fputc('\n', stderr);
+  else {
+    fprintf(stderr, "%s:0: ", "(unknown)");
```
matz, line 121: Do we need `fprintf()` here? It's a static string.

src/backtrace.c
```diff
@@ -117,9 +117,19 @@ print_backtrace(mrb_state *mrb, struct RObject *exc, struct RArray *backtrace)
...
+  }
+
+  if (exc == mrb->nomem_err) {
+    static const char nomem[] = "Out of memory (NoMemoryError)\n";
+    fwrite(nomem, sizeof(nomem) - 1, 1, stderr);
```
matz, line 126: I prefer no spaces around `-`. Just my preference.

src/backtrace.c
matz: Nevermind. I will work on removing `fprintf()` later.

PR #5859, opened 2022-11-20
Add task `rake doc:update-opcode.md`

dearblue:
Allow to update `doc/internal/opcode.md` file mechanically from `include/mruby/ops.h` file.

At the same time, the `doc/internal/opcode.md` file has been updated. Consecutive hyphens at the end of the table were removed as they would have created noise as markdown flavours.


NaokiM03:
Excellent improvement. I was concerned about the discrepancy between `send` and `call` in the `OP_SEND` instruction.

dearblue:
:scream_cat: I checked [opcode.md](https://github.com/mruby/mruby/blob/16ed467/doc/internal/opcode.md) again on GitHub and the `OP_SEND` and `OP_SSEND` comments are broken.
I will fix it later...

:thinking: And `OP_CALL` actually does "tail call", so I may change the comments on that too.

PR #5867, opened 2022-12-10
Combine gcc/clang jobs with matrix

dearblue:
The reason I limited myself to just `matrix.include` is that there are several environments that need to be specialized, and the whole description is easy to understand.
For example, the following description is not immediately comprehensible to me even when I write it myself. It is even more so if there are a lot of them.

```yaml
  matrix:
    os: ubuntu-2022.04, ubuntu-2020.04
    cc: gcc, clang
    include:
    - { cc: gcc, cxx: g++ }
    - { cc: clang, cxx: clang }
    - { os: ubuntu-latest, cc: gcc-4.8, cxx: g++-4.8, container: ubuntu:18.04 }
```

This is the same as:

```yaml
  matrix:
    include:
    - { os: ubuntu-2022.04, cc: gcc, cxx: g++ }
    - { os: ubuntu-2022.04, cc: clang, cxx: clang++ }
    - { os: ubuntu-2020.04, cc: gcc, cxx: g++ }
    - { os: ubuntu-2020.04, cc: clang, cxx: clang++ }
    - { os: ubuntu-latest, cc: gcc-4.8, cxx: g++-4.8, container: ubuntu:18.04 }
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,87 +6,26 @@ permissions:
   contents: read
 
 jobs:
-  Ubuntu-2204-gcc:
-    runs-on: ubuntu-22.04
+  GCC-CLANG:
+    name: "${{ matrix.os }}-${{ matrix.altname || matrix.cc }}"
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      max-parallel: 8
+      matrix:
+        include:
+          - {os: ubuntu-22.04, cc: gcc, cxx: g++}
+          - {os: ubuntu-22.04, cc: clang, cxx: clang++}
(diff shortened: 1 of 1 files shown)


dearblue:
Ruby-2.3 is available for Ubuntu-2016.04.
But I did not add it because it is already a build requirement for mruby to be Ruby-2.5 or higher.
\# However, it is a bit surprising that it even succeeds to `rake test`.

dearblue:
For now, let me summarize the whitespace that makes it look right...

matz:
I don't think the version number `2.5` is that important. It was the oldest maintained version at the time of the update (2021-02-09). Probably, we should update here regularly (or remove version number, since it's hard to find unsupported version e.g. `1.8`).

dearblue:
I understand about the version number.
I will now change this PR to just a commit that uses `matrix`.

PR #5868, opened 2022-12-11
Fix cloning gem when destination path contains spaces

boatrite:
This was how the clone command was previously being generated, for example:

```
git clone  --recursive --branch "master" --depth 1 https://github.com/mattn/mruby-onig-regexp.git /path/to mruby with spaces/mruby-3.1.0/build/repos/wasm/mruby-onig-regexp
```

Now it's changed so the path is quoted and will work when containing spaces:

```
git clone  --recursive --branch "master" --depth 1 https://github.com/mattn/mruby-onig-regexp.git "/path/to mruby with spaces/mruby-3.1.0/build/repos/wasm/mruby-onig-regexp"
```

Similar for the other commands.

Thank you!

### DIFF
--- a/lib/mruby/build/command.rb
+++ b/lib/mruby/build/command.rb
@@ -25,11 +25,7 @@ module MRuby
     end
 
     def shellquote(s)
-      if ENV['OS'] == 'Windows_NT'
-        "\"#{s}\""
-      else
-        "#{s}"
-      end
+      "\"#{s}\""
     end
 
     private


lib/mruby/build/command.rb
```diff
@@ -311,11 +311,11 @@ def run_reset_hard(dir, checksum_hash)
     end
 
     def commit_hash(dir)
-      `#{@command} --git-dir #{shellquote(dir +'/.git')} --work-tree #{shellquote(dir)} rev-parse --verify HEAD`.strip
+      `#{@command} --git-dir \"#{shellquote(dir +'/.git')}\" --work-tree \"#{shellquote(dir)}\" rev-parse --verify HEAD`.strip
```
dearblue, line 314: I think it is preferable to modify the `#shellquote` method to handle double-quoting.
The same applies to the `#current_branch` method below.

matz:
Could you squash the pull-request please?

matz:
LGTM

Issue #5869, opened 2022-12-12
codedump.c: OP_LOADINEG is displayed as OP_LOADI


NaokiM03:
Hi.

```ruby
-42
```

When the above code is compiled with the verbose flag, the following log is displayed.

![image](https://user-images.githubusercontent.com/37442712/206936809-cfe3a5fe-8f41-4feb-aee3-10617eff2ec6.png)

I expected the following.

![image](https://user-images.githubusercontent.com/37442712/206936934-964ef96e-9489-4bcb-a7b5-6eaba53525ed.png)

Is it intended that OP_LOADI is displayed ?

NaokiM03:
Ready to create a pull request.

NaokiM03:
> The `OP_` should be omitted in the display, so it becomes `LOADINEG` instead of `OP_LOADINEG`.

Oh, I'll fix it.

> depends on the person?

Yes, I asked before creating the pull request because I thought it might depend on the person.

matz:
It was intentional.

Since we wanted to save the sign bit for `OP_LOADI`, we separated `OP_LOADI` (positive 0-256 int) and `OP_LOADINEG` (negative -256 to 0). Recent assembler like x86_64 generates different byte sequence from the same instruction, so I thought it was OK to unify `OP_LOADI` and `OP_LOADINEG` in display.

The opinion may vary. If you think displaying exact (and correct) instruction name is crucial and less confusing, I respect your opinion.

NaokiM03:
Understood.

I was confused reading the bytecode, but few users may actually do that. If that is the case, it may be fine as is.
(Maybe if someone else thinks the same thing, they will see this isssue and understand.)

matz:
After consideration, I decided that it's better to disclose `OP_LOADINEG`, where we don't hide `OP_LOADI_1` etc.

matz:
@NaokiM03 To keep your name as a contributor, do you want to submit a pull-request for this issue?

NaokiM03:
@matz 
Created.
https://github.com/mruby/mruby/pull/5870

As you said, considering the current status of `OP_LOADI_x`, I thought it would be better to output the exact OPCODE. Thank you very much.

NaokiM03:
Now that the question has been answered, I will close this isssue. Thank you.

PR #5870, opened 2022-12-14
codedump.c: Use OP_LOADINEG instead of OP_LOADI #5869

NaokiM03:
This is a modification of the following issue.
https://github.com/mruby/mruby/issues/5869

image

### DIFF
--- a/src/codedump.c
+++ b/src/codedump.c
@@ -195,7 +195,7 @@ codedump(mrb_state *mrb, const mrb_irep *irep)
       print_lv_a(mrb, irep, a);
       break;
     CASE(OP_LOADINEG, BB):
-      printf("LOADI\tR%d\t-%d\t", a, b);
+      printf("LOADINEG\tR%d\t%d\t", a, b);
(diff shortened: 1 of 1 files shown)


matz:
Could you update the PR to print `-42` instead of `42`?

NaokiM03:
@matz 
I am concerned that the bytecode is 42, and that changing it may cause confusion.
Looking at `R(a) = mrb_int(-b)` in the documentation, I also think that `b` should be `42`.

What is the reason for wanting to `-42` ?

matz:
Because the instruction load a **negative** number and I want to display the **intention** of the instruction, rather than actual operand from the byte sequence.

NaokiM03:
@matz 
I think it is clear that `LOADINEG` is an instruction to load a negative integer...

Hmmm, sorry, I don't want to make a commit that is not what I intended, so I am closing this pull request.
Thank you for your time and for taking the time to answer my questions.

matz:
OK. Sorry for asking too much.

PR #5871, opened 2022-12-25
Fix typo in doc/limitations.md

ukrainskiysergey:
Since redefinition of "nil?" is possible in ruby, ruby's output should be "truthy".
And since redefinition of "nil?" is not possible in mruby, mruby's output should be "falsy".
It was the reversed.


matz:
Thank you!

PR #5879, opened 2022-12-26
Add `generate_active_gems_txt` to `rake --tasks`

jbampton:


dearblue:
This task is for remembering dependencies during the build process.
I don't think it is also intended for developers, let alone users.
Therefore, I am wondering if `desc` is really needed.

I have checked `git blame` to see how this task was introduced, and I include it here for reference.
Introduced by #5036 and updated by #5267.

PR #5902, opened 2023-01-05
Replace `MRB_NO_DIRECT_THREADING` with `MRB_USE_VM_SWITCH_DISPATCH`

dearblue:
The correct technique now is "Token Threading".
However, the new name was chosen because it was felt that if a different technique was added in the future, there would be no need to replace it.


matz:
I am not sure how proposed wording is better. Could you explain?

dearblue:
:black_cat: My only source of information is https://ja.wikipedia.org/wiki/スレッデッドコード, which may be suspect.

  - The current implementation is based on "Token threading".
  - "Direct threading" is not "Token threading".
  - "Token threading" is not "Direct threading".
  - "Threaded code" includes "Token threading", "Direct threading", "Huffman threading" and etc..

dearblue:
I fixed the conflicts and rebased.

matz:
I understand the wording (and I am sure it's valid, since Prof. Ertl mentioned). But since we see “switch threading” in the [reference](http://www.complang.tuwien.ac.at/forth/threaded-code.html), so if used precisely, the term threaded code itself is meaningless. Let me think about it for a while.

dearblue:
Oh, "Switch threading".
Hmmm, I'll leave this PR open for now, but switch to draft.

dearblue:
How about `MRB_USE_SWITCH_THREADING`?
Or is `MRB_USE_VM_SWITCH_DISPATCH` more understandable?

matz:
I prefer the term "switch dispatch" to "switch threading", for not using direct goto.

dearblue:
Thanks for your comment.
Replaced with `MRB_USE_VM_SWITCH_DISPATCH`.

And `DIRECT_THREADED` has been merged into `MRB_USE_VM_SWITCH_DISPATCH`.
So, please note that `#ifdef` and `#ifndef` have been replaced by the reverse.

PR #5904, opened 2023-01-09
Call `mrb_env_unshare()` in `mrb_top_run()` if necessary

dearblue:
This is to keep the local variables of the previously created blocks consistent in case the `mrbc_context` passed to `mrb_load_exec()` is `NULL` or different. Switching between `mrbc_context` pointers that are non `NULL` can be done safely by calling `mrbc_cleanup_local_variables()`.

Before this patch, the result of the following code is not as expected.

```console
% cat loadstr.c
#include <mruby.h>
#include <mruby/compile.h>

int
main(int argc, char *argv[])
{
  mrb_state *mrb = mrb_open();

  mrb_load_string(
      mrb,
      "(a, b, c, d, e, f, g) = [1, 2, 3, 4, 5, 6, 7] \n"
      "$lambda = -> { p [a, b, c, d, e, f, g] }");
  mrb_load_string(mrb, "$lambda.call");

  mrb_close(mrb);

  return 0;
}

% $(bin/mruby-config --cc --cflags --ldflags) loadstr.c $(bin/mruby-config --libs) && ./a.out
[main, nil, nil, main, nil, nil, main]
```

Also, since `mrb_env_unshare()` was not used before, the internal stack of simply detached `env` objects could show invalid addresses by `stack_extend()`. ref. https://github.com/kou/mruby-pp/commit/ef5951aca870183d8767cb61f6414240988ca35e

### DIFF
--- a/mrbgems/mruby-compiler/core/parse.y
+++ b/mrbgems/mruby-compiler/core/parse.y
@@ -6764,6 +6764,7 @@ mrbc_cleanup_local_variables(mrb_state *mrb, mrbc_context *c)
     c->syms = NULL;
     c->slen = 0;
   }
+  c->keep_lv = FALSE;
 }
 
 MRB_API void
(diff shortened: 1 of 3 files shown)


dearblue:
Additional note:
Strictly speaking, if the `slen` of different `mrbc_context` pointers are the same, the call to `mrbc_cleanup_local_variables()` may not be necessary, but it would only complicate the explanation.

PR #5928, opened 2023-02-12
Improved `rake install`

dearblue:
Summary:

  - ***I have done very little checking in the Windows environment. Someone may need to modify it.***
  - The `tasks/install.rake` file will be added.
  - Copy all header files to the build directory so that `bin/mruby-config` and `lib/libmruby.mak` files do not reference the original header files.
  - `rake install:full` will also install build targets other than "host".
  - `MRuby::Build#install_prefix = dir` will change the installation directory.
    The default is `<PREFIX>/` directory for "host" and `<PREFIX>/mruby/<build-name>/` directory for anything other than "host".
  - If the environment variable `DESTDIR` is set, it is referenced as the prefix of the actual write destination.
  - Text added to `doc/guides/compile.md` has been translated from Japanese to English by DeepL.


dearblue:
Fixed https://github.com/mruby/mruby/actions/runs/4156525680/jobs/7190371416

The variable `MRUBY_PACKAGE_DIR` is now added to `bin/mruby-config` by the `mruby-bin-config/mrbgem.rake` file.

PR #5938, opened 2023-02-28
Reorganize gem init final

dearblue:
The point is:

  - Don't call `exit()` when an exception occurs in `mrblib/*.rb`.
  - Ignore exceptions raised by gems finalizer.

### DIFF
--- a/tasks/mrbgems.rake
+++ b/tasks/mrbgems.rake
@@ -17,8 +17,8 @@ MRuby.each_target do
           s << "void GENERATED_TMP_mrb_#{g.funcname}_gem_init(mrb_state*);\n" \
                "void GENERATED_TMP_mrb_#{g.funcname}_gem_final(mrb_state*);\n"
         end
-        gem_init_calls = gem_func_gems.each_with_object('') do |g, s|
-          s << "  GENERATED_TMP_mrb_#{g.funcname}_gem_init(mrb);\n"
+        gem_funcs = gem_func_gems.each_with_object('') do |g, s|
+          s << "  { GENERATED_TMP_mrb_#{g.funcname}_gem_init },\n"
         end
         gem_final_calls = gem_func_gems.reverse_each.with_object('') do |g, s|
           s << "  GENERATED_TMP_mrb_#{g.funcname}_gem_final(mrb);\n"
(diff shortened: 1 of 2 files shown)


tasks/mrbgems.rake
```diff
@@ -34,20 +31,42 @@ MRuby.each_target do
...
         f.puts %Q[void]
         f.puts %Q[mrb_init_mrbgems(mrb_state *mrb) {]
-        f.write gem_init_calls
-        f.puts %Q[  mrb_state_atexit(mrb, mrb_final_mrbgems);] unless gem_final_calls.empty?
+        f.puts %Q[  int ai = mrb_gc_arena_save(mrb);]
+        f.puts %Q[  for (int i = 0; i < NUM_GEMS; i++) {]
+        f.puts %Q[    gem_funcs[i].init(mrb);]
+        f.puts %Q[    struct REnv *e = mrb_vm_ci_env(mrb->c->cibase);]
+        f.puts %Q[    mrb_vm_ci_env_clear(mrb->c->cibase);]
+        f.puts %Q[    mrb_env_unshare(mrb, e, FALSE);]
```
matz, line 75: After #5945 merged, we should remove `mrb_env_unshare()` here, do we?

tasks/mrbgems.rake
```diff
@@ -34,20 +31,42 @@ MRuby.each_target do
...
-          f.puts %Q[static void]
-          f.puts %Q[mrb_final_mrbgems(mrb_state *mrb) {]
-          f.write gem_final_calls
-          f.puts %Q[}]
-        end
+        f.puts %Q[]
+        f.puts %Q[static const struct {]
+        f.puts %Q[  void (*init)(mrb_state *);]
+        f.puts %Q[  void (*final)(mrb_state *);]
+        f.puts %Q[} gem_funcs[] = {]
```
matz, line 42: Do we need to pack init/finish functions in an array? What is the benefit?

tasks/mrbgems.rake
dearblue: Yup, I've been waiting your cue!

tasks/mrbgems.rake
dearblue: Without the function array, the function would be expanded directly, and error handling would have to be written each time as well.

dearblue: The final form is the same, but I would change the form at the commit level.

dearblue: Changes have been completed.
I also changed the commit message.

Issue #5932, opened 2023-02-24
How to create a block using the C API? mrb_yield keeps crashing!


neet93:
I'm trying to create a block using the C API, but I keep getting a segfault. This is the code below:

```
mrb_value proc_test(mrb_state *mrb, mrb_value self) {
        return mrb_false_value();                                            
}

struct RProc *p = mrb_proc_new_cfunc(mrb, proc_test);

mrb_value b = mrb_obj_value(p);

mrb_yield(mrb, b, mrb_nil_value());
```
The problem seems to the variable `mrb_value b`. I have no idea why!

Even when I use `mrb_p` trying to see if I'm getting the expected object value it just crashes! What's going on here?

Valgrind gives me this output in regard to the error:

```
Invalid read of size 8
0x12DE5A: mrb_yield.cold (vm.c:999)
0x130C45: data_manager_parse (MRB_DataManager.c:54)
0x15A42B: mrb_vm_exec (vm.c:1792)
0x162CF5: mrb_vm_run (vm.c:1282)
0x1630AD: mrb_run (vm.c:3100)
0x1630AD: mrb_funcall_with_block (vm.c:681)
0x1636E9: mrb_funcall_with_block (vm.c:637)
0x163AAD: mrb_funcall_argv (vm.c:692)
0x163AAD: mrb_funcall (vm.c:484)
0x1307C5: RunGame (CORE_Main.c:142)
0x1308FB: CORE_Main (CORE_Main.c:45)
0x12E068: main (Main.c:17)
Address 0x18 is not stack'd, malloc'd or (recently) free'd
```

Why is it an invalid read?? And what is mrb_yield.cold??

Can I get an example of how to create a block using the C API?

neet93:
Solved it!

dearblue:
I have come to the conclusion that this issue should be addressed.

This is because functions defined as methods can cause `SIGSEGV` if they use `mrb_yield()` or `mrb_yield_argv()` to call block arguments.
The block object to be given to such a method is created by `mrb_proc_new_cfunc()` as in this issue.

The place where `SIGSEGV` occurs is `MRB_PROC_ENV(p)->stack[0]` in the `mrb_yield()` function.
This is because `MRB_PROC_ENV(p)` is `NULL` for block objects created by `mrb_proc_new_cfunc()`.

The following is an example for reference:

```console
% cat 5932a.c
#include <mruby.h>
#include <mruby/compile.h>
#include <mruby/proc.h>
[...]

PR #5951, opened 2023-03-14
Safely retrieve `self` from proc in `mrb_yield()`

dearblue:
Safely retrieve `self` from proc in `mrb_yield()`

Passing a proc generated by the following method to `mrb_yield()` or `mrb_yield_argv()` no longer causes `SIGSEGV`:

- C functions made into proc objects with `mrb_proc_new_cfunc()`.
- A proc object which is the top level of a program generated by `mrb_load_string_cxt()` etc. with `mrbc_context::no_exec` enabled.

See also: #5932

### DIFF
--- a/src/proc.c
+++ b/src/proc.c
@@ -193,6 +193,29 @@ mrb_proc_cfunc_env_get(mrb_state *mrb, mrb_int idx)
   return e->stack[idx];
 }
 
+mrb_value
+mrb_proc_get_self(mrb_state *mrb, struct RProc *p, struct RClass **target_class_p)
+{
+  if (MRB_PROC_CFUNC_P(p)) {
+    *target_class_p = mrb->object_class;
+    return mrb_nil_value();
+  }
+  else {
+    struct REnv *e = p->e.env;
+
+    if (!e || e->tt != MRB_TT_ENV) {
+      *target_class_p = mrb->object_class;
+      return mrb_top_self(mrb);
+    }
+    else if (MRB_ENV_LEN(e) < 1) {
+      mrb_raise(mrb, E_ARGUMENT_ERROR, "self is lost (probably ran out of memory when the block became independent)");
+    }
+
+    *target_class_p = e->c;
+    return e->stack[0];
+  }
+}
+
 void
 mrb_proc_copy(mrb_state *mrb, struct RProc *a, struct RProc *b)
 {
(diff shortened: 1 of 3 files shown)


src/proc.c
```diff
@@ -193,6 +193,27 @@ mrb_proc_cfunc_env_get(mrb_state *mrb, mrb_int idx)
...
+  }
+  else {
+    struct REnv *e = p->e.env;
+
+    if (!e || e->tt != MRB_TT_ENV || MRB_ENV_LEN(e) < 1) {
+       /*
+        * Probably created by mrb_load_irep(), mrb_load_string() or etc.,
+        * or occurred a out of memory in mrb_env_unshare().
+        */
+      mrb_raise(mrb, E_ARGUMENT_ERROR, "self is lost");
```
matz, line 210: How about returning `mrb_top_self` instead of raising an exception?

dearblue: I am not very positive.

I'm concerned that if `mrb_top_self` is used, there will be a mismatch with the `self` that the user expects, and this will cause confusion.
Is it really appropriate to return `mrb_top_self`?

matz: I think so. As far as I see in the example in [#5932](https://github.com/mruby/mruby/issues/5932#issuecomment-1468129057), the issue happens with top-level proc. So the natural assumption for the receiver should be the toplevel self for **most of the cases**

On the other hand, if it raises exception, the user would be confused by not knowing how to handle this kind of error.

dearblue: I understand. Let's change it that way.

However, I would like to make an exception if `mrb_env_unshare()` encounters out of memory.
I expect this can be detected by `MRB_ENV_LEN(env) == 0`.

Relatedly, I would also allow the target class to be retrieved by `mrb_proc_get_self()`.

matz: Thank you!

src/proc.c
```diff
@@ -193,6 +193,30 @@ mrb_proc_cfunc_env_get(mrb_state *mrb, mrb_int idx)
...
+  if (MRB_PROC_CFUNC_P(p)) {
+    *target_class_p = mrb->object_class;
+    return mrb_nil_value();
+  }
+  else {
+    struct REnv *e = p->e.env;
+
+    if (!e || e->tt != MRB_TT_ENV) {
+      mrb_value self = mrb_top_self(mrb);
+      *target_class_p = mrb_class(mrb, self);
```
matz, line 208: In this case, top-level target class should be `mrb->object_class` I think.

dearblue:
Right, I agree.
Changed.

PR #5977, opened 2023-04-09
Make `MRuby::Build#enable_debug` compiler flag setting abstract

dearblue:
Previously, compiler flags were only added for GCC or similar. The situation has not changed, but it has become easier to improve.

I expect `MRuby::Command::Compiler#setup_debug` to be defined as a singleton method inside the block given to `MRuby::Toolchain.new`.


dearblue:
The method name of `MRuby::Command::Compiler#setup_debug` was `enable_debug` in the initial proposal.
However, I chose `setup_debug` because I thought it would be confused with `MRuby::Build#enable_debug`.

PR #5986, opened 2023-04-15
Fix bugs in `String#bytesplice`

dearblue:
Tests for overflow fail, which will be resolved by #5985.


dearblue:
Missing note: I hope #5988 will be merged in advance because the test fails with the `MRB_INT32 + MRB_USE_BIGINT` configuration.

PR #6013, opened 2023-05-26
Fix Android toolchain

CaptainJet:
The NDK no longer supports mips, plain armeabi, or gcc. Remove old build_config for armeabi.
Update minimum sdk.
Fix tooling paths and targets.


matz:
Thank you!

gsantner:
Hello @CaptainJet,

I tried to compile mruby using this new configuration. Unfortunately it errors with some missing library/include error regarding readline.h (indeed Android NDK doesn't contain a readline.h).

Was this also the case back then and is there any special prerequisites required for mruby on Android?

Call:
```
env MRUBY_CONFIG="android_arm64_v8a" ANDROID_NDK_HOME="<somepath>/android/sdk_linux64/ndk/23.1.7779620" rake
```

grafik

CaptainJet:
I generally recommend not compiling the executables like mirb when making the Android builds. I'm afraid I can't really help you debug it myself. Good luck though!

gsantner:
I'm actually more interested in getting ruby included so I can run some ruby code via C++/NDK on Android. Don't really need interactive in this case I guess 😄

gsantner:
@CaptainJet Thank you for the very fast response!

I'm actually more interested in getting ruby included so I can run some ruby code via C++/NDK on Android. Don't really need interactive in this case I guess 😄 

OK, so when I disable all additional `... command` at default.gembox then it compiles successfully.  

So isn't it wrong to specify default.gembox at the android configuration, as it is not compatible? I couldn't find any DISABLE_TOOLS=true environment flag either to disable this stuff so it could be built without manually touching the git state of mruby.

grafik

PR #6022, opened 2023-06-17
Build configuration for MS-DOS and DJGPP

chasonr:
This provides a build configuration for MS-DOS, cross-compiled on Linux using DJGPP.

* Most gems compile unmodified, with a few flags set in the build configuration.
* mruby-dir compiles only if chroot is disabled.
* mruby-io requires a few changes to avoid the libgen.h file, which DJGPP does not have, and to define fsuseconds_t.
* mruby-socket does not compile.
* mruby-cmath does not compile, because DJGPP does not provide complex.h. I am working on an alternative implementation that does not require complex.h.


matz:
Wow, impressive. Thank you!

PR #6024, opened 2023-06-18
Prepare stub functions for libmruby_core in core

dearblue:
The `MRuby::Build#libmruby_core_only_objs` accessor has been added for this purpose. It probably does not make sense for users to use this accessor.

And, the `mruby-bin-mrbc/src/stub.c` file has been moved to `mruby/src/coreonly/stub.c`. The object files generated from this file are passed to the newly created accessor.

### DIFF
--- a/lib/mruby/build.rb
+++ b/lib/mruby/build.rb
@@ -76,7 +76,8 @@ module MRuby
     include Rake::DSL
     include LoadGems
     attr_accessor :name, :bins, :exts, :file_separator, :build_dir, :gem_clone_dir, :defines, :libdir_name
-    attr_reader :products, :libmruby_core_objs, :libmruby_objs, :gems, :toolchains, :presym, :mrbc_build, :gem_dir_to_repo_url
+    attr_reader :products, :libmruby_core_only_objs, :libmruby_core_objs, :libmruby_objs
+    attr_reader :gems, :toolchains, :presym, :mrbc_build, :gem_dir_to_repo_url
 
     alias libmruby libmruby_objs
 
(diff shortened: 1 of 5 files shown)


matz:
I agree with the basic idea of this pull-request. But I don't like the directory name `coreonly` which role is not clear from the directory name.

dearblue:
Thanks for your comment.
:thinking: Hmmm, is `src/libmruby_core/` acceptable?

matz:
`src/nogem` maybe?

dearblue:
`MRB_NO_GEMS` may be defined in `libmruby.a` builds.

```ruby
# build configuration
MRuby::Build.new do
  toolchain
end
```

Also, I don't think `libmruby_core.a` is very good about `nogem` since it doesn't include `mruby/mrblib`.

However, if you think `src/nogem` is OK, I will update this patch accordingly.

matz:
I still feel unclear about what the term `libmruby_core_only_objs` means and what it is used for. Please explain a little more. Japanese is fine.

dearblue:
すでに `libmruby_core_objs` がありますが、これは `libmruby_objs` の一部としても使われています。
https://github.com/mruby/mruby/blob/bd668e4c9dc2244aa201addf95acbb8663596423/lib/mruby/build.rb#L121-L122

`libmruby_core_objs` を別の名前に変更する場合は互換性の面で少し気になったため、( (`libmruby_core` + `only`) + `objs`) という名前で提案しました。

matz:
Closing without merge — the cleanup is conceptually valid, but it's been blocked on directory naming since 2023 and the practical impact is narrow. The current per-binary stubs in `mrbgems/mruby-bin-mrbc/` already work for mrbc itself, and 3rd-party tools that link `libmruby_core.a` directly can copy a small stub.c just as easily as upstream maintaining one.

If a need surfaces (e.g., multiple in-tree tools rediscovering the same stubs), we can revisit with a single file like `src/core_stub.c` to avoid the directory-naming question. Thanks for the proposal.

PR #6032, opened 2023-07-29
Rake: update task clean to remove `bin` and `build` folders

jbampton:


dearblue:
The `#{MRUBY_ROOT}/build` and `{MRUBY_ROOT}/bin` directories, which are now hardcoded by this PR, can still be replaced.

  - Alternatives to the `#{MRUBY_ROOT}/build` directory can be specified with the `MRuby::Build#build_dir=` accessor.
  - The `{MRUBY_ROOT}/bin` directory can be replaced by the `INSTALL_DIR` environment variable.
    NOTE: There is no documentation for the `INSTALL_DIR` environment variable, so I believe it is a hidden feature.

Also, `#{MRUBY_ROOT}/build/repos/#{MRuby::Build#name}` contains a `git clone` of the third party mrbgems.
The git repository is also removed by `rake clean` with this PR, so people who are in the middle of modifying other people's code may have trouble with `rake clean`.
NOTE: The `MRuby::Build#gem_clone_dir=` accessor allows you to specify an alternative, but I believe it is a hidden feature that is not documented.

Therefore, I will create a PR to revert to.

In my opinion, if the content of this PR is needed, it would be better to define it as a different task than `rake clean` or `rake deep_clean`.

dearblue:
This is because the current `deep_clean` task only removes third-party GEMS directories related to the build target based on the "Build configuration file" at the time.
Unrelated GEMS directories will be kept intact (on a per-build-target basis).

matz:
The original meaning of `deep_clean` was “to clean everything generated during build”, not only in the last build, but in all builds. So it's natural to remove `bin` and `build` in the task, as default build target directories. I know some files may be generated outside those directories (according to configurations), but it's better than nothing.

Some software uses `distclean` task for the purpose, but I see no specific reason for separate tasks.

dearblue:
I understood. If that was the original purpose of `deep_clean`, I have no disagreement.

PR #6040, opened 2023-08-28
Convert to string before output in `puts`

dearblue:
For example, evaluating `puts nil` raised `NoMethodError` before.

### DIFF
--- a/mrbgems/mruby-print/mrblib/print.rb
+++ b/mrbgems/mruby-print/mrblib/print.rb
@@ -28,6 +28,7 @@ module Kernel
       if s.kind_of?(Array)
         puts(*s)
       else
+        s = s.to_s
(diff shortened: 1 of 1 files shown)


mrbgems/mruby-print/mrblib/print.rb
```diff
@@ -29,7 +29,7 @@ def puts(*args)
         puts(*s)
       else
         print s
-        print "\n" if (s[-1] != "\n")
+        print "\n" if !s.kind_of?(String) || s[-1] != "\n"
```
matz, line 32: How about putting `s = s.to_s` before calling first `print`?

dearblue: Changed.

PR #6056, opened 2023-09-18
Brush up `build_config/host-shared.rb` file

dearblue:
Summary:

  - VisualC++ does not supported.
  - `cc.flags` must maintain array objects.
  - Enable C++ exceptions for C++ friendly.

### DIFF
--- a/build_config/host-shared.rb
+++ b/build_config/host-shared.rb
@@ -1,23 +1,20 @@
+# NOTE: Currently, this configuration file does not support VisualC++!
+#       Your help is needed!
+
 MRuby::Build.new do |conf|
   # load specific toolchain settings
-
-  # Gets set by the VS command prompts.
-  if ENV['VisualStudioVersion'] || ENV['VSINSTALLDIR']
-    toolchain :visualcpp
-  else
-    toolchain :gcc
-  end
+  conf.toolchain
 
   # include the GEM box
   conf.gembox 'default'
 
   # C compiler settings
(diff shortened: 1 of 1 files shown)


dearblue:
I noticed that `bin/mruby` has a reference to `libmruby.so` as an absolute path when built with `rake MRUBY_CONFIG=host-shared`.
This means that if `rake install` is done, `bin/mruby` will need the original directory from which it was built when it is run.

Maybe I need to revisit some kind of fix related to "#6055" as well.
:thinking: Thinking about it for now.

Just in case: I think this PR is now fine to merge.

build_config/host-shared.rb
```diff
@@ -29,6 +26,9 @@
   # file separator
   # conf.file_separator = '/'
 
+  # for better compatibility with C++
+  conf.enable_cxx_exception
```
matz, line 30: I am concerned about the reliability of `enable_cxx_exception` (especially performance). Is it ok to turn it on all the time?

dearblue: I don't know for sure about reliability, so I commented out.
As for performance, I would like to find a way to improve it, since the degradation due to shared objectization is more noticeable.

The benchmark results are attached.
Settings can be found <https://gist.github.com/dearblue/8f4e58dd34b5c10a40e6bc605c32bf79#file-shared-benchmark>.
The run command is `ruby run-benchmark.rb shared.benchmark`.

bm-result-20231025T213440

Issue #6060, opened 2023-09-20
SEGFAULT Issue Related to Fiber Usage in ngx_mruby Development


pyama86:
mruby version:3.2.0

Hello. I am developing with ngx_mruby and I have encountered a SEGV issue related to Fiber and would like to consult with you about it.
```c
  mrb_value fiber_class = mrb_obj_value(mrb_class_get(mrb, "Fiber"));
  handler_proc = mrb_obj_value(mrb_closure_new(mrb, rproc->body.irep));
  if (mrb->exc) {
    ...
  }

  fiber_proc = (mrb_value *)ngx_palloc(r->pool, sizeof(mrb_value));
  *fiber_proc = mrb_funcall_with_block(mrb, fiber_class, mrb_intern_cstr(mrb, "new"), 0, NULL, handler_proc);
  ...

  handler_result = mrb_fiber_resume(mrb, *fiber_proc, 0, NULL);
```

A SEGV occurs when performing operations such as redirects in Nginx, moving back and forth between the Nginx context and the mruby context. The error appears at the following location.

```
Program received signal SIGSEGV, Segmentation fault.
0x00005555557a1ae3 in fiber_switch (mrb=0x555555ad01c0, self=..., len=0, a=0x0, resume=true, vmexec=true) at /home/pyama/src/github.com/matsumotory/ngx_mruby/mruby/mrbgems/mruby-fiber/src/fiber.c:260
260     value = c->stbase[0] = MRB_PROC_ENV(c->cibase->proc)->stack[0];
```

1. Here is the state during the first execution. At this point, c->cibase->proc->flags equals 1536.

```
com/matsumotory/ngx_mruby/src/http/ngx_http_mruby_module.c:841
841    if (mrb_test(ngx_mrb_start_fiber(r, state->mrb, code->proc, mrb_result))) {
(gdb) p *code
$40 = {code = {
  file = 0x555555b9bb20 "Nginx::Request.new.headers_out[\"Server\"] = \"global_ngx_mruby\"; Nginx.return Nginx::DECLINED",
  string = 0x555555b9bb20 "Nginx::Request.new.headers_out[\"Server\"] = \"global_ngx_mruby\"; Nginx.return Nginx::DECLINED"},
 code_type = NGX_MRB_CODE_TYPE_STRING, n = 0, cache = 1, proc = 0x555555b35500, ctx = 0x555555b8f0e0}


[...]

Issue #3056, opened 2015-12-23
calling `Fiber#resume` directly from C causes error


kazuho:
When `Fiber#resume` is called directly from C, the method returns the argument passed, without running any code on the fiber context.

Looking at the source code, it seems like that the fiber implementation is not handling such case.

In the current implementation, `fiber_switch` (called by `mrb_fiber_resume`) replaces the content (i.e. `mrb->c`) and returns, hoping that the VM will continue running the operations on the changed context.  This works fine if mrb_context_run is the caller of mrb_fiber_resume.

However in case `Fiber#resume` is called directly from C code, `mrb_fiber_resume` returns directly to `mrb_funcall_with_block` and then to the C code, without running the operations on the changed context.

Originally discussed in [this tweet](http://twitter.com/kazuho/status/679501838832373760).  Sorry for the lack of code for reproducing the issue.

matz:
Can you show us your code snippet?  It doesn't have to be reproducible.
I just want to know how you called #resume.

kazuho:
@matz Thank you for looking into the issue.

The diff is https://github.com/h2o/h2o/commit/3152e4c.  The commit removes the wrapper used to call `Fiber#resume`, and calls the method directly.

With the change, the mruby handler of H2O stops working correctly; when called for the first time, it returns the argument passed to `Fiber#resume`, succeeding calls end up in exceptions that say: `FiberError: can't cross C function boundary`.

Asmod4n:
Have you tried https://github.com/mruby/mruby/blob/master/mrbgems/mruby-fiber/src/fiber.c#L231 yet?

matz:
`mrb_fiber_resume` should work from C functions, but it currently does not.  I consider it's a bug, but it requires huge API change. Please use proc workaround for a while.

kazuho:
@matz Thank you for the answer.  Will use the proc workaround.

@Asmod4n Thank you for the suggestion.  Unfortunately as stated by @matz calling `mrb_fiber_resume` directly does not solve the issue.

> `fiber_switch` protects from calling Fiber#resume via mrb_funcall: 

FWIW, to be precise, the code does not protect `Fiber#resume` from being called through `mrb_funcall`; the guard exists to prevent fiber switches that involve a C stack.  For example if the call stack is C -> mruby -> C* -> ... `Fiber#resume`, it is impossible to swap the stack of mruby since the second call stack of C (marked with *) cannot be swapped.

[...]

PR #6063, opened 2023-09-25
Execution beyond the context of the fiber resume is required for some…

pyama86:
Hello. This PR is related to the following:
https://github.com/mruby/mruby/issues/6060

Originally, there was an issue in mruby where the stack would get corrupted when executing Fiber#resume in a C context, and as a workaround, there is the approach of wrapping Fiber execution in a Proc.
https://github.com/mruby/mruby/issues/3056

This workaround is used in [h2o](https://github.com/h2o/h2o) and [ngx_mruby](https://github.com/matsumotory/ngx_mrby).

However, with the addition of this commit included in mruby 3.2 and later, this workaround will no longer function.
https://github.com/mruby/mruby/pull/5782

My request is that until the issue with Fiber is resolved, I would like to request permission to continue using this workaround.
I would appreciate it if you could consider this.

### DIFF
--- a/mrbgems/mruby-fiber/src/fiber.c
+++ b/mrbgems/mruby-fiber/src/fiber.c
@@ -308,7 +308,6 @@ fiber_resume(mrb_state *mrb, mrb_value self)
   const mrb_value *a;
   mrb_int len;
 
-  fiber_check_cfunc(mrb, mrb->c);
   mrb_get_args(mrb, "*!", &a, &len);
   return fiber_switch(mrb, self, len, a, TRUE, FALSE);
 }


matz:
Is this still required after #6076?

dearblue:
The investigation is still in progress and the ngx_mruby test is not successful.

pyama86:
Yes, the following error continues to be output.
`can't cross C function boundary(FiberError)`

dearblue:
I was able to fix the remaining problem of switching to `mrb_fiber_resume()`.
I don't know about in production, but at least `nginx/test.sh` succeeds on my local FreeBSD environment, and [CI with GitHub Actions also succeeds](https://github.com/dearblue/ngx_mruby/actions/runs/6682425327).

Here is a summary of the results from my ngx_mruby experimental branch:.

  - The <https://github.com/mruby/mruby/commit/9ee7edc1ac02f95aefe6b8e2bbbe7c76237d5e36> patch is required for mruby 3.2.0.
  - The <https://github.com/mruby/mruby/commit/e7828bac4d93d8ef5243a4794e0648a20d41a747> patch is required when using `conf.enable_debug`.
  - The <https://github.com/dearblue/mruby-json/commit/ac8bdd20feb53f0fdf2fb766464aed8f6d438376> patch is required for mruby-json.

I haven't put all my thoughts together yet, but I plan to do the following

  - Split my experimental branch and propose some PR to ngx_mruby.
    <https://github.com/dearblue/ngx_mruby/commit/412f0a1af5891a428766adcf8dba9f1117210112>
  - Change `mrb_proc_new()` to `mrb_vm_proc_new()` and `mrb_closure_new()` to `mrb_vm_closure_new()`.
  - Add a new `mrb_proc_new()` with `MRB_API` modifier.
  - Add `mrb_fiber_new()` function.

pyama86:
That's great! I will test it on my end and contribute as needed.
I am very happy that this will eliminate Fiber's Hack from ngx_mruby.

dearblue:
I noticed that ngx_mruby was calling `Fiber#resume` inside `Fiber#resume`.
I have since tried and found that the current mruby no longer reproduces #5781 even when this PR is applied. However, the error is not displayed, so additional investigation is needed.

I suspect that the reason why it is no longer reproduced is due to a commit https://github.com/mruby/mruby/commit/a96a9f2c2bc64fbf7a5cd1371632c7fe13f3eec4 that changed the behavior of Fiber to that of "CRuby 3.0".

So I have come to the conclusion that we can use `mrb_funcall_id(mrb, fiber, MRB_SYM(resume), 0)` from the `main()` function without any problems at this time. As long as the native call stack allows it.
However, I don't think it can be mixed with `Fiber#transfer`, so I don't think this PR can be applied as is.

I need to do some more research, but I think the following restrictions need to be added regarding `Fiber#transfer`.

  - If `Fiber#trasnfer` is called from a fiber other than the root fiber, trace the calling fiber to make sure there is no C function call frame.
  - If `mrb_fiber_resume()` is called, check that the caller is not inside a direct or indirect `Fiber#transfer`.
  - If `Fiber#resume` is called inside a direct or indirect `Fiber#transfer`, check that the caller is Ruby.

pyama86:
Thank you for the improvements including ngx_mruby.
It seems that my issue is likely to be resolved, so I will close the PR.

dearblue:
For Notice.
I have send the proposed improvements as #6106.

PR #6097, opened 2023-11-05
Add a new `mrb_fiber_new()` with MRB_API

dearblue:


mrbgems/mruby-fiber/src/fiber.c
```diff
@@ -457,6 +460,18 @@ fiber_current(mrb_state *mrb, mrb_value self)
   return mrb_obj_value(mrb->c->fib);
 }
 
+MRB_API mrb_value
+mrb_fiber_new(mrb_state *mrb, const struct RProc *p)
+{
+  struct RClass *c = mrb_class_get_id(mrb, MRB_SYM(Fiber));
+  if (MRB_INSTANCE_TT(c) != MRB_TT_FIBER) {
```
matz, line 467: When `MRB_INSTANCE_TT(c) != MRB_TT_FIBER`, `MRB_OBJ_ALLOC` will cause `TypeError`.
Do we really need this check?

dearblue: Yes, I consider it necessary.

```console
% cat 1.c
#include <mruby.h>

int
main(int argc, char *argv[])
{
  mrb_state *mrb = mrb_open();
  struct RFiber *o = MRB_OBJ_ALLOC(mrb, MRB_TT_FIBER, mrb->object_class);
  mrb_p(mrb, mrb_obj_value(o));
  mrb_close(mrb);

  return 0;
}

% $(bin/mruby-config --cc --cflags --ldflags --libs) 1.c && ./a.out
#<Object:0x823fcf720>
```

This is because the instance type tag of the `Object` class is `MRB_TT_FALSE` and the `ttype` parameter of `mrb_obj_alloc()` is rarely checked in this case.
https://github.com/mruby/mruby/blob/9e50d67fe73e507627a4c1446d40a87a6c181983/src/gc.c#L477-L490

Or is it better to take the same approach as `Random`?
https://github.com/mruby/mruby/commit/9b9424fe3276811c805a09bc7948fd6958265563#diff-bf9e7692602cba54ba62bd71b76c439be9c9b71c6bb8d51ab996fc91cf1ee807

matz: I am going to review the condition in `mrb_obj_alloc()`.

Besides that, I don't think we should take `Random` approach. `Random` uses `TT_ISTRUCT` type that may be used by other classes. In contrast, `TT_FIBER` is only used by `Fiber` class.

matz: Ok, I will merge this PR as it is. Then I will try to update `mrb_obj_allocate`.

PR #6101, opened 2023-11-12
Reorganization of unwinding at `break` / `return`.

### LINKED ISSUE #6099: `ensure` section is not executed if the function exits via a return in a proc
In the following function:

```ruby
def outer
  puts "open"
  proc{ return 42 }.call
ensure
  puts "close"
end
```

the `ensure` block is not executed in this case, as it is with CRuby and JRuby.

Replacing the `proc` with just `return 42` behaves as expected.  So does exiting via a block argument, like this:

```ruby
def inner(&block)
  puts "open"
  return block.call
ensure
  puts "close"
end

def outer
  inner { return 42 }
end

outer()
```

dearblue:
The unwinding process that has been done at each point in time is now done in a merged block after looking for the last ci to `break` / `return`.

fix #6099

---

In my working branch, I am changing the parts related to `unwinding` and `RBreak`. This PR is a part of it. I will send PRs for other parts in stages.
https://github.com/mruby/mruby/compare/master...dearblue:mruby:wip/unwinding

The following benchmarks may be helpful to understand the impact on performance.
Please check `` host@(2) 2e881d329 Reorganization of unwinding at `break` / `return` `` for this PR.
bm-result-20231112T210909

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1062,13 +1062,8 @@ mrb_yield_cont(mrb_state *mrb, mrb_value b, mrb_value self, mrb_int argc, const 
 
 #define RBREAK_TAG_FOREACH(f) \
   f(RBREAK_TAG_BREAK, 0) \
-  f(RBREAK_TAG_BREAK_UPPER, 1) \
-  f(RBREAK_TAG_BREAK_INTARGET, 2) \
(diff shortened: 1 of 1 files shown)


matz:
Thank you.

PR #6103, opened 2023-11-23
RBreak remembers the CI location

dearblue:
It is now possible to specify return destination directly. This allows callinfo to distinguish between calls to the same proc object.

At the same time, the `Kernel#catch` method is adjusted. By removing the previously required double lambda object, the REnv object is no longer created as well.


dearblue:
For reference, I will upload the benchmark results including the [more advanced commits](https://github.com/mruby/mruby/compare/master...dearblue:mruby:PROBE/1) I showed last time.
At this point, I would like to send the PR up to ``(7) d08fe74cb Change `catch_handler_find()` argument``.
Anything beyond that needs to be explored to see if it is really valid.

bm-result-20231125T130730

dearblue:
Resolved conflicts caused by #6109.

Issue #6016, opened 2023-05-28
Support in Vcpkg


luodaoyi:
Can mruby support in [vcpkg](https://github.com/microsoft/vcpkg)?

rubyFeedback:
It may help laying down which steps would be necessary for that. What steps would be necessary for mruby to be available via vcpkg as well?

luodaoyi:
[https://learn.microsoft.com/zh-cn/vcpkg/examples/packaging-github-repo](https://learn.microsoft.com/zh-cn/vcpkg/examples/packaging-github-repos)
[https://learn.microsoft.com/zh-cn/vcpkg/examples/packaging-zipfiles](https://learn.microsoft.com/zh-cn/vcpkg/examples/packaging-zipfiles)

like this ?

> It may help laying down which steps would be necessary for that. What steps would be necessary for mruby to be available via vcpkg as well?

matz:
Thank you for the suggestion. Unfortunately, I don't know anything about vpkg. Actually, I have never programmed on Windows. If you (or anyone) send me a pull-request, I'd love to merge.

luodaoyi:
It's use cmake ...

dearblue:
Please note that I have never used vcpkg or written a cmake file, so this is all speculation.

As far as I can see, the vcpkg repository receives a pull request for build instructions for an arbitrary package.
The cmake file would be placed on the vcpkg side.
:crying_cat_face: No `ports/ruby`, so here's a link to python: https://github.com/microsoft/vcpkg/blob/2023.06.20/ports/python3/portfile.cmake

Also, it seems that cmake has a function to call external commands.
Therefore, it may be possible to use `rake DESTDIR=... install` from the vcpkg side.

In addition, it seems that vcpkg-specific patches for mruby can also be placed on the vcpkg side.

To sum up:

  - Probably no work needed in the mruby repository.
  - Probably someone interested will send a pull request to vcpkg (probably as `ports/mruby`).
  - Probably interested parties will update `ports/mruby` on the vcpkg side with a pull request.

luodaoyi:
>
> Please note that I have never used vcpkg or written a cmake file, so this is all speculation.
> 
> As far as I can see, the vcpkg repository receives a pull request for build instructions for an arbitrary package. The cmake file would be placed on the vcpkg side. 😿 No `ports/ruby`, so here's a link to python: https://github.com/microsoft/vcpkg/blob/2023.06.20/ports/python3/portfile.cmake
> 
[...]

PR #6133, opened 2023-12-28
Create source code packages by adding release tags

dearblue:
According to the [GitHub blog](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/), files by `git archive` are not guaranteed to have checksum persistence.
It is also stated that the archive file at this time is stable in bytes, but that this may change in the future.

---

Creating a tag in the format "x.y.z" will cause a build for release and the source code will be published in the form of a tarball.
If the tag name contains "-rc", it will be published as a "pre-release".
Release notes are taken from the contents of the `NEWS` file.

For GitHub Actions logs and previews, please follow the URL below (this repository will eventually be removed).

  - For version "192.168.0.1-rc"
      - https://github.com/dearblue/SANDBOX1/actions/runs/7348227811
      - https://github.com/dearblue/SANDBOX1/releases/tag/192.168.0.1-rc
  - For version "192.168.0.1"
      - https://github.com/dearblue/SANDBOX1/actions/runs/7348228299
      - https://github.com/dearblue/SANDBOX1/releases/tag/192.168.0.1

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,45 @@
+name: Release Builds
+
+on:
+  push:
+    tags:
+      - "*.*.*"
+
+permissions: read-all
+
+jobs:
+  source:
+    name: Source Code Releasing
+    permissions:
+      contents: write
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+    steps:
+      - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
+        uses: actions/checkout@v4
+      - name: Builds
+        id: builds
+        run: |
+          tag=$(echo "$GITHUB_REF" | sed 's@^refs/tags/@@')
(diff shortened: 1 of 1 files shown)


dearblue:
In the future, you may be able to do the following
However, this does not necessarily mean I will do it.

  - Files required as a package can be generated automatically in advance.
    For example, in the future, it will no longer be necessary to include "mrbgems/mruby-compiler/core/y.tab.c" in the repository.
  - In the future, even when `git submodules` are used to incorporate other repositories, the release package can be reliably included.
  - Automate the provision of pre-built binary packages.
    MSVC 32/64, mingw32, mingw32-x64, etc.?
  - It may be possible to provide source code packages that cannot be modified at build time but do not require Ruby.
    In this case, we might just generate a presym-related file and a "Makefile" or "CMakefile" file?
    This might be a solution for #6016.

dearblue:
The main change is that it no longer fires in the clone repository.

PR #6144, opened 2024-01-10
pre-commit: add hook `detect-aws-credentials`

jbampton:
https://github.com/pre-commit/pre-commit-hooks#detect-aws-credentials

Another check or test we can have.

"--allow-missing-credentials - Allow hook to pass when no credentials are detected."

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,6 +20,8 @@ repos:
       - id: check-shebang-scripts-are-executable
       - id: check-vcs-permalinks
       - id: check-yaml
+      - id: detect-aws-credentials
+        args: [--allow-missing-credentials]
       - id: detect-private-key
       - id: end-of-file-fixer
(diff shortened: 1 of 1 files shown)


matz:
The time for `pre-commit` checks is getting longer (especially prettier check). I think we have small chance to disclose AWS credentials in mruby repository. Persuade me, if you really want it.

jbampton:
The `CIFuzz / Fuzzing` check takes about 16 minutes and the `AppVeyor` build is also a long check.

The `pre-commit` check finishes in two minutes so it's not like its the last check to finish.  

`pre-commit` has a lot of checks / tests now about 24 checks and I think it is very useful. 

I think the more checks / tests we have the better. 

If you are concerned about how long `pre-commit` takes on your local machine you could `SKIP` checks.

For example:

`SKIP=prettier pre-commit run --all-files`

PR #6145, opened 2024-01-13
Enumerable#chunk implementation

hoshiumiarata:
This pull request implements the `chunk` method for `Enumerable`.
https://docs.ruby-lang.org/en/3.3/Enumerable.html#method-i-chunk
~It also adds `mruby-enumerator` to the dependencies of `mruby-enum-ext` (mainly for the use of `to_enum`, `Enumerator`).~


matz:
From the dependency and memory consumption point of view, I'd recommend adding `Array#chunk` method in the `mruby-enumerator` gem, just like `Array#zip` method.

hoshiumiarata:
@matz 
Thank you for review.
I moved `Array#chunk` to `mruby-enumerator`.

PR #6174, opened 2024-02-17
Fix `MRUBY_PACKAGE_DIR` in `mruby-config.bat`

### LINKED ISSUE #6156: '/LIBPATH' issue
I just discovered mruby, cloned it into `f:\MinGW32\src\Languages\mruby`,
built it with MSVC using `rake.bat --build-all --trace -v` and tested it a bit. 
Seems way cool!

But I discovered that the generated `bin/mruby-config.bat` outputs this for `--ldflags`:
```ruby
/NOLOGO /DEBUG /INCREMENTAL:NO /OPT:ICF /OPT:REF 
/LIBPATH:F:\MinGW32\src\Languages\mruby\build\host\bin\/lib
```

How is `link.exe` (from `bin\mruby-config.bat --ld`) supposed to understand this `host\bin\/lib` part?
Where does it come from? It better be:
```c
/LIBPATH:F:\MinGW32\src\Languages\mruby\build\host\bin\..\lib
```
or expanded to `f:\MinGW32\src\Languages\mruby\build\host\lib`.

dearblue:
Previously, the correct directory could not be obtained in some environments.
Therefore, changed the method to leave an element indicating the parent directory.

fix #6156

### DIFF
--- a/mrbgems/mruby-bin-config/mruby-config.bat
+++ b/mrbgems/mruby-bin-config/mruby-config.bat
@@ -1,16 +1,6 @@
 @echo off
 
-call :init "%~dp0."
-goto top
-
-:init
-rem call init2 for treatments last directory separator
-call :init2 "%~dp1."
-goto :eof
-
-:init2
-set MRUBY_PACKAGE_DIR=%~f1
-goto :eof
+set MRUBY_PACKAGE_DIR=%~dp0..
 
 :top
 shift


jbampton (commented):
The link is valid so the pre-commit check failure is just a false positive.

In future we may need to add a config file for the `markdown-link-check`.

https://github.com/tcort/markdown-link-check?tab=readme-ov-file#config-file-format

PR #6178, opened 2024-02-20
super-linter: set `VALIDATE_GITLEAKS` to `true`

jbampton:
Flag to enable the linting process of the secrets.

Just another check or test we can have.

Set VALIDATE_ALL_CODEBASE: true

https://github.com/super-linter/super-linter?tab=readme-ov-file#configure-super-linter


jbampton (commented):
pre-commit is failing on the `markdown-link-check` for a link that is not broken

Issue #6182, opened 2024-02-22
`mrb_read_float()` converts `"0.3"` with a small error compared to `strtod()`


dearblue:
As far as I can tell, `"0.6"` and `"0.7"` also do not have the same value as `strtod()`.

The reason I noticed this is [mruby-marshal](https://github.com/take-cheeze/mruby-marshal).
However, when I do `Marshal.dump/load`, the respective results for `0.6` and `0.7` seem to be equal.

```console
% bin/mirb
mirb - Embeddable Interactive Ruby Shell

> Marshal.load(Marshal.dump(0.3)) == 0.3
 => false
> Marshal.load(Marshal.dump(0.6)) == 0.6
 => true
> Marshal.load(Marshal.dump(0.7)) == 0.7
 => true
```

dearblue:
For information, it seems also to have the same problem with the original, https://github.com/mattn/strtod.
Adding values in `0.1` intervals from `0.2` to `0.9` and reporting only failures, the results are shown below.

```console
%  cc strtod.c -lm && ./a.out
CASE: .3
  ERR: .3, No error: 0
    E1 0.300000, 0.3, , 0
    E2 0.300000, 0.3, , 0
different value

CASE: .6
  ERR: .6, No error: 0
    E1 0.600000, 0.6, , 0
    E2 0.600000, 0.6, , 0
different value

CASE: .7
  ERR: .7, No error: 0
    E1 0.700000, 0.7, , 0
    E2 0.700000, 0.7, , 0
different value

CASE: 2.2250738585072010e-308
  ERR: 2.2250738585072010e-308, Result too large
    E1 0.000000, 0, , 34
    E2 0.000000, 2.22507e-308, , 34
different value

CASE: 2.2250738585072011e-308
  ERR: 2.2250738585072011e-308, Result too large
    E1 0.000000, 0, , 34
    E2 0.000000, 2.22507e-308, , 34
different value
```

<details>
<summary>patch for test of strtod.c</summary>

```diff
diff --git a/strtod.c b/strtod.c
index bd5378b..42bb574 100644
--- a/strtod.c
+++ b/strtod.c
@@ -148,8 +148,6 @@ test(char* str)
     char *e1, *e2;
     int x1, x2;
 
-    printf("CASE: %s\n", str);
-
     errno = 0;
     e1 = NULL;
     d1 = vim_strtod(str, &e1);
@@ -161,22 +159,29 @@ test(char* str)
     x2 = errno;
 
[...]

PR #6187, opened 2024-02-25
Added limitation about floating point numbers

dearblue:
ref. #6182


matz:
bd668e4c9 made reading float numbers more accurate. So we don't need this explanation anymore, I believe.

PR #6191, opened 2024-03-07
Speed up symbol equality comparison

leviongit:


src/vm.c
```diff
@@ -2662,6 +2662,9 @@ mrb_vm_exec(mrb_state *mrb, const struct RProc *proc, const mrb_code *pc)
       if (mrb_obj_eq(mrb, regs[a], regs[a+1])) {
         SET_TRUE_VALUE(regs[a]);
       }
+      else if (mrb_type(regs[a]) == MRB_TT_SYMBOL) {
```
matz, line 2665: Thank you for the pull-request. This is a nice idea. But if the condition is `mrb_immediate_p(regs[a])`, equalify comparison will be (slightly) faster not only for symbols, but other immediate values such as integers and float numbers.

src/symbol.c
```diff
@@ -691,4 +698,5 @@ mrb_init_symbol(mrb_state *mrb)
   mrb_define_method_id(mrb, sym, MRB_SYM(to_sym),  mrb_obj_itself,  MRB_ARGS_NONE());      /* 15.2.11.3.4 */
   mrb_define_method_id(mrb, sym, MRB_SYM(inspect), sym_inspect, MRB_ARGS_NONE());          /* 15.2.11.3.5(x) */
   mrb_define_method_id(mrb, sym, MRB_OPSYM(cmp),   sym_cmp,     MRB_ARGS_REQ(1));
+  mrb_define_method_id(mrb, sym, MRB_OPSYM(eq),    sym_eq,      MRB_ARGS_REQ(1));
```
matz, line 701: I don't think this `sym_eq` change would improve the performance, since `mrb_obj_equal_m` (that defines `Object#==`) has exactly the same implementation. Should be removed from the PR.

leviongit: `Symbol` inherits the `==` method from `Comparable` which does a lexical comparison. a new commit will override `==` to use `mrb_obj_equal_m`. (or should i make a separate pr for this?)

src/vm.c
leviongit: this unfortunately breaks some test cases, as it means that some (e.g. `Integer#==(Rational)`) equality checks are never performed.

PR #6197, opened 2024-03-10
Fix int_xor to call flo_xor.

niyarin:
This PR fixes the problem that `flo_xor` is not  called from `int_xor`.

Code like the following  works as expected:
```ruby
10^2.0
=> 10 #expected 8
````


matz:
Thank you for the PR. You have found a long-lived bug. 
I noticed CRuby raises exceptions for bit-operations with integers and floats.
Maybe we should fix the issue differently.

matz:
I remember that mruby used to use float numbers when integers overflowed before it provided big integers.
We implemented big integers, so bit operations for float numbers should be removed.

PR #6205, opened 2024-03-11
GitHub labeler: label more files

jbampton:
These are some of the config files for the linters/tests that run

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -48,6 +48,12 @@ github:
       - changed-files:
           - any-glob-to-any-file:
               - .github/**/*
+miscellaneous:
+  - any:
+      - changed-files:
+          - any-glob-to-any-file:
+              - .editorconfig
+              - .prettierrc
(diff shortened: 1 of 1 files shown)


jbampton:
See in refs #6204 that the pull request was not labeled.

Best to have labels for all files / folders.

dearblue:
I think it's a bit different to belong to the `build` label.

:thinking: Since I think the number is small and not a noteworthy classification, I propose `miscellaneous` and cast my vote.
Other label names I can think of are `edit`, `coding`, and `utility`.
If there is another preferred proposal, I may vote for that one.

jbampton:
Thanks @dearblue I went with `miscellaneous` for the new label

jbampton:
Hey @matz 

I forgot to mention that we need to add the new label for `miscellaneous` on the repo labels page here:

https://github.com/mruby/mruby/labels

Thanks

matz:
Done.

PR #6207, opened 2024-03-16
Add `rubocop` `pre-commit` hook; Remove redundant returns

jbampton:
We already run Rubocop on `mgem-list` as seen here:

https://github.com/mruby/mgem-list/blob/7944a361ab9acb0d673f15fbf0e2ce4b80f21514/.pre-commit-config.yaml#L46

This PR adds a basic Rubocop config file and the goal is to standardize the codebase.

Using `rubocop -A` the redundant returns have been autoremoved.

### DIFF
--- a/.github/linters/.rubocop.yml
+++ b/.github/linters/.rubocop.yml
@@ -1,0 +1,6 @@
+AllCops:
+  DisabledByDefault: true
+  TargetRubyVersion: 3.3.0
+
+Style/RedundantReturn:
+  Enabled: true
--- a/lib/mruby/build/load_gems.rb
+++ b/lib/mruby/build/load_gems.rb
@@ -68,19 +68,19 @@ module MRuby
 
       def full_gemdir
         return @gemdir unless @path
-        return File.join(@gemdir, @path)
+        File.join(@gemdir, @path)
       end
 
-      def canonical?()  return @canonical;                  end
-      def git?()        return !!@repo;                     end
-      def gemname()     return File.basename(@gemdir);      end
+      def canonical?()  @canonical;                  end
+      def git?()        !!@repo;                     end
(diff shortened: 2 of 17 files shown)


lib/mruby/build/load_gems.rb
```diff
@@ -68,32 +68,32 @@ def initialize(gemdir, repo, branch, commit, canonical, path = nil)
...
-          @branch == other.branch &&
-          @commit == other.commit &&
-          @canonical == other.canonical? &&
-          full_gemdir == other.full_gemdir
+        @gemdir == other.gemdir &&
+        @repo == other.repo &&
+        @branch == other.branch &&
+        @commit == other.commit &&
+        @canonical == other.canonical? &&
+        full_gemdir == other.full_gemdir
```
dearblue, line 88: `return` is not important. However, the indentation here is very ( :smile_cat: ) important, so could you keep it?

jbampton: Thanks @dearblue the indentation is now fixed

.github/linters/.rubocop.yml
```diff
@@ -0,0 +1,6 @@
+AllCops:
+  DisabledByDefault: true
+  TargetRubyVersion: 3.3.0
+
+Style/RedundantReturn:
```
matz, line 5: Is this rule really needed? I am not sure the changes from the rule make the code more readable.

jbampton: I started this PR thinking that we could use some rules from Rubocop to help organize and standardize some of the Ruby codebase.  

I think it is better to be consistent and removing these redundant returns sets a code standard going forwards. These extra returns are really extra characters and data that is not needed.

Maybe this rule is not needed but I think there could be some other Rubocop rules that are useful for mruby.

PR #6212, opened 2024-03-21
Add `gem.last_initializer`

dearblue:
Provides a mechanism for the GEM author to do the final processing after the blocks specified in `build.gem` have been processed.

This allows GEM author and user code to be processed in the following order:

 1. The block passed to `MRuby::Gem::Specification.new` in `mrbgem.rake`
 2. The block passed to `MRuby::Build#gem` in the build configuration file.
 3. The block passed to `MRuby::Gem::Specification#last_initializer` in `MRuby::Gem::Specification.new`.

---

When performing library detection, I would like to use it to auto-detect when GEM users do not specify anything.
https://github.com/dearblue/mruby-stacktrace/blob/TENTATIVE.1/mrbgem.rake#L123-L144

So far, if desired to specify a block for default settings, this can be accomplished by directly assigning it to an instance variable.
However, when viewed as a public API, I do not believe this is the preferred method.
https://github.com/dearblue/mruby-stacktrace/blob/TENTATIVE.1/mrbgem.rake#L117

### DIFF
--- a/lib/mruby/gem.rb
+++ b/lib/mruby/gem.rb
@@ -43,6 +43,7 @@ module MRuby
       def initialize(name, &block)
         @name = name
         @initializer = block
+        @last_initializer = nil
         @version = "0.0.0"
         @dependencies = []
         @conflicts = []
@@ -84,6 +85,7 @@ module MRuby
         build.libmruby_objs << @objs
 
         instance_eval(&@build_config_initializer) if @build_config_initializer
+        instance_eval(&@last_initializer) if @last_initializer
 
         repo_url = build.gem_dir_to_repo_url[dir]
         build.locks[repo_url]['version'] = version if repo_url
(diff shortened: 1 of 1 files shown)


matz:
Thank you for the proposal, however, I don't think the name `last_initializer` is appropriate, since it does not describe the true purpose of this initializer. Any idea (especially for English speakers)?

matz:
Thank you for this PR. Picked it into master as commit 2e429a034d (with you as author) — 14 months later than ideal, sorry for the wait.

The functional change is exactly as you proposed; the only difference is the name. Per matz's review request, I renamed `last_initializer` to **`post_user_config`** to make the timing relationship explicit (it runs after the user's `build.gem` block).

Initialization order is unchanged from your proposal:
1. block in `MRuby::Gem::Specification.new` (gem author defaults)
2. block in `build.gem` (user override)
3. block in `post_user_config` (gem author finalize, new)

This should let `mruby-stacktrace` retire the `instance_variable_set` workaround you mentioned. Thanks again for the contribution and for spotting the naming concern as a real one.

PR #6220, opened 2024-03-30
Reorganize `mrb_cache_entry` and `mrb_method_t` types

dearblue:
The purpose is to remove the `mid` field from the `mrb_cache_entry` structure. The resulting RAM requirement for the method cache is reduced from 5 words per entry to 4 words per entry for 32-bit CPUs.

The relevant changes are as follows:

  - Removed `MRB_USE_METHOD_T_STRUCT`.

    The `mrb_method_t` type is now always defined as a structure.

  - Include method IDs in `mrb_method_t`

    Change the `flags` member to `uint32_t`. The bitstring structure should be the same as the keys of the `mt` table in `class.c`.

I believe the impact on API compatibility with previous versions is minimal.

---

Attached is a benchmark comparison for your reference.
The targets are (1) the [current master](https://github.com/mruby/mruby/tree/61b1a4db8efecea4a09f7c4ce9449cd6f52f4bc6), (2) this PR, (3) [another alternative](https://github.com/dearblue/mruby/commit/b2bc93209565b578a81267cfe8bf8a0b6e4c2769) and (4) [3.3.0](https://github.com/mruby/mruby/tree/3.3.0).
Another alternative is to move the bits embedded in `struct mrb_cache_entry::mid` to `struct mrb_cache_entry::c0` (like "word-boxing"). [Link to difference](https://github.com/dearblue/mruby/commit/b2bc93209565b578a81267cfe8bf8a0b6e4c2769#diff-114597163077c0642dba00237dcabafba47b8e0486d136299bb669a15abae822L246-R252).

bm-result-20240330T184902

### DIFF
--- a/doc/guides/mrbconf.md
+++ b/doc/guides/mrbconf.md
@@ -236,12 +236,6 @@ end
 - Ignored if `MRB_NO_METHOD_CACHE` is defined.
 - Need to be the power of 2.
 
-`MRB_USE_METHOD_T_STRUCT`
-
-- Use C struct to represent `mrb_method_t`
-- No `MRB_USE_METHOD_T_STRUCT` requires highest 2 bits of function pointers to be zero
-- Define this macro on machines that use higher bits of pointers
-
 `MRB_USE_ALL_SYMBOLS`
 
 - Make it available `Symbol.all_symbols` in `mrbgems/mruby-symbol-ext`
--- a/include/mrbconf.h
+++ b/include/mrbconf.h
@@ -48,16 +48,6 @@
 //#define MRB_METHOD_CACHE_SIZE (1<<8)
 //#define MRB_USE_INLINE_METHOD_CACHE
 
-/* add -DMRB_USE_METHOD_T_STRUCT on machines that use higher bits of function pointers */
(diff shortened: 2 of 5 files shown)


dearblue:
The following is the number of CPU cycles obtained from `valgrind --tool=cachegrind -- bin/mruby benchmark/bm_fib.rb`.
The order of the lines is the same as in the previous benchmark.
The first group is the results from `gcc13 -O3` and the second group is the results from `gcc13 -Os`.

```
40,186,751,990 (100.0%)  PROGRAM TOTALS         master
40,656,510,074 (100.0%)  PROGRAM TOTALS         this PR
40,733,987,325 (100.0%)  PROGRAM TOTALS         another alternative
40,577,677,399 (100.0%)  PROGRAM TOTALS         3.3.0

61,020,950,744 (100.0%)  PROGRAM TOTALS         master
61,021,647,478 (100.0%)  PROGRAM TOTALS         this PR
61,177,299,954 (100.0%)  PROGRAM TOTALS         another alternative
61,333,670,456 (100.0%)  PROGRAM TOTALS         3.3.0
```

\# I only recently learned about the role and usage of [Cachegrind](https://valgrind.org/docs/manual/cg-manual.html) from the [SQLite documentation](https://www.sqlite.org/cpu.html).

PR #6227, opened 2024-04-04
fix: `Array#shuffle(!)` result distribution

leviongit:
previously the shuffle algorithm would leave the first element untouched about 1/3rd of the time

### DIFF
--- a/mrbgems/mruby-random/src/random.c
+++ b/mrbgems/mruby-random/src/random.c
@@ -261,7 +261,7 @@ static mrb_value
 mrb_ary_shuffle_bang(mrb_state *mrb, mrb_value ary)
 {
   if (RARRAY_LEN(ary) > 1) {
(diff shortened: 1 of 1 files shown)


Ferociousfeind:
I can mathematically confirm these changes are correct- there are "n!" ways to order a list of "n" elements, however the original functionality would randomly pick one of "n^n" elements.

This necessarily would produce lopsided behavior, as n^n is not an integer multiple of n!, except in the case where n=0, 1, or 2 (where n^n is (effectively) 0, 1 and 4 respectively, and n! is (effectively) 0, 1 and 2 respectively), in all other cases, the "n is never divisible by n-1" rule comes into play and makes some permutations appear more often than others.

This appears to manifest as the 0th element of an array moving around significantly less often than other elements, which this PR fixes.

matz:
Thank you.

PR #6230, opened 2024-04-07
Fix status of fiber after switched by exception raised

dearblue:
The state of a fiber switched due to an exception occurrence was incorrectly set to "Suspended".

```ruby
Fiber.new {
  begin
    Fiber.new { 0 / 0 }.resume
  rescue
    p Fiber.current
    # before  => #<Fiber:0x159f61e43ce0 fiber.rb:1 (suspended by resuming)>
    # after   => #<Fiber:0x159f61e43ce0 fiber.rb:1 (resumed)>
  end
}.resume
```

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -369,6 +369,20 @@ cipush(mrb_state *mrb, mrb_int push_stacks, uint8_t cci, struct RClass *target_c
   return ci;
 }
 
+static void
+fiber_terminate(mrb_state *mrb, struct mrb_context *c, mrb_callinfo *ci)
+{
+  mrb_assert(c != mrb->root_c);
+
+  c->status = MRB_FIBER_TERMINATED;
+
+  /* fiber termination should automatic yield or transfer to root */
+  mrb->c = c->prev;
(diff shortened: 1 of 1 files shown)


src/vm.c
```diff
@@ -369,6 +369,20 @@ cipush(mrb_state *mrb, mrb_int push_stacks, uint8_t cci, struct RClass *target_c
...
 }
 
+static void
+fiber_terminate(mrb_state *mrb, struct mrb_context *c, mrb_callinfo *ci)
+{
+  mrb_assert(c != mrb->root_c);
+
+  c->status = MRB_FIBER_TERMINATED;
+
+  /* fiber termination should automatic yield or transfer to root */
```
dearblue, line 379: The seemingly unnatural blank line compared to the deleted hunk below is due to a scheduled pull request to follow.

https://github.com/dearblue/mruby/commit/0965a2c24c55c2181b6ec83a0825eda25e30a2b3#diff-b7d97be9c98fd50d49b9f2d6ee9bd3e4da589e223eeb0a06824d7af57d23efe6

PR #6231, opened 2024-04-07
Add a way to let other gems handle closing of fds in mruby-io

Asmod4n:
I'm currently writing a gem which wants to integrate with mruby-io and mruby-socket.
At the moment mruby-io closes fds it didn't create itself and thus can cause problems when other gems except the fd to still be available.

This only changes how ```fptr_finalize``` handles the closing of fds and nothing else.
I've tried handling that in my gem by stopping a created mruby-io object from being garbage collected until it's not needed anymore, but that wasn't enough.

### DIFF
--- a/mrbgems/mruby-io/include/mruby/ext/io.h
+++ b/mrbgems/mruby-io/include/mruby/ext/io.h
@@ -41,7 +41,9 @@ struct mrb_io {
                writable:1,
                eof:1,
                sync:1,
-               is_socket:1;
+               is_socket:1,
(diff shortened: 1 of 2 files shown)


Asmod4n:
A short explanation: i'm writing a mruby gem to implement io_uring from linux.
This is fixing a otherwise impossible to handle race condition.

I've tried several ways to fix this in my gem, the only thing that worked is to mrb_gc_register every mruby-io object i created and never unregister them, because you can't know when its save to release them.

PR #6233, opened 2024-04-09
Free stack memory at end of fiber

dearblue:
Immediately frees the call stack and data stack at the end of a non-root fiber.
If the env object needs to be detached, the data stack is reused through `mrb_realloc()`.

Previously, it was not necessary to take into account that `c->cibase` could be `NULL`.
Note that this is no longer the case due to this patch.
In fact, changes to "mruby-fiber" are now required.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -374,7 +374,45 @@ fiber_terminate(mrb_state *mrb, struct mrb_context *c, mrb_callinfo *ci)
 {
   mrb_assert(c != mrb->root_c);
 
+  struct REnv *env = CI_ENV(ci);
+  mrb_assert(env == NULL || MRB_ENV_LEN(env) <= c->stend - ci->stack);
+
   c->status = MRB_FIBER_TERMINATED;
+  mrb_free(mrb, c->cibase);
+  c->cibase = c->ciend = c->ci = NULL;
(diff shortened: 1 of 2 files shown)


src/vm.c
```diff
@@ -374,7 +374,45 @@ fiber_terminate(mrb_state *mrb, struct mrb_context *c, mrb_callinfo *ci)
...
+
+    size_t len = (size_t)MRB_ENV_LEN(env);
+    if (len == 0) {
+      env->stack = NULL;
+      MRB_ENV_CLOSE(env);
+      mrb_free(mrb, stack);
+    }
+    else {
+      mrb_assert(stack == env->stack);
+      mrb_write_barrier(mrb, (struct RBasic*)env);
```
matz, line 400: Do we need `mrb_write_barrier()` here?

dearblue: Thanks for your review.

I am not sure exactly.
This `fiber_terminate()` is based on `mrb_env_unshare()`.

It was removed from `mrb_env_unshare()` once, but was reintroduced in commit 93648fc954c0133acc6e89860ba49f78ca60460f, so I added it.

matz: Oh, I see. I investigated and found out we need it. Thank you.

PR #6235, opened 2024-04-12
fix `Array#delete` always firing the block when deleting `nil`

leviongit:
reimplement `Array#delete` in c, fixing `ary.delete(nil, &blk)` firing the block regardless of removal

minimal reproduction:
```rb
ary = [nil]
ret = ary.delete(nil) { "not deleted?" }
```

result on `master`:
```rb
ret
# => "not deleted?"
ary
# => []
```

result after this patch:
```rb
ret 
# => nil
ary
# => []
```

### DIFF
--- a/mrblib/array.rb
+++ b/mrblib/array.rb
@@ -191,11 +191,32 @@ class Array
   #
   # Delete element with index +key+
   def delete(key, &block)
-    while i = self.index(key)
-      self.delete_at(i)
-      ret = key
+    ret = key
+    i = 0
+    j = 0
+    len = self.length
+    while i < len
+      elem = self[i]
+
+      if key == elem
+        ret = elem
+        i += 1
+        next
+      end
+
+      self[j] = elem if i != j
+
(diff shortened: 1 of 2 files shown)


matz:
Thanks for the PR. But as a principle, the mruby standard library does not call blocks from C code unless absolutely necessary (to avoid VM recursion). I do not consider it an absolute necessity this time.
We have several options:
* After PR merge, I will modify it not to call the block
* I fix the Ruby code
* @leviongit will update the PR

Which is your preference?

leviongit:
thank you for the comment. i will amend the pr.

matz:
I that case, you can implement the body part in C (which probably returns index or - 1; you may name it __delete), or you can fix the Ruby code.

leviongit:
i will check whether implementing the core part in c is worthwhile and will come back with results

leviongit:
this pr is ready to merge

matz:
Thank you!

PR #6253, opened 2024-04-23
Allow recycling fibers by GC if not referenced directly

dearblue:
The patch assumes that `struct REnv::cxt` only performs checks with the `OP_BREAK` and `OP_RETURN_BLK` instructions, and does not reference the entity.
Therefore, by changing to a weak reference, it is possible to collect fibers that are no longer directly referenced while in the suspended state.

However, we need to detach the living env objects that remain in the call stack of the fiber.
So, in effect, it involves a revert of following commits.
  - commit a3365d8b3fc957b1d1fd89526440358aa9402fc3
  - commit 57ffa1c1508f01434806b1b7301f4ae982eb9ae0

Examples of the effects of change are shown below.
Note that it was built with `rake MRUBY_CONFIG=host-debug`.

```ruby
f = Fiber.new { (x, y, z) = "X", "Y", "Z"; Fiber.yield -> { [x, y, z] } }
g = f.resume
GC.start
p ObjectSpace.memsize_of_all
# => 59532
g.call
# => ["X", "Y", "Z"]
f = nil
GC.start
ObjectSpace.memsize_of_all
# BEFORE => 59532
# AFTER  => 58044
g.call
# => ["X", "Y", "Z"]
```

---

In addition, it also takes the necessary precautions to prevent the possibility of infinite loops with this change.

### DIFF
--- a/src/gc.c
+++ b/src/gc.c
@@ -630,9 +630,8 @@ gc_mark_children(mrb_state *mrb, mrb_gc *gc, struct RBasic *obj)
     {
       struct REnv *e = (struct REnv*)obj;
 
-      if (MRB_ENV_ONSTACK_P(e) && e->cxt && e->cxt->fib) {
-        mrb_gc_mark(mrb, (struct RBasic*)e->cxt->fib);
-      }
+      // The data stack must always be protected from GC regardless of the MRB_ENV_CLOSE flag.
+      // This is because the data stack is not protected if the fiber is GC'd.
       mrb_int len = MRB_ENV_LEN(e);
       for (mrb_int i=0; i<len; i++) {
         mrb_gc_mark_value(mrb, e->stack[i]);
@@ -773,7 +772,20 @@ obj_free(mrb_state *mrb, struct RBasic *obj, mrb_bool end)
     {
       struct mrb_context *c = ((struct RFiber*)obj)->cxt;
 
-      if (c != mrb->root_c) {
+      if (c && c != mrb->root_c) {
+        if (!end && c->status != MRB_FIBER_TERMINATED) {
+          mrb_callinfo *ci = c->ci;
+          mrb_callinfo *ce = c->cibase;
+
+          while (ce <= ci) {
+            struct REnv *e = ci->u.env;
+            if (e && !is_dead(&mrb->gc, (struct RBasic*)e) &&
+                e->tt == MRB_TT_ENV && MRB_ENV_ONSTACK_P(e)) {
+              mrb_env_unshare(mrb, e, TRUE);
+            }
+            ci--;
+          }
+        }
         mrb_free_context(mrb, c);
       }
     }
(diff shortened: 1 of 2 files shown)


src/gc.c
```diff
@@ -773,7 +772,20 @@ obj_free(mrb_state *mrb, struct RBasic *obj, mrb_bool end)
     {
       struct mrb_context *c = ((struct RFiber*)obj)->cxt;
 
-      if (c != mrb->root_c) {
+      if (c && c != mrb->root_c) {
```
matz, line 775: Since `mrb->root_c` is never NULL, I don't think the `c &&` check is necessary.

dearblue: Thanks for your review.

I checked and it seems that omitting the `NULL` check on the `c` variable causes a crash due to a "null pointer dereference".

```console
% lldb18 -- bin/mruby -e 'Fiber.allocate; GC.start'
(lldb) target create "bin/mruby"
Current executable set to '/var/tmp/mruby/bin/mruby' (x86_64).
(lldb) settings set -- target.run-args  "-e" "Fiber.allocate; GC.start"
(lldb) r
Process 90188 launched: '/var/tmp/mruby/bin/mruby' (x86_64)
Process 90188 stopped
* thread #1, name = 'mruby', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x30)
    frame #0: 0x0000000000413264 mruby`obj_free(mrb=0x000008d4df809000, obj=0x000008d4df82ad70, end=false) at gc.c:776:24
   773        struct mrb_context *c = ((struct RFiber*)obj)->cxt;
   774
   775        if (c != mrb->root_c) {
-> 776          if (!end && c->status != MRB_FIBER_TERMINATED) {
   777            mrb_callinfo *ci = c->ci;
   778            mrb_callinfo *ce = c->cibase;
   779
```

`mrb_free_context()` is "NULL safe", so there is no problem.

src/gc.c
```diff
@@ -628,9 +630,8 @@ gc_mark_children(mrb_state *mrb, mrb_gc *gc, struct RBasic *obj)
     {
       struct REnv *e = (struct REnv*)obj;
 
-      if (MRB_ENV_ONSTACK_P(e) && e->cxt && e->cxt->fib) {
-        mrb_gc_mark(mrb, (struct RBasic*)e->cxt->fib);
-      }
```
dearblue, line 633: I was checking again and found a period when this section did not exist.

```console
% git log -L627,640:src/gc.c 22518b5b789293e7ddad7a94def44039937d8e8c
```

There is no mention of a commit message, but it corresponds to a revert of commit b6598e052fda9a27692156ab8b973ce6e3374ee3.

src/gc.c
matz: Well, thank you for the confirmation. My bad.

PR #6257, opened 2024-04-27
fix some mrbconf.md typos

jethrodaniel:


doc/guides/mrbconf.md
```diff
@@ -244,3 +244,7 @@ end
 `MRB_USE_VM_SWITCH_DISPATCH`
 
 - Turn on switch dispatch in VM loop
+
+`MRB_NO_GEMS`
+
+- Disable mgem usage
```
dearblue, line 250: `MRB_NO_GEMS` is automatically set if none of the gems are specified in the build configuration file.
https://github.com/mruby/mruby/blob/7edbff8c83cd19749da576aca945fcbf64602bd5/lib/mruby/build.rb#L368

I would question the need for the user to knowingly set this up.
Can you tell me what kind of use you envision?

jethrodaniel: Misunderstanding on my part, thank you for the clarification. 👍 

I agree, this shouldn't be in user-facing documentation

PR #6260, opened 2024-05-05
Remove `exc_caught` from `mrb_vm_exec()`

dearblue:
Only go to exception handling if `mrb->exc` is non-null.

This may cause some compatibility problems, but I doubt that it is necessary to maintain that compatibility.
Here is how I see the incompatibility with the change at this time:
  - If `mrb->exc` is non-null and `mrb_vm_exec()` is called, an exception will be thrown immediately.
  - If `MRB_THROW()` is used while `mrb->exc` is `NULL`, it will not go to exception handling.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1432,15 +1432,13 @@ mrb_vm_exec(mrb_state *mrb, const struct RProc *proc, const mrb_code *pc)
   };
 #endif
 
-  volatile mrb_bool exc_caught = FALSE;
 RETRY_TRY_BLOCK:
 
   MRB_TRY(&c_jmp) {
 
-  if (exc_caught) {
-    exc_caught = FALSE;
+  if (mrb->exc) {
     mrb_gc_arena_restore(mrb, ai);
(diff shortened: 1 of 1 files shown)


dearblue:
As a note, write down what I am thinking further as a change in behavior.

> If `mrb->exc` is non-null and `mrb_vm_exec()` is called, an exception will be thrown immediately.

Put `mrb->exc = NULL` before the `RETRY_TRY_BLOCK` label.
In principle, `mrb->exc` is kept `NULL` in VM.
Taking that idea further, it might not hurt to remove the first `if (mrb->exc)` condition, such as `OP_RETURN`.
However, `mrb_funcall_with_block()` and `mrb_yield_argv()` may need similar `mrb->exc` checks before and after C function calls.

> If `MRB_THROW()` is used while `mrb->exc` is `NULL`, it will not go to exception handling.

Put `mrb->exc = mrb_exc_new_lit(mrb, E_RUNTIME_ERROR, “”)` before `goto RETRY_TRY_BLOCK`.

PR #6276, opened 2024-05-26
Always run `atexit` on the top-level call frame

dearblue:
In the following example, the fiber context and call stack may be in an incomplete state.
  - In case another thread running mruby is terminated abnormally
  - In case of a global jump that is out of management by mruby


dearblue:
I should have done `MRB_ENV_CLOSE(ci->e->env)` with `ci->env->stack` of incomplete `ci` set to `NULL`.
I will send a corrected PR in the near future.

PR #6290, opened 2024-06-09
Allow to change the output directory name of the `libmruby` file

### LINKED ISSUE #6240: Differentiate between `lib` and `lib64` in the build settings.
See: https://docs.fedoraproject.org/en-US/packaging-guidelines/RPMMacros/#macros_installation
In Fedora, the default library  path for the `x86_64` architecture is `/usr/lib64`. I would like to be able to export some variable to before the build to allow `rake install` to use that path.

dearblue:
Add the `MRuby::Build#libdir_name` accessor and the `MRUBY_SYSTEM_LIBDIR_NAME` environment variable.

Resolved #6240

---

I was thinking too hard when I commented on #6240.
This PR limits the modification of `lib/` to the build task (as described in the documentation).

### DIFF
--- a/doc/guides/compile.md
+++ b/doc/guides/compile.md
@@ -117,6 +117,36 @@ set the character via `conf.file_separator`.
 conf.file_separator = '/'
 ```
 
+### Name of library directory
+
+In some environments, the `libmruby.a` file requires a different directory name than `lib`.
+You can be changed to any name by the `conf.libdir_name` accessor.
+
+```ruby
+conf.libdir_name = 'lib64'
+```
+
+Alternatively, it can be changed via the environment variable `MRUBY_SYSTEM_LIBDIR_NAME` when
+the `rake` command is run.
+
+```console
+$ export MRUBY_SYSTEM_LIBDIR_NAME=lib64
+$ rake clean all
+```
+
+NOTES:
+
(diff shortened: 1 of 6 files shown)


doc/guides/compile.md
```diff
@@ -117,6 +117,36 @@ set the character via `conf.file_separator`.
 conf.file_separator = '/'
 ```
 
+### Name of library directory
+
+In some environments, the `libmruby.a` file requires a different directory name than `lib`.
+You can be changed to any name by the `conf.libdir_name` accessor.
```
mkoncek, line 123: My suggestion for the sentence is:
```adoc
This directory can be changed to a different name using the `conf.libdir_name` accessor.
```

dearblue: Thanks for your suggestion. Unfortunately this time the merge was faster than I can check.
I am only converting from Japanese to English using [DeepL](https://www.deepl.com/translator). So I would be very grateful if you could improve the English text.
If you don't mind, could you please create a PR? Thank you very much.

mkoncek: Sure, at some point I can go over the whole text, although I am also not a native speaker...

dearblue: Okay, I will not say anything unreasonable.
Please let us know if you have any further concerns.

PR #6303, opened 2024-07-16
mrb_str_aset_m() should return replace instead of str

hasumikin:
`string[]=(idx, replace)` should return `replace`.

## Actual (wrong)

```
string.[]=(idx, replace) → string
string.[]=(idx, len, replace) → string
```

## Expected

```
string.[]=(idx, replace) → replace
string.[]=(idx, len, replace) → replace
```

## Sidenote

As of the current mruby-compiler, `(string[idx] = 'X')` creates not only "CALL_NODE" but also "ASGN_NODE" and "OP_MOVE", overriding the wrong return value. On the other hand, `string.[]=(idx, 'X')` creates only "CALL_NODE", exposing the wrong return value.

If my new mruby-compiler2, leveraging Prism, took the place of official compiler, `(string[idx] = 'X')` and `string.[]=(idx, 'X')` would be going to generate the same VM code without "OP_MOVE". So I paranoidly added tests.

FYI: You can find how the new mruby-compiler2's AST and VM code look like in mruby/c's issue (mruby/c had the same bug): https://github.com/mrubyc/mrubyc/pull/210

### DIFF
--- a/src/string.c
+++ b/src/string.c
@@ -1452,6 +1452,7 @@ mrb_str_aset(mrb_state *mrb, mrb_value str, mrb_value indx, mrb_value alen, mrb_
  *
  * Modify +self+ by replacing the content of +self+.
  * The portion of the string affected is determined using the same criteria as +String#[]+.
+ * The return value of this expression is +replace+.
  */
 static mrb_value
 mrb_str_aset_m(mrb_state *mrb, mrb_value str)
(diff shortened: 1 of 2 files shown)


matz:
Merged. But considering the behavior of the current CRuby parser, I assume the Prism generation is a bug.

hasumikin:
@matz Thank you.
I will talk to Kevin regarding the lack of NODE_ASGN

Issue #6307, opened 2024-07-21
Planned change patch for `mrb_vm_exec()`


dearblue:
There are over 10 patches and I plan to split them up and do a PR.
Link to patch collection: https://github.com/mruby/mruby/compare/master...dearblue:mruby:mruby-vm

When all patches are applied, I show how the object code size and stack size used by `mrb_vm_exec()` changes on several CPU architectures.

The comparison is the result of simply running the compiler from the command line and compiling `src/vm.c`.
The commit hashes to be compared are ac8846d and 873b8bb.

The filename of the object code is represented by `#{revision}-#{gcc-name}-vm.o`.

<details>
<summary>script and compilers</summary>

  - script

    ```sh
    for rev in ac8846d 873b8bb
    do
      git checkout $rev

      for gcc in /usr/local/bin/*-gcc-*.0
      do
        $gcc -Iinclude -Ibuild/host/include -fstack-usage -Os -c -o $rev-$(basename $gcc)-vm.o src/vm.c
      done
    done

    size ac8846d*.o 873b8bb*.o | sort -st- -k2
    grep -w mrb_vm_exec ac8846d-*.su 873b8bb-*.su | sort -st- -k2 | awk '{ printf "%3d %-15s %s\n", $2, $3, $1 }'
    ```

  - list of compilers on my FreeBSD 14.1/amd64

    ```
    /usr/local/bin/aarch64-unknown-freebsd14.0-gcc-12.3.0
    /usr/local/bin/arm-none-eabi-gcc-11.3.0
    /usr/local/bin/armv6-gnueabi-freebsd14.0-gcc-12.3.0
    /usr/local/bin/armv7-gnueabi-freebsd14.0-gcc-12.3.0
    /usr/local/bin/avr-gcc-14.1.0
    /usr/local/bin/i386-unknown-freebsd14.0-gcc-12.3.0
    /usr/local/bin/mips-unknown-freebsd14.0-gcc-9.5.0
    /usr/local/bin/mips64-unknown-freebsd14.0-gcc-9.5.0
    /usr/local/bin/powerpc-unknown-freebsd14.0-gcc-12.3.0
    /usr/local/bin/powerpc64-unknown-freebsd14.0-gcc-12.3.0
    /usr/local/bin/riscv32-unknown-elf-gcc-11.3.0
    /usr/local/bin/riscv64-unknown-freebsd14.0-gcc-12.3.0
[...]

PR #6309, opened 2024-07-21
Perform pseudo binary search with `mrb_type()`

dearblue:


matz:
It makes `mrb_type()` slower, at least on my Linux box.

dearblue:
Thanks for the review.

In my environment, I thought the end result would be more favorable with this patch.
When I revert this patch from #6307, I get the following

| file                              | ratio
| --------------------------------- | --------
| `benchmark/bm_ao_render.rb`       | 0 to 2.2% slower
| `benchmark/bm_app_lc_fizzbuzz.rb` | 1.3 to 1.6% slower
| `benchmark/bm_fib.rb`             | 12.2 to 26.6% slower
| `benchmark/bm_so_lists.rb`        | 1.6 to 2.6% faster

Comparing this PR with the one just before, the `benchmark/bm_fib.rb` results are better, but the others are little changed or worse.

If this is a concern, could you revert this patch?

PR #6310, opened 2024-07-21
Doubling the call stack when extending it

dearblue:
Previously it was "approximately double".

---

Simulating the current behavior and this patch's behavior, the following is shown:

  - before this patch

    ```console
    % ruby -e 'size = 32; limit = 512; loop { size = (size - 1) * 2; p size; break if size >= limit }'
    62
    122
    242
    482
    962
    ```

  - after this patch

    ```console
    % ruby -e 'size = 32; limit = 512; loop { size = size * 2; p size; break if size >= limit }'
    64
    128
    256
    512
    ```

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -336,19 +336,21 @@ cipush(mrb_state *mrb, mrb_int push_stacks, uint8_t cci, struct RClass *target_c
        const struct RProc *proc, struct RProc *blk, mrb_sym mid, uint16_t argc)
 {
   struct mrb_context *c = mrb->c;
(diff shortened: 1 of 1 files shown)


matz:
I am not sure about the benefit of this change. Could you explain, please?

dearblue:
Thank you for your review.

As presented as a simulation, the current expansion is roughly twice as wide.
If exactly doubled, it would be more predictable when specifying `MRB_CALL_LEVEL_MAX`.

matz:
Understood. I admit there's no clear benefit from the current behavior. Accepted.

PR #6314, opened 2024-07-23
Optimize even?/odd? for big integers

hoshiumiarata:
Using mod for big integers leads to poor performance, so I replaced it with bitwise AND.


matz:
Thank you!

Issue #3616, opened 2017-04-18
Heap use-after-free in gc_each_objects


clayton-shopify:
The following input demonstrates a crash:
```ruby
ObjectSpace.each_object { GC.start }
```
ASAN report:
```
==28981==ERROR: AddressSanitizer: heap-use-after-free on address 0x62f00000e410 at pc 0x000100f949bc bp 0x7fff5ece51e0 sp 0x7fff5ece51d8
READ of size 8 at 0x62f00000e410 thread T0
    #0 0x100f949bb in gc_each_objects gc.c:1504
    #1 0x100f948ca in mrb_objspace_each_objects gc.c:1511
    #2 0x1011161e6 in os_each_object mruby_objectspace.c:170
    #3 0x101082618 in mrb_vm_exec (mruby:x86_64+0x100171618)
    #4 0x10107779f in mrb_vm_run (mruby:x86_64+0x10016679f)
    #5 0x1010aa8b9 in mrb_top_run (mruby:x86_64+0x1001998b9)
    #6 0x10117bba5 in mrb_load_exec (mruby:x86_64+0x10026aba5)
    #7 0x10117c4f5 in mrb_load_file_cxt (mruby:x86_64+0x10026b4f5)
    #8 0x100f139e6 in main mruby.c:227
    #9 0x7fffbbbba234 in start (libdyld.dylib:x86_64+0x5234)

0x62f00000e410 is located 16 bytes inside of 49200-byte region [0x62f00000e400,0x62f00001a430)
freed by thread T0 here:
    #0 0x101328356 in wrap_free (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0x56356)
    #1 0x10100c44b in mrb_default_allocf (mruby:x86_64+0x1000fb44b)
    #2 0x100f8e9b9 in mrb_free gc.c:269
    #3 0x100f98e2f in incremental_sweep_phase gc.c:1063
    #4 0x100f9735c in incremental_gc gc.c:1104
    #5 0x100f93326 in incremental_gc_until gc.c:1120
    #6 0x100f93727 in clear_all_old gc.c:1146
    #7 0x100f8d8f1 in mrb_full_gc gc.c:1212
    #8 0x100f94cd3 in gc_start gc.c:1324
    #9 0x101082618 in mrb_vm_exec (mruby:x86_64+0x100171618)
    #10 0x10107779f in mrb_vm_run (mruby:x86_64+0x10016679f)
[...]

PR #6334, opened 2024-08-25
Remove `iterating` variable from `mrb_objspace_each_objects()`

dearblue:


dearblue:
Thank you for your review.

I have simulated it in my head again and I believe it is fine.
To enter an `if` block, it is limited to `mrb->gc.iterating == TRUE`.
To enter an `else` block, `mrb->gc.iterating = TRUE` is set and `mrb->gc.iterating = FALSE` is set when returning from `mrb_objspace_each_objects()`.
My conclusion is that there is no need to remember the previous state by the `iterating` variable.

I also checked the issue in #3616, which was the impetus for the change to the current form, but could not reproduce the problem.

PR #6341, opened 2024-09-03
Need to restore the GC arena after some function calls

dearblue:
When calling `mrb_equal()` or `mrb_funcall()` family functions, the GC arena should be restored if the loop is repeated by a non-immediate return value.

In my opinion, restoring the GC arena is unnecessary when a non-immediate (true) value causes the function to return (e.g. the `mrb_ary_index_m()` function).

The patch does not take into account the case of recursive calls and may be incomplete.


matz:
As a general principle `equal` operations should not allocate any object (if possible), and we don't need to restore the arena index in those cases. I merged this PR just to be conservative.

PR #6342, opened 2024-09-04
Implementing hash iterator.

juchem:
`mrb_hash_foreach`, provides a foreach implementation on hashes using callbacks.

Although sufficient for ruby code, some C and C++ idioms can be made much simpler and easier with a good old `for` loop.

This pull request implements an iterator that allows user-controlled loops over hash elements in C code, as follows:
```
for (mrb_hash_iterator i = mrb_hash_iterator_new(h);
  mrb_hash_iterator_remaining(&i);
  mrb_hash_iterator_move_next(&i)
) {
  mrb_value key = mrb_hash_iterator_key(&i);
  mrb_value value = mrb_hash_iterator_value(&i);
  /* ... */
}
```

The code is comprised of `struct mrb_hash_iterator`, 4 accompanying functions and 1 macro with inline documentation.
```
MRB_API mrb_hash_iterator mrb_hash_iterator_new(struct RHash *h);
MRB_API mrb_bool mrb_hash_iterator_move_next(mrb_hash_iterator *it);
MRB_API mrb_value mrb_hash_iterator_key(mrb_hash_iterator *it);
MRB_API mrb_value mrb_hash_iterator_value(mrb_hash_iterator *it);
/* macro */ uint32_t mrb_hash_iterator_remaining(mrb_hash_iterator *it);
```

### DIFF
--- a/mrbgems/mruby-test-hash-iterator/test/iterator.c
+++ b/mrbgems/mruby-test-hash-iterator/test/iterator.c
@@ -4,6 +4,7 @@
 #include <mruby/data.h>
 #include <mruby/hash.h>
 #include <mruby/value.h>
+#include <mruby/variable.h>
 
 struct hash_iterator {
   mrb_value hash;
@@ -13,7 +14,6 @@ struct hash_iterator {
 static void
 hash_iterator_free(mrb_state *mrb, void *p) {
   struct hash_iterator *iterator = (struct hash_iterator *)p;
-  mrb_gc_unregister(mrb, iterator->hash);
   mrb_free(mrb, iterator);
 }
 
@@ -32,7 +32,7 @@ hash_iterator_initialize(mrb_state *mrb, mrb_value self)
   mrb_data_init(self, iterator, &hash_iterator_type);
 
   iterator->hash = mrb_get_arg1(mrb);
-  mrb_gc_register(mrb, iterator->hash);
+  mrb_iv_set(mrb, self, mrb_intern_cstr(mrb, "hash"), iterator->hash);
 
   iterator->it = mrb_hash_iterator_new(mrb_hash_ptr(iterator->hash));
 
--- a/mrbgems/mruby-test-hash-iterator/test/iterator.c
+++ b/mrbgems/mruby-test-hash-iterator/test/iterator.c
@@ -1,24 +1,91 @@
 #include <mruby.h>
+#include <mruby/class.h>
 #include <mruby/compile.h>
+#include <mruby/data.h>
 #include <mruby/hash.h>
 #include <mruby/value.h>
 
-#include <assert.h>
-#include <stdlib.h>
+struct hash_iterator {
+  mrb_value hash;
+  mrb_hash_iterator it;
+};
+
+static void
+hash_iterator_free(mrb_state *mrb, void *p) {
+  struct hash_iterator *iterator = (struct hash_iterator *)p;
+  mrb_gc_unregister(mrb, iterator->hash);
+  mrb_free(mrb, iterator);
+}
+
+static mrb_data_type hash_iterator_type = {"HashIterator", hash_iterator_free};
+
+static mrb_value
+hash_iterator_initialize(mrb_state *mrb, mrb_value self)
+{
+  struct hash_iterator *iterator = (struct hash_iterator *)DATA_PTR(self);
+  if (iterator) {
+    hash_iterator_free(mrb, iterator);
+    mrb_data_init(self, NULL, &hash_iterator_type);
+  }
+
+  iterator = (struct hash_iterator *)mrb_malloc(mrb, sizeof(struct hash_iterator));
+  mrb_data_init(self, iterator, &hash_iterator_type);
+
+  iterator->hash = mrb_get_arg1(mrb);
+  mrb_gc_register(mrb, iterator->hash);
+
+  iterator->it = mrb_hash_iterator_new(mrb_hash_ptr(iterator->hash));
+
+  return self;
+}
+
+static mrb_value
+hash_iterator_move_next(mrb_state *mrb, mrb_value self)
+{
+  struct hash_iterator *iterator = (struct hash_iterator *)DATA_PTR(self);
+  mrb_bool has_next = mrb_hash_iterator_move_next(&iterator->it);
+
+  return mrb_bool_value(has_next);
+}
+
+static mrb_value
+hash_iterator_remaining(mrb_state *mrb, mrb_value self)
+{
+  struct hash_iterator *iterator = (struct hash_iterator *)DATA_PTR(self);
+  uint32_t remaining = mrb_hash_iterator_remaining(&iterator->it);
+
+  return mrb_int_value(mrb, (mrb_int)remaining);
+}
+
+static mrb_value
+hash_iterator_key(mrb_state *mrb, mrb_value self)
+{
+  struct hash_iterator *iterator = (struct hash_iterator *)DATA_PTR(self);
+
+  return mrb_hash_iterator_key(&iterator->it);
+}
+
+static mrb_value
+hash_iterator_value(mrb_state *mrb, mrb_value self)
+{
+  struct hash_iterator *iterator = (struct hash_iterator *)DATA_PTR(self);
+
+  return mrb_hash_iterator_value(&iterator->it);
+}
 
 #define CHECK(...) do { \
     mrb_bool result = (__VA_ARGS__); \
-    assert(result && (#__VA_ARGS__)); \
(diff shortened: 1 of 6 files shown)


juchem:
I've created a test under `test/t/hash_iterator.c` but I'm not sure what's the best way to ensure it runs as part of the test suite.

matz:
If the hash is modified in the loop, the code will crash. Is this fragility OK?

juchem:
@matz isn't the same true for `mrb_hash_foreach`? I mostly copied the logic used in `mrb_hash_foreach` to implement this.

The quirk of invalidation upon modification is inherent to most enumerables.
Some libraries provide a iterator-safe version of modifying operations for this specific reason, like C++'s [`std::remove_if`](https://en.cppreference.com/w/cpp/algorithm/remove) or [`std::vector::erase` that implicitly does a `move_next`](https://en.cppreference.com/w/cpp/container/vector/erase).

I wholeheartedly believe this to be a useful idiom, isomorphic to `foreach` and, therefore, inheriting the same caveats and limitations.

That being said, I respect your authority over `mruby` and understand it is your prerogative to whether accept or reject this proposal.

If there are specific cases that you believe the current implementation doesn't take into account, I'd be more than happy to implement tests for it and fix the implementation if necessary.

If your decision is to not include this at all, I can understand and accept it.

Any any case, thank you for the consideration offered so far.

dearblue:
I think `mrb_hash_foreach()` will raise an exception if there is a major change during the loop, because the `h_check_modified()` macro function is monitoring it.

In the Hash Iterator API, how about putting a prohibition in the documentation against making changes in the middle of the loop?

In my opinion, the API can be left as it is, but depending on the internal implementation, it may be possible to continue processing after the hash object has been modified without crashing.
However, skipped elements and reporting of the same elements are expected. Not immediately, but I would like to experiment a bit.

dearblue:
I don't think there is a documented way to write tests for the C API provided by mruby core.

I don't know if this will help you, but I'll give you two examples anyway:

  - A way like `RIStruct`.

    [Write as a test-only mruby gem](https://github.com/mruby/mruby/tree/3.3.0/mrbgems/mruby-test-inline-struct).
    For example, add a gem named `mrbgems/mruby-test-hash-capi`. Maybe another name would be better?

  - A way like `mrb_format()`.

    Put the call from C in the [mrbgems/mruby-test/vformat.c](https://github.com/mruby/mruby/blob/3.3.0/mrbgems/mruby-test/vformat.c) file and write the test in the [test/t/vformat.rb](https://github.com/mruby/mruby/blob/3.3.0/test/t/vformat.rb) file.
    The `mrb_init_test_vformat()` function, an initialization function in the [mrbgems/mruby-test/vformat.c](https://github.com/mruby/mruby/blob/3.3.0/mrbgems/mruby-test/vformat.c) file, is called from `mrb_init_test_driver()` function in the file [mrbgems/mruby-test/driver.c](https://github.com/mruby/mruby/blob/3.3.0/mrbgems/mruby-test/driver.c).

juchem:
@dearblue I appreciate the comprehensive comments. I'll dig some more in the references you provided and will get back here either with changes to the PR or further questions.

src/hash.c
```diff
@@ -1111,6 +1111,52 @@ mrb_hash_foreach(mrb_state *mrb, struct RHash *h, mrb_hash_foreach_func *func, v
...
+  }
+
+  return it;
+}
+
+MRB_API mrb_bool
+mrb_hash_iterator_move_next(mrb_hash_iterator *it)
+{
+  if (0 < it->rem) {
+    while (entry_deleted_p(++it->entry));
```
dearblue, line 1140: If `it->rem` is 1, then `++it->entry` will cause a buffer overrun.
This may be the reason why all VisualC++ 32-bit versions by appveyor are failing.

juchem: I'll look into that. I may have to change the API slightly to make it safer. Will report back soon.

dearblue: I'm sure there are other ways, but as a point of reference, I made the following changes and valgrind no longer reports problems here.

```patch
diff --git a/src/hash.c b/src/hash.c
index eb049d0e7..12b5e0215 100644
--- a/src/hash.c
+++ b/src/hash.c
@@ -1136,10 +1136,9 @@ mrb_hash_iterator_new(struct RHash *h)
 MRB_API mrb_bool
 mrb_hash_iterator_move_next(mrb_hash_iterator *it)
 {
-  if (0 < it->rem) {
+  if (0 < it->rem && 0 < --it->rem) {
     while (entry_deleted_p(++it->entry));
-    --it->rem;
-    return it->rem > 0;
+    return TRUE;
   }

   return FALSE;
```

mrbgems/mruby-test-hash-iterator/test/iterator.c
```diff
@@ -0,0 +1,228 @@
...
+
+struct hash_iterator {
+  mrb_value hash;
+  mrb_hash_iterator it;
+};
+
+static void
+hash_iterator_free(mrb_state *mrb, void *p) {
+  struct hash_iterator *iterator = (struct hash_iterator *)p;
+  mrb_gc_unregister(mrb, iterator->hash);
```
dearblue, line 16: Please remove this `mrb_gc_unregister()`.

dearblue (commented):
I didn't know this before, but using `mrb_gc_unregister()` inside the RData::dfree function may cause use-after-free.

Therefore, instead of using the combination of `mrb_gc_register()` and `mrb_gc_unregister()` in RData objects, replacing it with `mrb_iv_set()` is required.

The scenario for when use-after-free occurs begins with a call to `mrb_close()`.

  - From `mrb_close()`, `mrb_gc_destroy()` is called.
  - The array object used in `mrb_gc_register()` is destroyed.
  - Heap pages containing array objects used in `mrb_gc_register()` are released.
  - The RData::dfree function is called, and `mrb_gc_unregister()` is called.
  - Since the array object for the gc register does not already exist, any attempt to reference it will result in use-after-free.

juchem:
@dearblue the 32 bit targets are still having issues, even after the changes above

dearblue:
To see where the test is failing, run `rake -v test`.
Or directly as `bin/mrbtest -v`.

dearblue:
I found the cause.
The problem I tried to fix in #6338 was also occurring in this PR.
https://ci.appveyor.com/project/dearblue/mruby/builds/50581024/job/7gp0jj98uh01ja0n#L1533

I merged #6338 at my hand and tried it, and the problem no longer occurs.
https://ci.appveyor.com/project/dearblue/mruby/builds/50581058

Please wait until #6338 is accepted or another solution is found.

juchem:
> To see where the test is failing, run `rake -v test`. Or directly as `bin/mrbtest -v`.

I meant the appveyor CI tests. I don't have the same environment in my machine to repro.

juchem:
> I found the cause. The problem I tried to fix in #6338 was also occurring in this PR. https://ci.appveyor.com/project/dearblue/mruby/builds/50581024/job/7gp0jj98uh01ja0n#L1533
> 
> I merged #6338 at my hand and tried it, and the problem no longer occurs. https://ci.appveyor.com/project/dearblue/mruby/builds/50581058
> 
> Please wait until #6338 is accepted or another solution is found.

I've rebased on top of your PR's branch to re-run CI with the fix.

juchem:
rebased to `origin/master` since #6338 has been merged

matz:
Thank you for this PR. Closing without merge after re-reviewing — the safety concern from my September 2024 comment was not resolved, and there has been no activity since 2024-09-13.

Recap: existing `mrb_hash_foreach` is wrapped in `H_CHECK_MODIFIED` (see `src/hash.c`), which snapshots the hash's `flags`, `hsh.ht`, `ht_ea_capa`, and `ht_ea` before iteration and raises `RuntimeError` if any of them changed during the loop — safe failure on mid-iteration modification.

The proposed `mrb_hash_iterator_*` API caches a raw `hash_entry*` at construction time and has no equivalent guard. AR->HT promotion, EA reallocation, EA compaction, or GC between `move_next` calls invalidate that pointer silently — use-after-free instead of a clean exception.

There is also no test in the PR exercising modification during iteration, which is the case where the API is unsafe.

If you would like to revisit, the path would be: store a snapshot of `flags`/`tbl`/`ht_ea_capa`/`ht_ea` inside `mrb_hash_iterator` and validate it in `mrb_hash_iterator_move_next`, plus add a test that mutates the hash mid-iteration and expects `RuntimeError`. Also reconsider exposing `struct hash_entry` (currently private to `src/hash.c`) in the public header, since that pins its layout as ABI.

Thanks again for the contribution.

PR #6358, opened 2024-09-19
gha: add Ubuntu 24.04 to the build

jbampton:
Remove Ubuntu 20.04

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,10 +15,10 @@ jobs:
       max-parallel: 8
       matrix:
         include:
+          - {os: ubuntu-24.04, cc: gcc, cxx: g++}
+          - {os: ubuntu-24.04, cc: clang, cxx: clang++}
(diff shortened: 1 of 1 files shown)


dearblue:
I was curious about the failure of [Build & Test / ubuntu-24.04-gcc](https://github.com/mruby/mruby/actions/runs/10947712388/job/30397065455?pr=6358) and checked.
Since Ubuntu 24.04 seems to use gcc-13.2.0, I tried with gcc-13.3.0 on FreeBSD, and `bin/mrbtest` did indeed crash.
However, it is not reproduced on gcc-12.4.0, gcc-14.2.0, and gcc-devel-15.0.0.s20240901 on FreeBSD, and the test passes without problems when `gcc13 -O2` is set as an optimization switch.

So I suspect that the reason for the failure is a gcc13-specific optimization bug.

I will investigate this a bit further.

dearblue:
@jbampton,
#6363 has been merged. Therefore, it is expected that this PR will pass if rebased.

PR #6362, opened 2024-09-22
gha: add Ubuntu 24.04 clang to the build

jbampton:


matz:
Increasing build matrix is not preferable. Should we remove older versions of Ubuntu?

jbampton:
Closing.  We now have #6358 instead

PR #6364, opened 2024-09-22
docs: fix spelling

jbampton:


dearblue:
I think it would be better to submit this proposal to Upstream (https://github.com/ruby/lrama) first.

matz:
I applied the patch anyway. I still recommend upstream PR.

jbampton:
Done 

https://github.com/ruby/lrama/pull/470

PR #6370, opened 2024-09-26
[limitations.md] note `super` behaviour with named keyword arguments

leviongit:


matz:
I would rather fix this limitation by the next release.

PR #6373, opened 2024-10-06
Add build config for Milk-V Duo (RISC-V Linux) board

vickash:
This RISC-V board comes in variants as low as 64MB RAM, and runs a stripped down version of Linux, so I thought mruby would be a good idea. 

Confirmed the following run on the board (Milk-V Duo 256MB / SG2002 SoC):
- C program using mruby as a library, also cross-compiled on host
- The `mrbc` `mruby` and `mirb` binaries produced

All the standard gems compile, except for the test gem. I get this error:
```
NoMethodError: undefined method `source_exts' for nil:NilClass
```
If there's a way to fix, so I can run the tests on the board too, please let me know.

### DIFF
--- a/build_config/milkv_duo.rb
+++ b/build_config/milkv_duo.rb
@@ -93,7 +93,7 @@ MRuby::CrossBuild.new("milkv_duo") do |conf|
   conf.gem 'mrbgems/mruby-string-ext/'
   conf.gem 'mrbgems/mruby-struct/'
   conf.gem 'mrbgems/mruby-symbol-ext/'
-  conf.gem 'mrbgems/mruby-test-inline-struct/'
+  # conf.gem 'mrbgems/mruby-test-inline-struct/'
   # conf.gem 'mrbgems/mruby-test/'
   conf.gem 'mrbgems/mruby-time/'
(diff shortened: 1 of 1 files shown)


build_config/milkv_duo.rb
```diff
@@ -0,0 +1,100 @@
...
+  conf.gem 'mrbgems/mruby-range-ext/'
+  conf.gem 'mrbgems/mruby-rational/'
+  conf.gem 'mrbgems/mruby-set/'
+  conf.gem 'mrbgems/mruby-sleep/'
+  conf.gem 'mrbgems/mruby-socket/'
+  conf.gem 'mrbgems/mruby-sprintf/'
+  conf.gem 'mrbgems/mruby-string-ext/'
+  conf.gem 'mrbgems/mruby-struct/'
+  conf.gem 'mrbgems/mruby-symbol-ext/'
+  conf.gem 'mrbgems/mruby-test-inline-struct/'
```
matz, line 96: `mruby-test-inline-struct` is only for a test. Probably you don't need it for Milk-V.

vickash: Ok. Commented out.

PR #6384, opened 2024-10-18
docs: fix grammar and spelling

jbampton:


matz:
I think they should be fixed upstream.

jbampton:
> I think they should be fixed upstream.

This is now fixed upstream. Thanks.

PR #6391, opened 2024-10-23
Fixed argument forwarding in `instance_exec`

### LINKED ISSUE #6389: instance_exec named block args don't work properly
Using mruby head the following code correctly prints 1 in cruby but it prints 2 in mruby.

```
instance_exec(a: 1) do |a: 2|
  puts a
end
```

And without the default value it fails altogether

```
instance_exec(a: 1) do |a:|
  puts a
end
```

`instance_exec: missing keyword: a (ArgumentError)`

dearblue:
However, on C, there is no easy way to pass keyword arguments. Therefore, when called `Kernel#instance_exec` on C, keyword arguments are converted to positional arguments. This is a limitation of current mruby.

fixed #6389

---

During our work we found two problems.

One had to do with commit messages, and calling `mrb_exec_irep()` from C did not pass arguments.
The other was that calls like `instance_exec(1, 2, 3, &:puts)` would result in a `__send__: main is not a symbol nor a string (TypeError)` exception.

I will put these together a bit more and add each as a separate issue.

### DIFF
--- a/include/mruby/internal.h
+++ b/include/mruby/internal.h
@@ -195,6 +195,7 @@ mrb_int mrb_ci_bidx(mrb_callinfo *ci);
 mrb_int mrb_ci_nregs(mrb_callinfo *ci);
 mrb_value mrb_exec_irep(mrb_state *mrb, mrb_value self, struct RProc *p);
 mrb_value mrb_obj_instance_eval(mrb_state*, mrb_value);
+mrb_value mrb_object_exec(mrb_state *mrb, mrb_value self, struct RClass *target_class);
 mrb_value mrb_mod_module_eval(mrb_state*, mrb_value);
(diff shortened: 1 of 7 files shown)


matz:
Thank you! The same change is probably needed for Class#class_exec in mruby-class-ext too.

dearblue:
Thank you for the review.
I see, indeed. I will update.

dearblue:
Main changes.

  - Fixed argument transfer in `Module#class_exec` as well.
  - Added a common function `mrb_object_exec()` called from `class_exec` and `instance_exec` in `src/vm.c`.

And the argument transfer problem with `mrb_exec_irep()` called from C. It was not difficult to fix.
It is sent as a separate PR.
If that one is taken in first, the branch that calls `mrb_yield_with_classs()` in `mrb_object_exec()` in this PR becomes unnecessary.

PR #6397, opened 2024-10-31
boxing_no.h: rename BOXNIX_SET_VALUE -> BOXNO_SET_VALUE

wataash:
I want to rename it because I was confused about what `NIX` meant.


matz:
Boxing_no. h was originally boxing_nix. h. When we renamed the header file, we should have renamed macro names as well. I accept the renaming.

PR #6399, opened 2024-10-31
Add macOS 15 to the build

jbampton:
macOS 15 (Sequoia) is now available as a public beta in GitHub Actions

https://github.com/actions/runner-images/issues/10686

https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md


.github/workflows/build.yml
```diff
@@ -20,7 +20,8 @@ jobs:
           - {os: ubuntu-24.04, cc: clang, cxx: clang++}
           - {os: ubuntu-22.04, cc: gcc, cxx: g++}
           - {os: ubuntu-22.04, cc: clang, cxx: clang++}
-          - {os: macos-latest, cc: clang, cxx: clang++}
+          - {os: macos-15, cc: clang, cxx: clang++}
```
matz, line 23: Is it possible to make 2 builds instead of 3 on macos?

PR #6414, opened 2024-11-12
Rehash hash tables after element removal

pvande:
This resolves errors where key updates following a deletion or shift could cause a hash to misreport its size or report duplicate keys.

In particular, insertions in the hash table implementation calculate a slot index based on the hashcode of the key. If that slot index is occupied, then it moves on to check a secondary slot, a tertiary slot, and so on until it finds an inactive slot. If some keys `A` and `B` both map to the same slot index, it could happen that `A` is added, then `B` is added, then `A` is removed from the hash — this leaves the slot that `B` *would* occupy inactive, and a subsequent update to `B` will be treated as an insertion.

This unexpected behavior shouldn't cause issues on hash reads, since the latest value will be seen first, but *does* cause the hash size to be misreported, the set of keys to include a duplicate entry, and [unconfirmed] may leave an unexpected value for `B` in place following a deletion of that key.

The simplest solution to this issue is to `rehash` following any entry removal, and that behavior is alluded to within `h_rehash`, but the additional cost of a rehash on each removal may be unreasonably high. In fact, most hash entries can be safely removed without requiring a rehash — a rehash on removal is only necessary when the removed entry had previously blocked an insertion.

This optimization is implemented by storing a parallel array of slot conflict markers, which are updated when adding data to the hash table, and tested on removal. A more optimal solution would likely involve packing these markers into the existing `ib` array.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -84,6 +84,7 @@ typedef struct hash_table {
   uint32_t ea_capa;
   uint32_t ea_n_used;
 #endif
+  uint32_t *ib_conflict;
   uint32_t ib[];
 } hash_table;
 
@@ -297,6 +298,7 @@ static void ht_init(
 static void ht_set(mrb_state *mrb, struct RHash *h, mrb_value key, mrb_value val);
 static void ht_set_without_ib_adjustment(
   mrb_state *mrb, struct RHash *h, mrb_value key, mrb_value val);
+static void ht_rehash_after_delete(mrb_state *mrb, struct RHash *h, index_buckets_iter *it);
 
 static uint32_t
 next_power2(uint32_t v)
@@ -717,6 +719,22 @@ ib_it_entry(index_buckets_iter *it)
   return ea_get(ht_ea(it->h), it->ea_index);
 }
 
+static mrb_bool
+ib_it_conflict(const index_buckets_iter *it)
+{
+  uint32_t idx = it->bit / 32;
+  uint32_t mask = 1 << (it->bit % 32);
+  return h_ht(it->h)->ib_conflict[idx] & mask;
+}
+
+static void
+ib_it_conflict_mark(const index_buckets_iter *it)
+{
+  uint32_t idx = it->bit / 32;
+  uint32_t mask = 1 << (it->bit % 32);
+  h_ht(it->h)->ib_conflict[idx] |= mask;
+}
+
 static uint32_t
 ib_capa_to_bit(uint32_t capa)
 {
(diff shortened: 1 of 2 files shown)


dearblue:
Well spotted!

This issue also leads to unwanted objects not being recycled until `Hash#rehash` if `Hash#delete` is done.
The following code is executed with the mruby command built with `rake -m MRUBY_CONFIG=host-debug clean all`.

```ruby
hash = 17.times.with_object({}) { |i, h| h[i] = true }
p ObjectSpace.memsize_of_all    # => 58545
hash[10] = "1" * (1 << 20)
hash.delete(0)
hash[10] = true
p ObjectSpace.memsize_of_all    # => 1107170 (before patched)
hash.rehash
p ObjectSpace.memsize_of_all    # => 58229
```

However, I am concerned that this PR patch adds a field to the structure of the hash object.
Furthermore, that it may be `rehash`ed upon deletion, which may affect the handling of keyword arguments.
Since the mruby implementation calls `mrb_hash_delete_key()` in processing keyword arguments.

I don't know if there is another way, but I will explore ways to fix it.

pvande:
@dearblue I am absolutely not married to this particular fix implementation — I was also reticent about adding another struct member, but the two alternatives I readily saw (unconditionally rehashing after removal, or squeezing the flag alongside the slot indices in `ib`) seemed worse (for performance and readability, respectively).

As for the rehash on delete, I've been wracking my brain to avoid that too, to no avail. The only elements impacted are the ones that were displaced from the cleared slot (or from one of the subsequently used slots) when inserted. Maintaining a list of "forwarding addresses" to do a more targeted fix seems worse for space efficiency, but could be faster for large hashes. Trying to dynamically search for a record to pull back seems worse.

mruby does already mark the slot with a tombstone after deletions, so it *is* possible to resolve this issue by treating tombstones as "unwritable" (hash slot usage becomes less optimal, insertion into a hash with all entries deleted *could* trigger a hash reallocation), or by rehashing (and restarting) whenever a write sees a tombstone slot. Each of these has different tradeoffs, but may be more ideal.

pvande:
@dearblue I've opened #6421 as an alternative fix, which may have preferable tradeoffs.

dearblue:
Nice insight.
To be honest, I haven't read `src/hash.c` very hard, so I haven't fully grasped it.

I prefer #6421, but when I tried an extreme microbenchmark, #6421 crashed on `SIGSEGV`.

```ruby
hash = 17.times.with_object({}) { |i, a| a[i] = true }

q = 0
10000.times {
  h = hash.dup
  h.delete q
  h[10 - q] = true
  h[q] = true
  i = 10 - q
}
```

matz:
Do you have error reproducing script?

pvande:
> Do you have error reproducing script?

The two tests included in the PR both reproduce the error.

PR #6420, opened 2024-11-13
Update `labeler.yml`: add label for the `tools` directory

jbampton:
PR #6384 was not auto labeled by the actions bot.

So the `tools` directory needs a label.

We should be consistent and have a label for all files and folders.

After this PR is merged we need to add a new label for `tools` here:

https://github.com/mruby/mruby/labels

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -71,3 +71,9 @@ oss-fuzz:
       - changed-files:
           - any-glob-to-any-file:
               - oss-fuzz/**/*
+
+tools:
+  - any:
+      - changed-files:
+          - any-glob-to-any-file:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -71,3 +71,9 @@ oss-fuzz:
       - changed-files:
           - any-glob-to-any-file:
               - oss-fuzz/**/*
+
```
jbampton, line 74:
```suggestion
```

jbampton:
Hi @matz 

To finish this PR we need to add a label for `tools` on the labels page:

https://github.com/mruby/mruby/labels

PR #6421, opened 2024-11-15
Rehash hashes when insertion encounters tombstones

pvande:
This is an alternative resolution to the issue mentioned in #6414. This may be considered preferable, as it avoids the need for any additional storage, and defers rehashing until it is absolutely necessary.

Specifically, while #6414 triggers a rehash whenever a conflicted slot is removed, this PR defers the rehash until a hash write actually encounters a slot marked as deleted. Since writing to a previously deleted slot may result in redundant data (if the previous occupant of that slot had blocked another insertion), we cannot push the work any later; since knowing whether the slot had previously blocked another insertion requires additional storage, we cannot skip the rehash.

---

A third fix implementation was also considered, in which tombstones were treated as "occupied" for the purposes of hash writes. This would allow us to avoid coupling the intermittent cost of a rehash to either deletion or insertion, by paying the much cost of ongoing hash collision resolution until the hash resized (or was explicitly rehashed). In practice, this is tricky to manage without tracking the number of `empty` slots remaining in the hash or doing a full iteration of the hash first.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -294,9 +294,11 @@ static uint32_t ib_bit_to_capa(uint32_t bit);
 static void ht_init(
   mrb_state *mrb, struct RHash *h, uint32_t size,
   hash_entry *ea, uint32_t ea_capa, hash_table *ht, uint32_t ib_bit);
+static void h_set(mrb_state *mrb, struct RHash *h, mrb_value key, mrb_value val);
 static void ht_set(mrb_state *mrb, struct RHash *h, mrb_value key, mrb_value val);
 static void ht_set_without_ib_adjustment(
   mrb_state *mrb, struct RHash *h, mrb_value key, mrb_value val);
+static void ht_rehash(mrb_state *mrb, struct RHash *h);
 
 static uint32_t
 next_power2(uint32_t v)
@@ -852,7 +854,11 @@ ht_set_without_ib_adjustment(mrb_state *mrb, struct RHash *h,
 {
   mrb_assert(ht_size(h) < ib_bit_to_capa(ib_bit(h)));
   ib_cycle_by_key(mrb, h, key, it, {
-    if (ib_it_active_p(it)) {
+    if (ib_it_deleted_p(it)) {
+      ht_rehash(mrb, h);
+      return h_set(mrb, h, key, val);
+    }
+    else if (ib_it_active_p(it)) {
       if (!obj_eql(mrb, key, ib_it_entry(it)->key, h)) continue;
       ib_it_entry(it)->val = val;
     }
--- a/test/t/hash.rb
+++ b/test/t/hash.rb
@@ -975,3 +975,43 @@ assert('test value omission') do
   y = 2
   assert_equal({x:1, y:2}, {x:, y:})
 end
+
+assert('hash tables are properly rehashed as necessary after #delete') do
+  keys = []
+  20.times { |key| keys << key }
+
+  root = {}
+  keys.each { |key| root[key] = true }
+
+  keys.each do |delete_key|
+    keys.each do |update_key|
+      next if update_key == delete_key
+
+      hash = root.dup
+      hash.delete(delete_key)
+      hash[update_key] = true
+
+      assert_equal(19, hash.size, "Hash size should not grow on update")
+      assert_equal(19, hash.keys.size, "Keys should not change on update")
+    end
+  end
+end
+
+assert('hash tables are properly rehashed as necessary after #shift') do
+  keys = []
+  20.times { |key| keys << key }
+
+  root = {}
+  keys.each { |key| root[key] = true }
+
+  keys.each do |update_key|
+    next if update_key == keys.first
+
+    hash = root.dup
+    hash.shift
+    hash[update_key] = true
+
+    assert_equal(19, hash.size, "Hash size should not grow on update")
+    assert_equal(19, hash.keys.size, "Keys should not change on update")
+  end
+end


pvande:
@dearblue 

> […] #6421 crashed on `SIGSEGV`.

This was triggering when the hash table was down-leveling to an array table after rehash; I've pushed up a fix that makes a recursive call to `h_set`, rather than simply retrying the `ht_set` implementation.

dearblue:
Looks good!
As for performance, it doesn't seem to make much difference if it's just a deletion.

I don't have commit rights, so please wait for matz's decision.

matz:
We have introduced tombstones to avoid rehashing. Evert rehashing allocates the internal table of the same size as the existing one. If we execute rehashing every time we see tombstones in deletion/insertion, I worry about memory consumption. Maybe filling tombstones would work better.

pvande:
@matz Unfortunately, filling tombstones is the current behavior that's causing issues today.

To restate the problem, hash writes that hit tombstones _can_ leave the hash in an inconsistent state.

```
Insert A ──┐          
           │          
         ┌─▼─┬───┬───┐
         │ A │   │   │
         └───┴───┴───┘
                      
Insert B ──┬───┐      
           x   │      
         ┌───┬─▼─┬───┐
         │ A │ B │   │
         └───┴───┴───┘
         ┌───┬───┬───┐
Delete A │ ~ │ B │   │
         └───┴───┴───┘
                      
Update B ──┐  (This is incorrectly treated as an insertion.)
           │          
         ┌─▼─┬───┬───┐
         │ B │ B │   │
         └───┴───┴───┘
```

So far, four different resolutions have been discussed:

* Backfill slots after deletion.
  * This avoids a full rehashing by selectively moving just the elements that were previously displaced.
  * This comes with a storage cost, since each slot needs to track a linked list of forwarding addresses.
  * This *probably* also entirely replaces the need for tombstones.
* Rehash after deleting from a contested slot.
  * We can track the contested state of each slot a cost of 1 bit per slot.
  * Rehashing after every deletion would be simpler, but unreasonably expensive.
  * This is the solution implemented in #6414.
  * This *probably* also entirely replaces the need for tombstones.
* Rehash when insertion encounters a tombstone.
  * This has no additional storage cost.
  * This is the solution implemented in this PR (#6421).
* Make tombstones unwritable.
  * This treats tombstoned slots as thought they're occupied.
    * This implies that the number of empty slots is no longer `capa - size`, and probably should be tracked separately (four bytes).
  * This defers the rehashing work until the hash resizes, when the rehashing would have to happen anyway.
  * This has the drawback that hash growth is harder to predict.
    * This is particularly a problem for hashes with a mix of deletions and insertions.
    * Hashes grow based on `size_at_last_resize + insertions_since_last_resize`; you could delete most of the entries from a hash, and the next insertion would *still* cause the hash to grow.

I'm happy to make an attempt at that last solution, but I *suspect* that any reduction in reallocations will be offset by wasted storage space.

dearblue:
@pvande,
What about the idea of storing the first tombstone location (`index_buckets_iter`) and traversing the bucket to `ib_it_empty_p()`?
If the same key is found before reaching `ib_it_empty_p()`, it will be updated.
If not, and there is a first tombstone location, it will be added there.
Otherwise, it is added at the location of `ib_it_empty_p()`.

dearblue:
> The worst case scenario for that solution is that every slot is non-empty, and there's at least one tombstone — every hash write then becomes `O(n)` until the next rehash or resize. It's an open discussion as to whether that's better or worse than an intermittently slower insert under some use cases.

Before it was replaced by tombstone(s), there were meaningful key(s) lined up there, which means that the cost was more necessary before the removal.
Maybe we don't need to worry about it too much.

pvande:
> Before it was replaced by tombstone(s), there were meaningful key(s) lined up there, which means that the cost was more necessary before the removal.
> 
> Maybe we don't need to worry about it too much.

You're correct that those non-empty slots would previously have implied nearly the same costs for insertions. The difference is that the hash would eventually grow, redistributing the live keys and reducing the likelihood of a collision. The worst case for a hash of live keys is also not as long lived — each insertion needs to test *as most* `size` slots, and it's only that bad when every insertion (miraculously) hashes to the same slot. Removing elements also currently reduces that cost, albeit while causing the error we're trying to cause.

On the other hand, with the proposed "mostly unwritable tombstones", a hash with only 16 live elements *can* cause a traversal of all `capa` slots (if there no empty slots). This state will continue to be in effect until `size == capa`, or an explicit rehash happens. It's also not improved by element removals, meaning that code that averages one-in-one-out may find itself with `O(capa)` insertions indefinitely.

I really can't say which approach is going to have the lowest expected cost for the most common use cases. Each solution discussed so far has a cost for one use case or another. My _suspicion_ is that rehash after insert introduces the lowest *overall* additional cost (as a significant side-effect of a less common case), though it's not hard to find the pathological case for that solution either (e.g. removing and reinserting the same key results in a rehash every time).

dearblue:
Reading the code, I found that indeed `ht_rehash()` always reallocates memory.
https://github.com/mruby/mruby/blob/f27bc6ab3264a300508ad99d57334c5d6ed9b5e1/src/hash.c#L930-L966

So, in addition to https://github.com/mruby/mruby/pull/6421#issuecomment-2480886674 to avoid calling `ht_rehash()`,

  - On update, replace the cached tombstone location, if it exists.
    Previous slots are replaced with tombstones.
  - If there are more than a certain number of tombstones in a row, they will be moved together.

How about these ideas?

But I think this is more in the scope of “optimization” rather than “fixing” the problem...

More to the point, I see why the size is decremented in `ht_rehash()`.
This block may not be necessary once issue #6414 is resolved.
https://github.com/mruby/mruby/blob/f27bc6ab3264a300508ad99d57334c5d6ed9b5e1/src/hash.c#L947-L952

matz:
The bug was fixed. The fix skips tombstones in search for the time being. Probably we will rehash the table in some heuristics.

pvande:
> The bug was fixed. The fix skips tombstones in search for the time being. Probably we will rehash the table in some heuristics.

I'd still need to test it, but the commit in master looks as though it introduces an infinite loop when inserting into a hash with at least one deletion and no empty slots. It's possible that I've missed something, though.

matz:
Since `h_set()` adjusts the table size before entering `ib_it_cycle_by_key()` loop, I think we won't face infinite loop problem in reality.

pvande:
> Since `h_set()` adjusts the table size before entering `ib_it_cycle_by_key()` loop, I think we won't face infinite loop problem in reality.

Looking at it again, I hadn't recognized that `ea_n_used` is effectively `size + deleted`, and that the hash *is* therefore properly tracking the number of `empty` slots. Not sure why I got infinite looping implementing this same change last week, but I have no trouble calling it a skill issue.

Thanks again for looking at this, and for taking the time to talk with me about it!

matz:
The original bug was fixed. Thank you for finding and reporting the bug.

PR #6426, opened 2024-11-18
Add GitHub Actions workflow for coverage reporting

hoshiumiarata:
I experimented with adding a GitHub Actions workflow to report coverage.
The workflow is triggered on every push to the repository, generates a HTML coverage report and uploads it as an artifact.
Additionally, it shows the coverage percentage in the summary.

The decisions I made (not necessarily the best ones):
- It runs only on one OS (Ubuntu). It's possible to run on multiple configurations and then merge the results, but that seems like overkill to me.
- It uses Clang. At first, I tried to use GCC, but it seems that GCC generates invisible branches in the code in some cases. This leads to false positives in the coverage report. Clang takes a different "source code-first" approach, so it is more accurate.
- It uses gcovr. I also experimented with lcov, but it generated errors (it's possible to silence them, but I decided to use gcovr for now).
- It ignores `/test/` directories and `_test.c` files.
- The artifact will be available for ~2 weeks~ 3 days.

Sample summary:
image

Sample coverage report:
https://github.com/mruby/mruby/actions/runs/11910195640

I hope it will be useful, and I would be happy to hear any feedback.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,6 +1,6 @@
 name: Coverage
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   coverage:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -28,10 +28,10 @@ jobs:
         run: |
           {
             echo "# Coverage report"
-            echo "\`\`\`"
+            echo \`\`\`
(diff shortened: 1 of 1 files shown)


hoshiumiarata:
@matz 
Thank you.
>2 weeks may be too long for 3.5 MB files.

I changed it to 3 days.

matz:
Thank you!

PR #6432, opened 2024-11-20
Moving code in macro arguments out of macros

dearblue:
In `src/hash.c`, there are code blocks that are passed as macro arguments. These code blocks are interpreted as part of the macro function, so breakpoints cannot be set in the debugger. Also, the gcov command will aggregate them to the caller, and the code in the block will not be counted.

This patch will prevent them from being interpreted as part of a macro, and thus the aforementioned problems will no longer occur.

---

This patch has a performance impact.
Checked with `valgrind --tool=cachegrind`, `bin/mrbtest` increases the number of instruction cycles by 0.4 %, and the microbenchmark shown at https://github.com/mruby/mruby/pull/6414#issuecomment- 2478972123 increases the number of instruction cycles by about 10 %.

### DIFF
--- a/src/hash.c
+++ b/src/hash.c
@@ -168,55 +168,34 @@ DEFINE_GETTER(h, size, uint32_t, size)                          /* h_size       
 DEFINE_ACCESSOR(h, ht, hash_table*, hsh.ht)                     /* h_ht         h_set_ht         */
 DEFINE_SWITCHER(ht, HT)                                         /* h_ht_on  h_ht_off  h_ht_p     */
 
-#define ea_each_used(ea, n_used, entry_var, code) do {                        \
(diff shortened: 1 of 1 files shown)


matz:
Recent commit (inlining h_set_without_ib_adjustment) caused conflict. I will merge this once you have resolved the conflict.

dearblue:
The change is to remove the prototype declaration by moving the definition position of `entry_skip_deleted()`.

The prototype declaration of `ib_it_entry()` added since the last time is necessary for the convenience of the definition position of the newly added `ib_it_find_by_key()`.
Moving `ib_it_find_by_key()` makes the prototype declaration unnecessary, but I think it would be better to be closer to `ib_it_init()`.

matz:
Thank you!

PR #6440, opened 2024-12-01
Fix `redo` keyword

dearblue:
The main purpose is to fix #6439, but includes additional testing and a fix to allow `redo` from nested loopinfo.

The additional test code should be correct as it was also confirmed by “ruby 3.4.0preview2 (2024-10-07 master 32c733f57b) +PRISM [amd64-freebsd14]”.

PR #6442, opened 2024-12-03
Fixed wrong range condition in `OP_JMPUW`

### LINKED ISSUE #6441: `break` inside `while` loop will execute the `ensure` block outside of the `while` loop
It seems that `break` inside a `while` loop will execute the `ensure` block outside of the `while` loop.
For example:

```ruby
begin
  while true
    puts 1
    break
  end
ensure
  puts 2
end
```

mruby (master) will output:

```
1
2
2
```

While Ruby 3.3 will output:

```
1
2
```

dearblue:
fixed #6441

---

Even if merged with #6440, each added test will pass.


matz:
Do we need both #6440 and #6442?

dearblue:
Thank you for your review.
Yes, I consider it necessary. I perceive those issues as separate.

PR #6488, opened 2025-03-09
Fix codespell hook; fix spelling and update codespell ignored words list

jbampton:
After updating from upstream master and running pre-commit locally I found that the codespell hook failed as seen in the image below:

Screenshot from 2025-03-09 18-48-35


grapegirl24:
hi

PR #6498, opened 2025-03-27
Clean up and standardize the pre-commit config

jbampton:


jbampton (commented):
The pre-commit failure for codespell is fixed in #6488

PR #6516, opened 2025-04-24
Make Kernel#` public again

vickash:
I'm updating one of my projects to use 3.4.0, and it uses `` Kernel#` `` to find and load a Linux kernel module. Making `` Kernel#` `` private broke that.

This reverts it to public, so that works again. Not sure if the original change was intentional?


matz:
You can cherrypick the fix. Thank you for the report.

PR #6521, opened 2025-05-02
Add codespell config file `.codespellrc`

jbampton:
Can now run codespell from the command line standalone and it will use our standard config file. Running codespell with pre-commit will also use this config.

Useful when regenerating the ignored words file on the command line.

https://github.com/codespell-project/codespell?tab=readme-ov-file#using-a-config-file

https://github.com/codespell-project/codespell?tab=readme-ov-file#pre-commit-hook


matz:
It looks good to me. Can we merge this?

PR #6524, opened 2025-05-04
Fix Kernel#p when no argument

hasumikin:
## Expected
```ruby
p
=> nil
```

## Actual
```ruby
p
=> []
```


matz:
Hmm, I merged it, but the tests have undesirable side effect (print some values during the tests).

hasumikin:
@matz I also thought that test is not a good implementation, but I couldn't come up with a neat solution. Let's think about it a bit more.

PR #6398, opened 2024-10-31
Test running the Super-Linter with all langauge checks

jbampton:

PR #6527, opened 2025-05-13
Standardize indents in `ruby.proto`

jbampton:
The Super-Linter found this issue today when I was working on #6398 

So just a minor PR for fixing the indentation


matz:
Only concern is oss-fuzz is a contributed code and I do not maintain.

PR #6533, opened 2025-05-27
Fix calling `extended` callback

katafrakt:
It seems that between 3.4.0 and current HEAD the behaviour of `extend` has changed. I noticed it, because the test for `mruby-class-attribute` [failed when ran against head](https://github.com/katafrakt/mruby-class-attribute/actions/runs/15242364473/job/42981521865?pr=1). `git bisect` pointed me to 6c72f8b378fc8743ec1775b3e2d7dd507ec2b394 and indeed I noticed that the handling of `extended` callback seems to be looking for it in the singleton class of the object which is being extended instead of the extension module.

With this change, tests for `mruby-class-attribute` pass. I also added a test with the PR. Hope it does not break other things.


matz:
It was caused by my mistake in 6c72f8b
Thank you for fixing it.

PR #6544, opened 2025-06-04
Fix build strings that must be mutable

zenspider:
Tested with `RUBYOPT=--enable-frozen-string-literal make`


zenspider:
@matz what is mruby's release schedule like? IOW, when might this fix be released (as well as the lrama update/frozen-string fixes)?

matz:
Usually once a year, March or April. But we don't have any fixed schedule.

PR #6559, opened 2025-06-23
Refactor str_succ_bang for in-place modification

matz:
This commit refactors the str_succ_bang function in mrbgems/mruby-string-ext/src/string.c to perform string succession in-place.

The previous implementation used an intermediate temporary string, which incurred memory and performance overhead. This change aims to reduce memory usage and improve performance by modifying the string buffer directly, using mrb_str_resize and memmove when prepending characters is necessary (e.g., '9' -> '10').

Formatting changes (newlines before 'else') have also been applied as per user request.


matz:
Test fails. Please retry.

PR #6577, opened 2025-07-18
Sharing arrays with `ary_dup()`

dearblue:
Until now, the array entity has always been allocated from the heap and copied. However, this allows the array entity to be shared with the source if possible.


matz:
If the source array is not shared, we have to allocate a shared structure. In addition, if we modify the duped array right after calling #dup, the sharing effort itself is just waste. What do you think? Maybe, as Gemini suggested, we just use #replace, only if the original array is already shared.

PR #6579, opened 2025-07-19
Unified implementation of `BasicObject#instance_eval` and `Module#module_eval`

dearblue:
Their implementation logic is almost identical.


matz:
Thanks for this PR!

Master has moved since July 2025 — `exec_irep` was renamed to `eval_irep` in commit 3d0b5c848a, which made this PR conflict on the helper's body. Rather than asking for a rebase, I picked the change into master as commit cf20687ee7 with you as author and the rename mechanically applied. The structure of your `object_eval` helper is preserved.

Closing in favor of cf20687ee7.

PR #6583, opened 2025-07-25
Add fallback to local build_config.rb before using default configuration

buty4649:
This change allows loading a local `build_config.rb` file when it exists in the current directory, instead of always using the default configuration.

**Use Case:**
I frequently work with multiple mruby projects and consolidate mruby source code. Currently, I need to specify `MRUBY_CONFIG=./build_config.rb` when running rake, which is cumbersome. This change eliminates that manual step.

**Implementation:**
- Checks for `./build_config.rb` before using `build_config/default.rb`

### DIFF
--- a/lib/mruby/build.rb
+++ b/lib/mruby/build.rb
@@ -60,11 +60,7 @@ module MRuby
       def mruby_config_path
         path = ENV['MRUBY_CONFIG'] || ENV['CONFIG']
         if path.nil? || path.empty?
-          if File.file?("./build_config.rb")
-            path = "./build_config.rb"
-          else
-            path = "#{MRUBY_ROOT}/build_config/default.rb"
-          end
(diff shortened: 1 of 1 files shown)


dearblue:
There is also a `build_config.rb` file in the top directory of the mruby repository.
This file warns users to use `build_config/default.rb` and stops the build.

Therefore, some action is required.
The solutions I can think of are:

  - (1) If `<MRUBY_ROOT>/build_config.rb` is used, implicitly load `<MRUBY_ROOT>/build_config/default.rb`.
  - (2) If the current working directory is MRUBY_ROOT, ignore `<MRUBY_ROOT>/build_config.rb`.

I'll vote for (1).

matz:
> There is also a `build_config.rb` file in the top directory of the mruby repository. This file warns users to use `build_config/default.rb` and stops the build.

Correct. I have to fix.

PR #6592, opened 2025-07-28
pre-commit autocreate Markdown report of `rake -T` output

jbampton:
How it Works:

When you attempt to git commit, pre-commit will execute the generate-rake-docs hook.

It will only run this specific hook if you have staged any files that end with .rake (e.g., Rakefile, lib/tasks/my_task.rake).

The Python script (generate_rake_docs.py) will then:

Execute rake -T to get the list of Rake tasks.

Create (or update) doc/guides/RAKE_TASKS.md with this output.

Calculate the SHA256 checksum of RAKE_TASKS.md and save it to doc/guides/checksums/RAKE_TASKS.md.sha256.

After the script finishes, pre-commit automatically checks if any files in your working directory were modified by the hook and are not currently staged.

If RAKE_TASKS.md or its checksum file were updated by the script (meaning they changed since the last commit or are new) and you haven't git added them, pre-commit will fail the commit. It will print a message telling you which files need to be staged.

You then simply git add doc/guides/RAKE_TASKS.md doc/guides/checksums/RAKE_TASKS.md.sha256 and commit again. The hook will run again, find the files are now staged and up-to-date, and allow the commit to proceed.This ensures that your Rake task documentation and its checksum are always in sync with your Rakefiles before a commit is finalized.

Example below for this hook when I changed a Rake file with dummy content the hook failed and the reports were regenerated:

Screenshot from 2025-07-29 00-45-04


matz:
Thank you for this PR. Closing — auto-generated markdown reports with checksum verification have caused friction on this project: regenerated content drifting from the checksum blocks unrelated commits, and the generated artifact must be kept in sync with the source command (`rake -T`) on every change.

For the `rake -T` output specifically, anyone who needs the task list can run the command directly. Maintaining a generated/checksummed markdown copy in the tree adds complexity for limited value. Closing in favor of "run the command when you need the output."

Thanks again for the suggestion.

PR #6603, opened 2025-08-02
Restrict `mrb_obj_alloc()` to accept only classes

dearblue:
Until now, the type tag of the `cls` parameter in `mrb_obj_alloc()` was not limited to `MRB_TT_CLASS`. The reason for this was that when performing `prepend` operation, objects with `MRB_TT_MODULE` or `MRB_TT_SCLASS` type tags were passed as the `cls` parameter.

This patch simplifies `mrb_obj_alloc()` itself by having the calling side perform the necessary processing.

***Compatibility Note***

When users call `mrb_obj_alloc()`, providing something other than a class (e.g., a module) will now cause a `TypeError` exception. If users intentionally want to provide a module or singleton class, they must pass `NULL` as the `cls` parameter and modify the class of the returned object.

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -1998,7 +1998,8 @@ mrb_prepend_module(mrb_state *mrb, struct RClass *c, struct RClass *m)
 {
   mrb_check_frozen(mrb, c);
   if (!(c->flags & MRB_FL_CLASS_IS_PREPENDED)) {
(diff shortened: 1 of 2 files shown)


dearblue:
This patch raises a `TypeError` exception when `#clone` a class that has been `#prepend`.
It needs to mature a bit.

PR #6617, opened 2025-08-26
Fix various minor problems and speed up build

zenspider:
- **.gitignore: build -> /build**
- **Fix some typos in doc/guides/mrbgems.md**
- **Rakefile: make the whole thing parallel unless SERIAL=1**
- **Fix warning about tautological comparison in pack.c**


matz:
Thank you! Claude lint failure seems to be caused by my configuration mistake. Probably we should stop that lint.

dearblue:
@zenspider
I created #6631 to revert the commit that enabled parallel execution of rake by default.
The main reason is stated in the commit message, but there is also an issue with mruby itself during parallel builds (#6630).
If you have any comments, please post them to #6631.

PR #6620, opened 2025-09-09
Implement `to_hash` calling in implicit conversion contexts

leviongit:
Previously mruby did not call `#to_hash` on objects in contexts like the keyword splat operator. This patch implements that feature for feature parity with CRuby. Additionally a call to `mrb_ensure_hash_type` in the implementation of `OP_HASHADD` has been changed to an `mrb_assert` as the only situation in which it could trigger would be if the bytecode itself contradicted its assumptions (a similar `assert` was already present in the implementation of `OP_HASHCAT`).

### DIFF
--- a/src/class.c
+++ b/src/class.c
@@ -1421,15 +1421,16 @@ get_args_v(mrb_state *mrb, mrb_args_format format, void** ptr, va_list *ap)
 
         p = GET_ARG(mrb_value*);
         if (pickarg) {
+          mrb_value arg = *pickarg;
           if (!(altmode && mrb_nil_p(*pickarg))) {
             switch (c) {
             case 'C': ensure_class_type(mrb, *pickarg); break;
-            case 'S': mrb_ensure_string_type(mrb, *pickarg); break;
-            case 'A': mrb_ensure_array_type(mrb, *pickarg); break;
-            case 'H': mrb_ensure_hash_type(mrb, *pickarg); break;
+            case 'S': arg = mrb_ensure_string_type(mrb, *pickarg); break;
+            case 'A': arg = mrb_ensure_array_type(mrb, *pickarg); break;
+            case 'H': arg = mrb_ensure_hash_type(mrb, *pickarg); break;
             }
           }
-          *p = *pickarg;
+          *p = arg;
         }
       }
       break;
--- a/src/hash.c
+++ b/src/hash.c
@@ -1962,7 +1962,7 @@ mrb_hash_merge(mrb_state *mrb, mrb_value hash1, mrb_value hash2)
   struct RHash *h1, *h2;
 
   hash_modify(mrb, hash1);
-  mrb_ensure_hash_type(mrb, hash2);
+  hash2 = mrb_ensure_hash_type(mrb, hash2);
   h1 = mrb_hash_ptr(hash1);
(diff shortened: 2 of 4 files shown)


dearblue:
hold on, hold on
If this patch is to be accepted, we need to change mruby's security strategy first.

As far as I recall, mruby deliberately avoids implicit conversions.
One reason for this relates to security holes.

For example, if an array pointer held by an array object is obtained outside a `for` loop, it's possible for that array object to be modified by an mruby VM call inside the loop, invalidating the array pointer.
This means arbitrary Ruby code could cause a use-after-free.

In the past, similar security issues existed not only in `for` loops but also in the calling context of the `mrb_get_args()` function.
After this was discovered, implicit conversions were completely removed from `mrb_get_args()`.

This applies to hash objects, string objects, and the mruby VM's data stack, and there may be others I am not aware of.
Therefore, caution is required when performing implicit conversions using the `mrb_ensure_*_type()` functions.

leviongit:
> hold on, hold on If this patch is to be accepted, we need to change mruby's security strategy first.
> 
> As far as I recall, mruby deliberately avoids implicit conversions. One reason for this relates to security holes.
> 
> For example, if an array pointer held by an array object is obtained outside a `for` loop, it's possible for that array object to be modified by an mruby VM call inside the loop, invalidating the array pointer. This means arbitrary Ruby code could cause a use-after-free.

I believe that (at least in this very case) this shouldn't happen. Because the objects that require calling `#to_hash` aren't yet constructed, and thus aren't referenceable by user code (I may be wrong though).

> This applies to hash objects, string objects, and the mruby VM's data stack, and there may be others I am not aware of. Therefore, caution is required when performing implicit conversions using the `mrb_ensure_*_type()` functions.

Currently only `mrb_ensure_hash_type` applies additional logic, other `mrb_ensure_*_type` functions remain unchanged, but their return value is used (on that matter, it may be useful to mark some functions with e.g. `__attribute__((nodiscard))` (hidden behind a macro like `MRB_NODISCARD`) for static analysis tools/newer compilers).

leviongit:
> Taking `mrb_get_args()` as an example, the user-defined `#to_hash` method may be called, potentially performing operations that extend the data stack (`mrb->c->stbase`).
> After that, when control returns to `mrb_get_args()`, referencing the `argv` variable obtained at the beginning of the function results in a use-after-free vulnerability.

ah, yes, the relocation of the stack on extension hasn't crossed my mind when i was trying to think of a way this change could impact memory correctness.
 
> If the behavior of `mrb_ensure_hash_type()` is changed, I think other `mrb_ensure_*_type()` functions would also need to be modified for consistency.

i'd defer changing other `mrb_ensure_*_type` functions to also be more in line with cruby until we figure out whether implicit conversions can even be safely achieved with mruby's cconv/vm architecture

leviongit:
an idea, instead of `mrb_callinfo` objects having a `mrb_value *stack` member, they could have a `int32_t stoff` member to be used in conjunction with `mrb_context`'s `stbase` member, s.t. the current value of `mrb_callinfo::stack` could be replaced with the computation of `&mrb->c->stbase[mrb->c->ci->stoff]`
@dearblue do you see any obvious downsides to this approach (except for possibly some slowdowns if the computed values aren't cached)?

PR #6632, opened 2025-09-30
Fix assert_false to match FalseClass properly

buty4649:
Modified `assert_false` to match only `FalseClass`.

#### Notes

While the current implementation works fine for boolean falsiness evaluation, since L171 uses `obj == false` to match against `FalseClass`, I believe `assert_false` should be consistent and only match `FalseClass` specifically.

This change ensures semantic consistency in the assertion behavior.

### DIFF
--- a/test/assert.rb
+++ b/test/assert.rb
@@ -168,10 +168,10 @@ def assert_true(obj, msg = nil, diff = nil)
 end
 
 def assert_false(obj, msg = nil, diff = nil)
-  unless obj == false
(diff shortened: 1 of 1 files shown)


buty4649:
You're absolutely right. I also think assert_false testing for falsy values is more intuitive. I was confused because line 171 was specifically testing for the literal `false` value, as mentioned in the description. Since no change is needed after all, I'll close this PR.

PR #6633, opened 2025-10-01
Fix a heap-buffer-overflow in in str strip! methods.

oliverchang:
This issue was originally discovered by OSS-Fuzz:

https://issues.oss-fuzz.com/issues/428404023

The root cause was that str_strip_bang modified the string content and length in-place but failed to null-terminate the string at its new length.

When this modified, non-null-terminated string was duplicated, the buffer may be resized, dropping the old null terminator (via str_uminus -> mrb_str_dup -> str_replace -> str_share). When this is later passed to mrb_raisef using the %!s format specifier, mrb_vformat called strlen on the underlying non-null terminated buffer pointer.

The fix adds explicit null-termination in str_strip_bang, str_lstrip_bang, and str_rstrip_bang after the string length is updated.


matz:
Thank you. I could not take time to check oss-fuzz recently.

PR #6643, opened 2025-10-16
Fix crash caused by an incorrect node type check in `codegen_masgn`

meder:
Use the dedicated function `get_node_type(t)` to correctly determine the node type, instead of the incorrect `node_to_int(t->car)` check.

OSS-Fuzz issue details:
https://oss-fuzz.com/testcase-detail/4760884737277952
https://g-issues.oss-fuzz.com/issues/449498801


matz:
wait, this pull-request made our tests failed. Let me investigate.

PR #6654, opened 2025-10-25
Replace the `JUMP` macro with `NEXT`

dearblue:
The `JUMP` macro is defined as an alias for `NEXT`. At this point, I don't see much need to strictly distinguish them semantically either.


matz:
It is distinguished for 2 reasons:

- documentation that notes the instruction changed pc position inside
- room to trap jumps in the future

So I think we should keep the JUMP macro.

PR #6671, opened 2025-11-23
Add `mruby/` to the public header file path for mrbgems

dearblue:
Commit f81cadfed5eb1035bef056fc4a440e581733064a identifies this as a local header file, but it is actually externally exposed. This means it is included in the output of `mruby-config --cflags` and is therefore accessible to projects linking with pre-built libmruby.

To avoid conflicts with other projects, we add `mruby/` as a directory prefix.


matz:
By c521d83, header files under `include/` are only visible to gems that depend on them. If you want to expose your API, put the header files under `include/export`.

Issue #6597, opened 2025-07-29
15 top tools from the Ruby ecosystem that can be effectively wrapped as `pre-commit` hooks: Gemini can make mistakes, so double-check it


jbampton:
Based on the search results, here's a list of 15 top tools from the Ruby ecosystem that can be effectively wrapped as `pre-commit` hooks. These tools focus specifically on Ruby code quality, style, security, and related aspects.

The `pre-commit` framework can wrap almost any command-line tool. For many Ruby tools, there are often existing `pre-commit` hook repositories (like `mattlqx/pre-commit-ruby` or `jumanjihouse/pre-commit-hooks`) that provide ready-to-use configurations. If a specific hook isn't available, you can create a `language: system` or `language: ruby` hook.

Here are the tools:

1.  **RuboCop:**
    * **Purpose:** The most popular static code analyzer and code formatter for Ruby, enforcing community style guides and best practices. Highly configurable.
    * **Why for pre-commit:** Ensures consistent code style and catches common errors before commit. Can auto-correct many offenses.

2.  **Brakeman:**
    * **Purpose:** A static analysis security vulnerability scanner for Ruby on Rails applications.
    * **Why for pre-commit:** Catches common security vulnerabilities early in the development cycle, preventing them from reaching the repository.

3.  **Reek:**
    * **Purpose:** Code smell detector for Ruby. It examines Ruby code and reports "code smells" – indicators of potential problems in the code.
    * **Why for pre-commit:** Helps maintain a clean and maintainable codebase by identifying areas for refactoring.

4.  **Standard Ruby (standardrb):**
    * **Purpose:** An opinionated code formatter that combines `RuboCop` with a set of pre-defined rules, aiming for zero configuration. It's an alternative to highly customizing RuboCop.
    * **Why for pre-commit:** Enforces a consistent, opinionated code style with minimal setup, reducing style debates.

5.  **Fasterer:**
    * **Purpose:** Helps to make Ruby code faster by suggesting performance improvements based on patterns from `fast-ruby`.
    * **Why for pre-commit:** Identifies potential performance bottlenecks or less efficient Ruby idioms.

6.  **Bundler-Audit (bundle-audit):**
    * **Purpose:** Checks your `Gemfile.lock` for vulnerable versions of gems. It uses the Ruby Advisory Database.
    * **Why for pre-commit:** Ensures that committed dependency changes do not introduce known security vulnerabilities.

7.  **ERB Linter (erb-lint):**
[...]

PR #6692, opened 2025-12-30
pre-commit add manual hooks bundle `install` and `audit`

jbampton:
refs #6597 

# Ruby Dependency Hooks (Manual)

We use `pre-commit` to manage our local development checks. The hooks for `bundle-install` and `bundle-audit` are configured as **manual stages** to balance security with development speed.

### 1. Zero Friction Workflow
Standard hooks run on every `git commit`. Because `bundle-audit` requires a network connection to update its database, moving it to a `manual` stage ensures your daily commits remain instant.

### 2. Intelligent Triggering
Even though these are manual, they still respect the `files` filter. They will only run if `Gemfile` or `Gemfile.lock` have been modified, preventing wasted time during unrelated code changes.

### 3. Standardized Audits
Defining this in `.pre-commit-config.yaml` ensures every developer uses the same command and updates the `ruby-advisory-db` before scanning, keeping the team in sync.

---

## Usage Instructions

Trigger these before pushing a Pull Request or after modifying dependencies.

### Run all manual hooks
`pre-commit run --hook-stage manual`

### Run only the security audit
`pre-commit run bundle-audit --hook-stage manual`

### Run on all files (CI/CD mode)
`pre-commit run --all-files --hook-stage manual`

---

## Hook Descriptions

| Hook ID | Description |
| :--- | :--- |
| **bundle-install** | Ensures the local environment matches the Gemfile.lock. |
| **bundle-audit** | Checks for known vulnerabilities (CVEs) in your gems. |

### DIFF
--- a/.github/linters/mlc_config.json
+++ b/.github/linters/mlc_config.json
@@ -1,6 +1,9 @@
 {
   "ignorePatterns": [
     {
+      "pattern": "^https://api.star-history.com"
+    },
+    {
       "pattern": "^https://github.com/mruby/mruby/commit/"
     },
     {
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,6 +13,9 @@ jobs:
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
         uses: actions/checkout@v6
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4" # Not needed with a .ruby-version, .tool-versions or mise.toml
       - name: Install
         run: |
           python -m pip install --upgrade pip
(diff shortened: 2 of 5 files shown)


matz:
Thank you for this PR. Closing — mruby is an embeddable C library that uses Bundler only for its build/test tooling, not as a runtime dependency. Adding `bundle install` and `bundle audit` as pre-commit manual hooks ties pre-commit hygiene to the host Ruby/Bundler environment in a way that doesn't fit how mruby is typically developed (often by C-focused contributors who may not have a Ruby toolchain ready).

Security-sensitive `Gemfile` changes are infrequent enough that ad-hoc `bundle audit` runs are sufficient when needed; folding them into pre-commit doesn't earn its complexity. Closing.

Thanks again for the suggestion.

Issue #6698, opened 2026-01-09
Bigint tests fail on architectures other than `x86_64` and `i386`


mkoncek:
After updating from 3.3.0 to 3.4.0, I encountered consistent test failures on other CPU architectures (https://copr.fedorainfracloud.org/coprs/mkoncek/mruby/build/9987571/).
For a specific log, see for example here at the bottom: https://download.copr.fedorainfracloud.org/results/mkoncek/mruby/fedora-rawhide-s390x/09987571-mruby/builder-live.log.gz

I haven't delved into it. I will gladly help with finding the issue.

PR #6700, opened 2026-01-12
Fix float/double pack/unpack on s390x

khasinski:
The `pack_float`, `pack_double`, `unpack_float`, and `unpack_double` functions accessed float/double bytes via a union with `uint8_t array`, assuming `bytes[0]` is always the LSB. This is only true on little-endian hosts.

Fix by using the same bit-shift approach as the integer pack functions (like `pack_quad`, `unpack_quad`). Reinterpret float/double as `uint32`/`uint64` and use shifts to extract/assemble bytes in an endian-independent way.

Tested with QEMU, I do not have a mainframe at home 😅 

Fixes: #6698 (s390x test failures)


mkoncek:
Tested this patch in Fedora Copr and the tests passed: https://copr.fedorainfracloud.org/coprs/mkoncek/mruby/build/10000135/

PR #6712, opened 2026-01-29
Fix build error with `MRB_ARY_NO_EMBED`

dearblue:


matz:
Thanks for this PR!

The original motivation (build error with `MRB_ARY_NO_EMBED`) was already addressed in master via `#ifndef MRB_ARY_NO_EMBED` guards in commit 78658d67e — different style from this PR but same effect.

The remaining changes from this PR are valuable as standalone API additions: I picked them into master as commit f8649031a6 with you as author, namely:

- `ARY_GETMEM(a, ptr, len)` and `RARRAY_GETMEM(a, ptr, len)` (mirroring CRuby's `RSTRING_GETMEM`)
- Typing `ARY_EMBED_PTR` as `((mrb_value*)NULL)` for the no-embed case

The vm.c portion of this PR was dropped since master already uses the `#ifndef` form for those two sites. Closing in favor of f8649031a6.

PR #6742, opened 2026-03-12
pre-commit: remove markdown-link-check hook

matz:
## Summary
- Remove `markdown-link-check` pre-commit hook which causes flaky CI failures due to external websites blocking automated requests (e.g. 403 from doxygen.nl)
- Remove its config file `.github/linters/mlc_config.json`
- Remove `pretty-format-json` and `check-json` hooks since no JSON files remain tracked in the repository

## Test plan
- [ ] Verify pre-commit CI workflow passes without markdown-link-check
- [ ] Verify `check-hooks-apply` meta-hook passes (no dead hooks)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,8 +20,8 @@ repos:
       - id: prettier
         name: run prettier
         description: format files with prettier
-        entry: prettier --write '**/*.json' '**/*.md' '**/*.yaml' '**/*.yml'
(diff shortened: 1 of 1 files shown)


jbampton (commented):
We should remove the `'**/*.json'` from prettier pre-commit:

https://github.com/mruby/mruby/blob/ab378ec01e036c3b2b7f2d6b73785ecf243a25e7/.pre-commit-config.yaml#L23

PR #6783, opened 2026-04-11
gha: pin workflows to hash

jbampton:
https://github.com/actions/cache/commit/668228422ae6a00e4ad889ee87cd7109ec5666a7

https://github.com/actions/checkout/commit/de0fac2e4500dabe0009e67214ff5f5447ce83dd

https://github.com/actions/labeler/commit/634933edcd8ababfe52f92936142cc22ac488b1b

https://github.com/actions/upload-artifact/commit/043fb46d1a93c77aae656e7c1c64a875d1fc6a0a

https://github.com/j178/prek-action/commit/53276d8b0d10f8b6672aa85b4588c6921d0370cc

https://github.com/ls-lint/action/commit/02e380fe8733d499cbfc9e22276de5085508a5bd

https://github.com/super-linter/super-linter/commit/9e863354e3ff62e0727d37183162c4a88873df41

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       LD: ${{ matrix.cc }}
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
       - name: Ruby version
@@ -46,13 +46,13 @@ jobs:
     timeout-minutes: 15
     steps:
       - name: "Checkout ${{ github.ref }} ( ${{ github.sha }} )"
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
       - name: Ruby version
         run: ruby -v
       - name: Cache cosmocc
-        uses: actions/cache@v5
+        uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # v5.0.4
         id: cache-cosmocc
         with:
           path: ~/cosmo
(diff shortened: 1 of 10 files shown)


matz:
Can you explain why we need to pin by hash, not by version?

jbampton:
From Google Gemini below:

---

Pinning GitHub Actions workflows to a specific **SHA-1 commit hash** instead of a mutable tag (like `@v3`) or a branch (like `@main`) is considered a security best practice for several critical reasons.

Here is why this approach is the gold standard for supply chain security:

### 1. Protection Against "Tag Shifting" (Immutability)
In Git, tags are mutable. A maintainer (or a malicious actor who has compromised the maintainer’s account) can move a tag like `v3` to point to a different commit at any time.

If you use `uses: actions/checkout@v3`, you are trusting that `@v3` will always point to the code you expect. If that tag is moved to a version containing a bug or malicious code, your workflow will automatically pull the "poisoned" version the next time it runs. A **commit hash**, however, is an immutable identifier; it can never be changed to point to different code.

### 2. Prevention of Supply Chain Attacks
GitHub Actions often run with high privileges, such as access to your `GITHUB_TOKEN`, secret API keys, or production environments. If a popular third-party action is compromised, an attacker could update the version tag to include code that exfiltrates your secrets.

By pinning to a hash, you ensure that even if the developer's account is hacked and they release a new "v3," your CI/CD pipeline remains locked to the specific, audited version of the code you originally integrated.



### 3. Predictability and Determinism
Software development relies on reproducibility. If a build passes today but fails tomorrow without any changes to your code, it is usually because an external dependency changed. 
* **With Tags:** An action developer might release a "minor" update that accidentally introduces a breaking change.
* **With Hashes:** Your workflow environment remains identical across every run. This makes debugging much easier because you can rule out external action updates as the cause of a failure.

### 4. Meeting Compliance Standards
Security frameworks like **SLSA (Supply-chain Levels for Software Artifacts)** and the **OpenSSF Scorecard** require or strongly recommend pinning dependencies by hash. For organizations seeking high security certifications, hash-pinning is often a mandatory requirement.

---

### The Downside: Maintenance Overhead
The primary drawback of pinning to a hash is that you no longer receive automatic security patches or bug fixes from the action maintainer. 

**The Recommended Compromise:**
To balance security with maintainability, developers often use the following format:
```yaml
- uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
```
* **The Hash:** Provides the actual security lock.
* **The Comment:** Tells human developers which version the hash corresponds to.

To manage the manual updates required by this method, tools like **Dependabot** can be configured to automatically open Pull Requests that update these hashes when new versions of the actions are released.

PR #6793, opened 2026-04-16
readfloat: correctly round fraction via division by exact 10^n

mattn:
`mrb_read_float` used `frac * pow10_negative[n]` to apply the fractional scale. Since `10^-n = 1/(2^n * 5^n)` leaves a factor of `5^n` in the denominator, it cannot be represented as `m * 2^e` and `pow10_negative[n]` is only a rounded approximation — the multiplication leaves up to ~1 ulp of error.

Dividing by `pow10_positive[n]` instead makes the divisor exact for `n <= 22` (there `10^n = 2^n * 5^n` fits in the 52-bit mantissa), leaving the division as the only rounding step — the result is correctly rounded. `"0.3".to_f` now matches the `0.3` literal and `strtod("0.3")` bit-for-bit.

Sorry — this was already wrong in the original `strtod.c` I wrote that this file descends from.

### DIFF
--- a/src/readfloat.c
+++ b/src/readfloat.c
@@ -168,10 +168,11 @@ mrb_read_float(const char *str, char **endp, double *fp)
     res = (double)int_part;
   }
   else {
-    // Fast path: combine integer and fractional parts
+    // Divide by the exact 10^n (exact for n <= 22) rather than multiplying
+    // by the inexact 10^-n, so the fraction is correctly rounded.
     res = (double)int_part;
     if (frac_digits > 0) {
-      res += (double)frac_part * mrb_pow10(-frac_digits);
+      res += (double)frac_part / mrb_pow10(frac_digits);
     }
   }
 


mattn:
Before this change, `"0.3".to_f` produced `0x3FD3333333333334` — one ulp off from the `0.3` literal and libc `strtod("0.3")` (`0x3FD3333333333333`). With this fix they match bit-for-bit.

mattn:
```
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

int main(void) {
  double mul = 3 * 1e-1;          // multiply by inexact 10^-1
  double div = 3 / 1e1;           // divide by exact 10^1
  double ref = strtod("0.3", NULL);  // libc reference

  unsigned long long m, d, r;
  memcpy(&m, &mul, 8);
  memcpy(&d, &div, 8);
  memcpy(&r, &ref, 8);

  printf("3 * 1e-1    = %.17g  %016llx\n", mul, m);
  printf("3 / 1e1     = %.17g  %016llx\n", div, d);
  printf("strtod(0.3) = %.17g  %016llx\n", ref, r);
  return 0;
}

// 3 * 1e-1    = 0.30000000000000004  3fd3333333333334
// 3 / 1e1     = 0.29999999999999999  3fd3333333333333
// strtod(0.3) = 0.29999999999999999  3fd3333333333333
```

PR #6822, opened 2026-05-10
Run prek; remove trailing whitespace; cleanup markdown

jbampton:
Clean up in C files and Markdown.

### DIFF
--- a/mrbgems/hw-gpio/README.md
+++ b/mrbgems/hw-gpio/README.md
@@ -36,14 +36,14 @@ end
 
 These flags can be combined with bitwise OR.
 
-| Flag                | Value  | Description                |
-|---------------------|--------|----------------------------|
-| `GPIO::IN`          | `0x01` | Input mode                 |
-| `GPIO::OUT`         | `0x02` | Output mode                |
-| `GPIO::HIGH_Z`      | `0x04` | High-impedance (tri-state) |
-| `GPIO::PULL_UP`     | `0x08` | Enable pull-up resistor    |
-| `GPIO::PULL_DOWN`   | `0x10` | Enable pull-down resistor  |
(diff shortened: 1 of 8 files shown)


matz:
Thanks for the cleanup. The trailing whitespace removal in `oss-fuzz/` and the markdown table normalization look good.

One small request: please revert the `\*data` escapes in `mrbgems/hw-spi/README.md` (lines for `SPI#write` and `SPI#transfer`). `*data` here is a Ruby splat parameter in the method signature and reads more naturally without the backslash. GFM only treats `*` as italic when paired with a closing `*`, so the escape isn't needed in these positions, and the unescaped form is what Ruby users expect to see in a method signature.

Once that's fixed I'll merge.

jbampton:
All changes in this PR were from auto formatting from the prek hooks including prettier.

The prettier hook has added the backslashes in the markdown.

matz:
Merged. Thanks for the cleanup — the trailing-whitespace removal in `oss-fuzz/` and the markdown table normalization landed as-is. I added a small follow-up commit to revert the prettier-introduced `\*data` escapes on the `SPI#write` / `SPI#transfer` headings; in those positions `*data` is a Ruby splat parameter and reads more naturally without the backslash (GFM only treats `*` as italic when paired with a closing `*`, which never happens in a method signature). mruby doesn't enforce prettier in CI, so the unescaped form stays canonical here.

jbampton:
On master branch `prek run -a --hook-stage manual` is failing because of prettier.

We do run the manual hooks with the GitHub CI and a lot of the previous merged PRs have been failing for this:

https://github.com/mruby/mruby/pulls?q=is%3Apr+is%3Aclosed

PR #6840, opened 2026-05-19
Fix pre-commit manual hooks update `.prettierignore`

jbampton:


matz:
Thanks for catching the recurring failure.

Just fixed it in 94dc602529 (master) by wrapping the two
splat-bearing method signatures in backticks (heading as code),
which lets prettier accept them while keeping the unescaped
`*data` notation I kept in f91936b06e. This way the file
stays under prettier's coverage for future drift.

Closing this one since the failure should be gone on the next
run. Appreciate the attention to the CI noise.

PR #6861, opened 2026-05-26
Add catch/throw to limitations.md

azimux:
Hey hey! I was in here looking for limitations related to something else and noticed catch/throw wasn't in here. So figured I'd toss it in here and model it after the binding entry! Because of that, went ahead and mentioned the stdlib gembox since I saw it in there.

Cheers!

### DIFF
--- a/doc/limitations.md
+++ b/doc/limitations.md
@@ -139,6 +139,12 @@ Behavior of the operator wasn't changed.
 in the `metaprog` gembox). Without this gem, `binding` is not
 available.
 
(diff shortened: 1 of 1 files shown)


matz:
Thanks for thinking about the docs, but I'm going to pass on this one.

mruby implements much of Ruby's standard surface area as mrbgems, so `catch`/`throw` being in `mruby-catch` is just the normal pattern, not a notable limitation. If we listed every feature whose availability depends on a gem, this document would grow without bound and lose its purpose.

The existing `binding` entry is on the edge for the same reason. I'd rather not add peers to it.

Thanks again for taking the time to look around.

PR #6879, opened 2026-06-04
Add alias support for mruby-method and mruby-proc-ext

Asmod4n:
Ive hit a Segfault when using mruby-method with an aliased C mruby function definition.
This fixed it, i also added some tests to verify the fix works.


Asmod4n:
@matz didn't see the issues in the other functions gemini flags.

Wouldn't a global helper method be better here for accessing this which does the alias resolution itself?
That would be out of scope for this PR i guess since it would touch much more code.

matz:
Thanks, this is a clean fix and the tests are thorough. Confirmed the segfault on master and that this resolves it.

On the helper idea: yes, the `while (MRB_PROC_ALIAS_P(p)) p = p->upper` idiom now repeats across several places (`mrb_proc_eql`, `mrb_proc_source_location`, and now arity/parameters), so factoring it into a small internal `mrb_proc_resolve_alias` (or similar) makes sense. I agree it is out of scope for this PR. I will merge this as-is; a follow-up that introduces the helper and converts the existing sites would be welcome.

PR #6915, opened 2026-06-26
mruby-regexp: correct regexp offset conversion for UTF-8 and binary strings

nattzn:
## Summary

This PR follows up on the review feedback from #6913 and fixes regexp offset conversion and binary string handling in `mruby-regexp`.

The main fixes are:

* Return character offsets from `Regexp#=~` when `MRB_UTF8_STRING` is enabled
* Preserve byte-indexed positions for `ASCII-8BIT` / `BINARY` strings
* Make the regexp engine advance byte-by-byte for forced binary strings
* Preserve the existing UTF-8 regexp character behavior for normal non-binary strings
* Preserve the observable mruby behavior for non-`MRB_UTF8_STRING` builds

### Return character offsets from `Regexp#=~`

In `MRB_UTF8_STRING` builds, `Regexp#=~` now converts the internal byte offset to a public character offset before returning it.

Without `MRB_UTF8_STRING`, the existing byte-offset behavior is preserved.

### Keep `Regexp#match(str, pos)` semantics unchanged

`Regexp#match(str, pos)` and `Regexp#match?(str, pos)` keep the same public semantics. This PR fixes the internal position conversion so that `ASCII-8BIT` / `BINARY` strings keep byte-indexed `pos` values instead of being converted as UTF-8 character offsets.

Normal non-binary strings in `MRB_UTF8_STRING` builds continue to use character offsets, and non-`MRB_UTF8_STRING` builds continue to use byte offsets.

### Pass binary string state into the regexp engine

`mrb_re_exec` now receives whether the subject string is binary. This allows the regexp engine to choose the correct advancement behavior:

* Normal non-binary strings: advance by UTF-8 regexp character length
* `ASCII-8BIT` / `BINARY` strings: advance by one byte

This applies to regexp execution paths including:

* Pike VM execution
* Backtracking execution
* `.` matching
* Character classes
* lookahead / lookbehind recursion
* zero-width match advancement in `gsub`

Small internal helpers were added to centralize the byte-vs-UTF-8 behavior:

* `mrb_re_charlen`
* `mrb_re_decode_char`
* `mrb_re_utf8_continuation_p`

## Affected behavior

The table below summarizes the effective units used by the affected APIs and regexp behavior.

Bold entries indicate behavior fixed by this PR.

| Method / behavior                      |   No `MRB_UTF8_STRING` | `MRB_UTF8_STRING` normal non-binary strings | `mruby-encoding` + `ASCII-8BIT` / `BINARY` strings |
| -------------------------------------- | ---------------------: | ------------------------------------------: | -------------------------------------------------: |
| `Regexp#=~` return value               |            byte offset |                        ***character offset*** |                                        byte offset |
| `Regexp#match(str, pos)` `pos`         |            byte offset |                            character offset |                                    ***byte offset*** |
| `Regexp#match?(str, pos)` `pos`        |            byte offset |                            character offset |                                    ***byte offset*** |
| `MatchData#begin(n)` / `#end(n)`       |            byte offset |                            character offset |                                        byte offset |
| `.` / character class advancement      | UTF-8 regexp character |                      UTF-8 regexp character |                                           ***byte*** |
| `String#scan(/./)`                     | UTF-8 regexp character |                      UTF-8 regexp character |                                           ***byte*** |
| `String#sub(/./)` / `String#gsub(/./)` | UTF-8 regexp character |                      UTF-8 regexp character |                                           ***byte*** |
| zero-width match advancement           | UTF-8 regexp character |                      UTF-8 regexp character |                                           ***byte*** |

Note that “UTF-8 regexp character” here means the unit used by the regexp engine while matching, such as how far `.` advances. It is different from public character offsets returned by APIs such as `Regexp#=~` or `MatchData#begin`.

For normal non-binary strings, the existing regexp engine behavior is preserved. For example, `\xC3\xA9` continues to be treated as one regexp character by `.`.

For forced `ASCII-8BIT` / `BINARY` strings, regexp matching now advances byte-by-byte.

## CRuby compatibility

Compatibility was checked against CRuby `4.0.4`.

The checked cases include:

* Normal non-binary strings
* Forced `ASCII-8BIT` / `BINARY` strings
* Out-of-range and negative `pos` behavior

In the checked cases, the `mruby-encoding` build now matches CRuby behavior.

Representative examples:

```ruby
s = "\xC3\xA9x"
bin = s.dup.force_encoding("ASCII-8BIT")

s.scan(/./).map(&:bytes)
# => [[195, 169], [120]]

s.sub(/./, "Z").bytes
# => [90, 120]

bin.scan(/./).map(&:bytes)
# => [[195], [169], [120]]

bin.sub(/./, "Z").bytes
# => [90, 169, 120]

/x/.match(bin, 2).begin(0)
# => 2
```

## Tests

Added tests covering three behavior groups:

* Normal non-binary strings and public character offsets
* Non-`MRB_UTF8_STRING` builds and existing byte-offset behavior
* Forced `ASCII-8BIT` / `BINARY` strings and byte-by-byte regexp advancement

## Verification

Successfully across three configurations:

* no `MRB_UTF8_STRING`, no `mruby-encoding`
* `MRB_UTF8_STRING` enabled, no `mruby-encoding`
* `MRB_UTF8_STRING` enabled with `mruby-encoding`

### DIFF
--- a/mrbgems/mruby-regexp/include/re_internal.h
+++ b/mrbgems/mruby-regexp/include/re_internal.h
@@ -131,16 +131,38 @@ mrb_regexp_pattern* mrb_re_compile(mrb_state *mrb, const char *pattern, mrb_int 
 /* Free a compiled pattern */
 void mrb_re_free(mrb_state *mrb, mrb_regexp_pattern *pat);
 
+/* UTF-8 helpers */
+int mrb_re_utf8_charlen(const char *s, const char *end);
+uint32_t mrb_re_utf8_decode(const char *s, int *len);
+mrb_bool mrb_re_is_word_char(uint32_t c);
+
+static inline int
+mrb_re_charlen(const char *s, const char *end, mrb_bool binary)
+{
+  return binary ? 1 : mrb_re_utf8_charlen(s, end);
+}
+
+static inline uint32_t
+mrb_re_decode_char(const char *s, int *len, mrb_bool binary)
+{
+  if (binary) {
+    if (len) *len = 1;
+    return (uint8_t)*s;
+  }
+  return mrb_re_utf8_decode(s, len);
+}
+
+static inline mrb_bool
+mrb_re_utf8_continuation_p(const char *s)
+{
+  return (((uint8_t)*s & 0xC0) == 0x80);
+}
+
 /* Execute a match.
    Returns number of captures filled (0 = no match).
    captures[2*n] = start, captures[2*n+1] = end for group n. */
 int mrb_re_exec(mrb_state *mrb, const mrb_regexp_pattern *pat,
             const char *str, mrb_int len, mrb_int start,
-            int *captures, int captures_size);
-
-/* UTF-8 helpers */
-int mrb_re_utf8_charlen(const char *s, const char *end);
-uint32_t mrb_re_utf8_decode(const char *s, int *len);
-mrb_bool mrb_re_is_word_char(uint32_t c);
+            int *captures, int captures_size, mrb_bool binary);
 
 #endif /* MRB_RE_INTERNAL_H */
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -37,6 +37,7 @@ class String
     parts = []
     pos = 0
     len = self.bytesize
+    binary = Regexp.__binary_string?(self)
     while pos <= len
       md = pattern.__byte_match(self, pos)
       break unless md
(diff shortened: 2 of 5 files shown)


nattzn:
Addressed three points:

1. Capped `byte_off` to the actual string byte length before scanning in `re_byte_to_char`, preventing
possible out-of-bounds reads for oversized byte offsets.

2. Added a fast path in `re_char_to_byte` for positive `char_off > len`, returning `-1` immediately since a
character offset cannot exceed the byte length.

3. Updated block-form `String#gsub` to use the same binary-aware zero-width advance as the replacement-string path. Binary strings now advance by byte, while UTF-8 strings advance by character. Since I could not find a suitable Ruby-level API for checking the internal binary flag, I added the internal helper `Regexp.__binary_string?(str)`. Added regression tests for `gsub(//)` and `gsub(/(?=.)/)` to keep both forms consistent.

PR #6917, opened 2026-06-28
Add new actions runner images to `build.yml`

jbampton:
Adds macOS 26 and removes macOS 14

https://github.com/actions/runner-images#available-images

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,8 +20,8 @@ jobs:
           - {os: ubuntu-24.04, cc: clang, cxx: clang++}
           - {os: ubuntu-22.04, cc: gcc, cxx: g++}
           - {os: ubuntu-22.04, cc: clang, cxx: clang++}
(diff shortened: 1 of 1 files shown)


matz:
Thanks for keeping our CI images current, John.

One concern: I would rather not let the OS matrix grow without bound. My preference is to keep it at two generations per OS, so when we add a newer image we should drop the oldest one in the same change. For this PR that would mean:

- Ubuntu: add `26.04`, remove `22.04` (keep `26.04` and `24.04`)
- macOS: add `26`, remove `14` (keep `26` and `15`)

Also, `ubuntu-26.04` is still marked as preview/beta in actions/runner-images, so its toolchain can shift or the image can be pulled without notice. I am fine adding `macos-26` now since it is GA, but I would prefer to hold `ubuntu-26.04` until it reaches GA.

Could you update the PR along these lines? Thanks again.

matz:
Merged, thanks for the quick follow-up John. We will add ubuntu-26.04 once it reaches GA.

Issue #6922, opened 2026-07-04
Two scheduler bugs in mruby-task: sleep suspends the wrong task, terminate self-check never fires


ARYA-mgc:
Hi @matz , I'm pretty new to mruby (this is my first time digging into the mruby  codebase), and I ran into some weird task-scheduling behaviorwhile working with mruby-task that led me down a rabbit hole into `task.c`. Found two bugs in the process, filing them together since they're both small and in the same file - I'll open separate PRs per the contributing guide, just wanted to write up what I found first incase I'm missing some context on why this is written the way it is.

---

## Bug 1: `sleep` with no argument suspends the wrong task

`Kernel#sleep` with no arguments is supposed to suspend the *calling* task indefinitely. What it actually suspends is whichever task happens to be at the head of the ready queue, which isn't always the same task.

`mrbgems/mruby-task/src/task.c`, `mrb_f_sleep`:

```c
mrb_int n = mrb_get_args(mrb, "|f", &sec);
if (n == 0) {
  mrb_task *t = q_ready_;   // head of ready queue, not necessarily the caller
  ...
  t->status = MRB_TASK_STATUS_SUSPENDED;
  ...
}
```

`q_ready_` is just the head of the priority-sorted ready list. What you
actually want here is the task that's currently executing, and there's
already a macro for that  - `MRB2TASK(mrb)` - used elsewhere in this
same file for exactly this purpose (e.g. the join-waiter wakeup path).

**How I ended up looking at this:** three tasks, A(50) / B(100) /
C(128), ready queue sorted A→B→C. C is running and calls bare `sleep`
expecting to suspend itself.

- Expected: C suspends.
- Actual: A gets suspended instead, C keeps running.

**Fix I'm proposing:**
```c
mrb_task *t = MRB2TASK(mrb);
```
(plus a guard for `mrb->c == mrb->root_c`, since `MRB2TASK` assumes
`mrb->c` is embedded in an `mrb_task` -  calling it from the root
context would be UB.)

---

## Bug 2: `terminate_task_internal` self-check never actually fires

When a task terminates itself, the scheduler is supposed to trigger a
context switch so the next task runs. The check for that looks like
it's dead code:

```c
mrb_task_disable_irq();
mrb_task_q_delete(mrb, t);       // removed from ready queue
t->status = MRB_TASK_STATUS_DORMANT;
t->c.status = MRB_TASK_STOPPED;
mrb_task_q_insert(mrb, t);        // now in the dormant queue
[...]

PR #6923, opened 2026-07-04
fix(task): self-termination now triggers context switch

ARYA-mgc:
terminate_task_internal() moved the task to the dormant queue before checking `t == q_ready_` for the context-switch decision. Since the task was already removed from the ready queue, the comparison was always false, leaving a self-terminated task running as a zombie until the next natural exit point.

Capture `was_running` before the queue transition and use it for the switch decision.

Part of #6922 (see issue for a related sleep()-suspends-wrong-task bug, filed as a separate PR).

### DIFF
--- a/mrbgems/mruby-task/src/task.c
+++ b/mrbgems/mruby-task/src/task.c
@@ -1450,6 +1450,8 @@ terminate_task_internal(mrb_state *mrb, mrb_task *t)
 {
   if (t->status == MRB_TASK_STATUS_DORMANT) return;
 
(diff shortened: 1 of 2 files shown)


matz:
Thanks for the careful write-up in #6922, this was a clean catch. The `t == q_ready_` check could never be true after the task had already been moved to the dormant queue, so a self-terminated task really did run one extra statement. Capturing `was_running` before the transition matches how `suspend_task_internal` already decides on the switch.

Merged into master as 715689514 (I pushed the reviewed commit directly together with #6924, since the two touched the same tail of test/task.rb). The regression test is in too.

PR #6930, opened 2026-07-08
mruby-task: survive OOM during task context initialization

sylph01:
### Symptom

With `MRB_USE_TASK_SCHEDULER`, if the system is near heap exhaustion when a task's context is (re)initialized, the VM later crashes — or, on embedded targets, silently corrupts memory — at a point far removed from the failing allocation.

### Mechanism

`task_init_context()` allocates the task's VM stack and callinfo stack with `mrb_malloc()`, which raises `NoMemoryError` via longjmp on failure. A task whose (re)initialization failed that way stays linked in the scheduler queues with a zeroed context (`ci == NULL`). `execute_task()` then dereferences `t->c.ci->proc`.

On hosted builds that is a NULL-page segfault. On RP2350 (and other targets where address 0 is readable — e.g. boot ROM), the read *succeeds* and yields ROM bytes as a fake `RProc` pointer; the VM wanders off into wild reads and wild jumps, and the failure surfaces as heap corruption with no visible connection to the OOM that caused it. We observed exactly this in the field on a Raspberry Pi Pico 2 W running PicoRuby/R2P2: near heap exhaustion, every irb command re-initializes a sandbox task's context, and the resulting crashes ranged from allocator HardFaults to jumps through boot-code sequences.

### Fix (two layers)

- `task_init_context()` allocates both buffers with `mrb_malloc_simple()` and commits neither on failure; the task is retired coherently (`MRB_TASK_STATUS_DORMANT` / `MRB_TASK_STOPPED`) *before* `NoMemoryError` is raised, so it can never be scheduled with a half-built context.
- `execute_task()` retires a context-less task instead of dereferencing it — checked before `mrb->c` is switched, so the scheduler's own context stays usable. This makes the scheduler robust even if some other path produces the same state.

### Verification

Found and verified with an allocation-fault-injection harness on the PicoRuby POSIX build under AddressSanitizer (fail every Nth allocation in bursts, so the GC-and-retry in `mrb_realloc_simple` cannot absorb the failure): a deterministic SEGV at `execute_task()`'s `t->c.ci->proc` read after ~5,500 injected `NoMemoryError` unwinds; gone after this change, with the process surviving repeated storms (tasks die with clean `NoMemoryError`, the scheduler and remaining tasks keep running). On hardware (Pico 2 W), a heap-exhaustion repro that previously died in one of three corruption flavors now ends with a clean `NoMemoryError` cascade and no fault-status bits set.

(No portable mrbtest regression test is included: triggering the window requires failing a specific internal allocation. Happy to add one if there's an accepted pattern for OOM injection in the test suite.)

### DIFF
--- a/mrbgems/mruby-task/src/task.c
+++ b/mrbgems/mruby-task/src/task.c
@@ -259,7 +259,14 @@ task_init_context(mrb_state *mrb, mrb_task *t, const struct RProc *proc)
   if (stbase == NULL || cibase == NULL) {
     mrb_free(mrb, stbase);
     mrb_free(mrb, cibase);
-    t->status = MRB_TASK_STATUS_DORMANT;
+    /* Mark only the CONTEXT as stopped. t->status must keep its
+     * current value: q_get_queue() derives a task's queue from
+     * t->status, so flipping it to DORMANT while the task is still
+     * linked in another queue would make every later q_delete search
+     * the wrong list (task_cleanup_if_stopped would then spin on an
+     * unremovable queue head). The scheduler's existing safety nets
+     * see c->status == MRB_TASK_STOPPED, unlink the task from its
+     * true queue, and transition it to DORMANT coherently. */
     c->status = MRB_TASK_STOPPED;
     mrb_exc_raise(mrb, mrb_obj_value(mrb->nomem_err));
   }
--- a/mrbgems/mruby-task/src/task.c
+++ b/mrbgems/mruby-task/src/task.c
@@ -248,7 +248,22 @@ task_init_context(mrb_state *mrb, mrb_task *t, const struct RProc *proc)
   if (proc->body.irep->nregs > slen) {
     slen += proc->body.irep->nregs;
   }
-  c->stbase = (mrb_value*)mrb_malloc(mrb, slen * sizeof(mrb_value));
+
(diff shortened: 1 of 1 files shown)


sylph01:
Good catch — confirmed and fixed in 0e8291690.

The analysis is exactly right: `q_get_queue()` derives a task's queue from `t->status`, so flipping it to `DORMANT` while the task is still linked in its previous queue makes every later `q_delete` search the wrong list — `task_cleanup_if_stopped()` would then return TRUE without unlinking, and the scheduler loop would spin on an unremovable `q_ready_` head (or, via the `execute_task` retire path, the task would be inserted into a second queue while still linked in the first).

The error path now marks only the context (`c->status = MRB_TASK_STOPPED`), as suggested, and leaves the queue transition to the existing safety nets, which unlink via the task's true status and move it to DORMANT coherently. A comment in the error path records the invariant (`t->status` must always match the queue the task is physically linked in). The `execute_task` retire path was already safe on this point — it deletes before changing status — and is unchanged.

Re-verified: full suite green on the POSIX HAL (1854/1854, 0 crash), and the allocation-fault-injection harness that originally found the bug runs its NoMemoryError storms without hangs or sanitizer reports on the updated branch.

PR #6931, opened 2026-07-08
mruby-task: exclude the tick IRQ while GC marks the task queues

sylph01:
### Problem

`mrb_task_mark_all()` walks the four task queues while the scheduler tick (an interrupt on microcontroller HALs, a signal handler on the POSIX HAL) can concurrently relink tasks between queues (`q_waiting_` → `q_ready_` on wakeup). A queue node can be unlinked or relinked mid-walk, so the GC can skip live tasks (whose stacks then get swept while in use) or follow a stale `next` pointer.

### Fix

Hold the tick-IRQ exclusion (`mrb_task_disable_irq()` / `mrb_task_enable_irq()`) across the queue walk in `mrb_task_mark_all()`, the same discipline every queue-mutating path in task.c already follows.

### Notes

Found by code inspection while chasing a heap corruption on PicoRuby/RP2350 (the corruption itself turned out to be a different bug — see the companion PR "survive OOM during task context initialization" — but this window is real and the fix follows the existing locking convention). No behavior change outside the marking phase; full test suite green on the POSIX HAL.

### DIFF
--- a/mrbgems/mruby-task/src/task.c
+++ b/mrbgems/mruby-task/src/task.c
@@ -67,37 +67,6 @@ static const struct mrb_data_type mrb_task_type = {
 };
 
 /*
- * Nesting-safe scheduler-IRQ exclusion.
- *
- * The HAL primitives are not required to nest (sigprocmask on POSIX,
- * plain NVIC enable/disable on microcontrollers): a naive inner
- * enable would reopen the exclusion for its outer section. Nesting
- * does happen — GC marking (mrb_task_mark_all) excludes the tick, and
- * a GC cycle can be triggered by an allocation made inside an
- * already-excluded section (e.g. Task.stat building its result hash).
- * Every exclusion in this file therefore goes through this per-VM
- * depth counter; only the outermost level touches the HAL. The
- * counter is only ever accessed from the VM's own thread — the tick
- * itself never takes the exclusion.
- */
-static void
-task_irq_disable(mrb_state *mrb)
-{
-  if (mrb->task.irq_nesting++ == 0) {
-    mrb_task_disable_irq();
-  }
-}
-
-static void
-task_irq_enable(mrb_state *mrb)
-{
-  mrb_assert(mrb->task.irq_nesting > 0);
-  if (--mrb->task.irq_nesting == 0) {
-    mrb_task_enable_irq();
-  }
-}
-
-/*
  * GC marking function for all tasks
  * Called from gc.c during root_scan_phase
  */
@@ -125,7 +94,7 @@ mrb_task_mark_all(mrb_state *mrb)
      links into garbage (observed on RP2350 as ASCII string bytes where
      next pointers should be). Exclude the scheduler IRQ for the whole
      walk — same family as the gc.iterating guard in vm.c. */
-  task_irq_disable(mrb);
+  mrb_task_excl_enter(mrb);
   for (qi = 0; qi < 4; qi++) {
     mrb_task *t = mrb->task.queues[qi];
     while (t) {
@@ -181,7 +150,7 @@ mrb_task_mark_all(mrb_state *mrb)
       t = t->next;
     }
   }
-  task_irq_enable(mrb);
+  mrb_task_excl_exit(mrb);
 }
 
 /*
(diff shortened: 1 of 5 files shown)


sylph01:
Thanks for the review — the nesting concern was valid, and it turned out to share a root with the Windows CI failures. Both are addressed in 5622d96.

**Windows CI failure:** GC can run before mruby-task's gem init (allocations during earlier gem inits trigger it, and the GC-stress full-debug build makes that deterministic). At that point `mrb_task_mark_all()` took the exclusion before the Windows HAL had initialized its `CRITICAL_SECTION` (that happens in `mrb_hal_task_init`), so `EnterCriticalSection` ran on uninitialized memory and crashed mrbtest. `mrb_task_mark_all()` now returns before touching the HAL when all queues are empty — tasks can only be queued after HAL init, so non-empty queues imply the exclusion exists. (The POSIX HAL survived the same sequence only by accident: `sigprocmask` on a zeroed mask is harmless, which is why this never showed on the Linux/macOS jobs.)

**Nesting (review comment):** confirmed real — `Task.stat` allocates its result hash with the exclusion held, and an allocation there can trigger a GC cycle whose `mrb_task_mark_all()` would have re-enabled the tick for the remainder of `Task.stat`'s critical section on HALs whose primitives don't nest (`sigprocmask`, plain interrupt enable on microcontrollers; the Windows `CRITICAL_SECTION` is recursive and was safe). Rather than special-casing the marking path, every exclusion in task.c now goes through a per-VM depth counter (`task_irq_disable`/`task_irq_enable`); only the outermost level touches the HAL. The counter lives in `mrb_task_state` and is only ever accessed from the VM's own thread — the tick itself never takes the exclusion — and being per-VM rather than static keeps multiple VMs on different threads independent under the Windows HAL's shared lock.

The raw HAL calls left in task_queue.c are allocation-free pointer manipulation (no GC can trigger inside those sections), so they cannot nest; happy to convert them to the counted wrappers too if you'd prefer uniformity.

Full suite green on the POSIX HAL after the change (1854/1854, 0 crash).

hasumikin:
@sylph01 The raw `mrb_task_disable_irq()` / `mrb_task_enable_irq()` calls left in `task_queue.c` look safe today because the protected regions only do queue/list pointer updates and do not allocate, so they should not trigger GC and re-enter `mrb_task_mark_all()`.

That said, now that `task.c` needs nesting-safe IRQ exclusion, should we either expose a small shared helper for counted IRQ exclusion, or leave a short comment in `task_queue.c` explaining why raw HAL calls are intentionally OK there? My concern is mostly future maintenance: a later change that adds an allocation inside one of these sections could silently reintroduce the nesting problem this PR  just fixed.

mrbgems/mruby-task/src/task.c
```diff
@@ -66,6 +66,37 @@ static const struct mrb_data_type mrb_task_type = {
...
+ * depth counter; only the outermost level touches the HAL. The
+ * counter is only ever accessed from the VM's own thread — the tick
+ * itself never takes the exclusion.
+ */
+static void
+task_irq_disable(mrb_state *mrb)
+{
+  if (mrb->task.irq_nesting++ == 0) {
+    mrb_task_disable_irq();
+  }
```
hasumikin: Since `irq_nesting` is a `uint8_t`, would it be worth adding a debug-only guard before incrementing it?
Something like:

```suggestion
  mrb_assert(mrb->task.irq_nesting < UINT8_MAX);
  if (mrb->task.irq_nesting++ == 0) {
    mrb_task_disable_irq();
  }
```

I do not expect this to overflow in normal use, but if a future change accidentally creates unbounded nesting, wrapping from 255 to 0 would incorrectly call the HAL disable path again and later unbalance the enable path. An mrb_assert should be enough here since it only affects debug builds and documents the invariant without adding release-build overhead.

sylph01:
@hasumikin I'll go with a shared nesting-safe IRQ exclusion helper for uniformity. Having the pattern kept throughout the code will help future implementers know about this problem.
Also, to avoid confusing the `task_irq_enable/disable` helper with `mrb_task_enable/disable_irq`, I will change the naming of this to `mrb_task_excl_enter/exit(mrb)`.

update: This change is @ 36d5781.

hasumikin:
LGTM

mrbgems/mruby-task/src/task.c
sylph01: Looks good, let me get this line in manually

PR #6942, opened 2026-07-10
mruby-task: add mrb_hal_task_switch_hook to the HAL contract

sylph01:
## Summary

The scheduler's only platform servicing point today is `mrb_hal_task_idle_cpu()`, which runs when no task is READY. Platforms that must service an external stack from thread context hang that work off the idle hook — and it silently stops running under load.

This PR adds `mrb_hal_task_switch_hook()` to the task HAL: called at the scheduler's control points in thread context, with a cadence bounded by the timeslice regardless of what Ruby code is doing. Ports that need nothing provide an empty function (posix/win/glib implementations included).

## Problem: two loads defeat idle-only servicing

**1. A compute-bound task.** If the ready queue never empties, `mrb_hal_task_idle_cpu()` never runs. Field case: PicoRuby/R2P2 on a Raspberry Pi Pico 2 W (RP2350) drives the CYW43 WiFi chip and lwIP in poll mode, pumping `cyw43_arch_poll()` from the idle hook. lwIP in NO_SYS/poll mode is not interrupt-safe, so the tick interrupt is not an option — the pump *needs* a thread-context beat. With an open TCP connection and a pure-Ruby busy loop, the stack starved: 5/5 host-side probes stalled >25 s (the peer's kernel retransmits kept the connection alive; a liveness gap rather than a crash, but fatal for anything latency-sensitive: ACKs, retransmits, DHCP renewal all stop).

**2. A scheduler-driven GC drain (since #6938).** The pending-GC branch in `mrb_task_run()` loops `mrb_gc_step()` → `continue` without reaching the idle hook. That is the right priority order for GC vs. sleeping, but it makes a garbage-heavy idle phase a second way to never service the platform. Ordinary steps are tightly bounded, but #6938 documents an unbounded worst single step (the gray-stack-overflow rescan), and a long drain is many steps back-to-back either way.

## API

```c
/* task_hal.h */
void mrb_hal_task_switch_hook(mrb_state *mrb);
```

Contract (documented in the header): called each time the scheduler regains control from a task or completes an idle GC step; thread context; must be cheap when there is nothing to service; must NOT sleep or wait for interrupts (that remains `mrb_hal_task_idle_cpu()`'s job).

Call sites:

- `task_run_body()`, after every `execute_task()` return — i.e. on timeslice expiry, task block, or task completion. This bounds the servicing interval by the timeslice (`MRB_TIMESLICE_TICK_COUNT × MRB_TICK_UNIT`; 12 ms in the default config) even under a task that never yields.
- `task_run_body()`, after each `mrb_gc_step()` in the pending-GC drain branch — so a drain phase services the platform at step granularity.
- `mrb_task_run_once()`, after `execute_task()` only. The GC branch there deliberately does not call it: run_once returns to the host after one step, so the host's own event loop is the natural servicing point.

## Relationship to #6938

Complementary. #6938 moves GC work into idle windows; this hook guarantees platform servicing cadence independently of load. The drain-branch call is what keeps them composable: without it, enabling `GC.scheduler_driven` would reintroduce for garbage-heavy idle exactly the starvation this hook removes for compute-bound tasks.

## Verification

- `MRUBY_CONFIG=gctest rake test`: 788/788 OK (exercises the modified loop including the scheduler-driven GC paths; the hook is the posix no-op there).
- Device, compute-bound case, before/after (PicoRuby/R2P2 on Pico 2 W, CYW43 + lwIP poll mode, hook implemented as the cyw43 pump): board runs a 110 s pure-Ruby busy loop (~1.05 M iterations) with an open TCP connection; host pushes 5 payloads into the compute window and measures ACK latency via TCP_INFO. Idle-only pumping: 5/5 pushes stalled >25 s (recovery only after the loop ended). With the hook: 5/5 ACKed sub-second (0.20-0.30 s; one 3.6 s outlier consistent with the radio's known power-management wake bimodality, far below the starvation signature).
- Device, GC-drain case (same board, `GC.scheduler_driven = true`, built with MRB_GC_PROFILE): a workload alternating garbage churn with 30-50 ms sleeps, so idle windows are spent draining GC debt. Host pushed 13 probes across two runs into those windows: 13/13 ACKed in 0.20-0.30 s. GC.stat confirmed the drain branch was hot: prof_step_count=120, prof_sync_count=0 for the 60-cycle run — the idle windows really were mrb_gc_step() time, serviced by the new drain-branch hook call, not ordinary idle_cpu time.

## Notes

- Why not the tick interrupt: the beat must be thread-context — pollable stacks (lwIP NO_SYS being the canonical one) are not IRQ-safe, and a HAL hook in IRQ context would invite exactly the reentrancy bugs the poll model exists to avoid.
- Cost: one direct call per scheduler iteration. Ports with nothing to service provide an empty function; LTO builds can drop the call entirely.

### DIFF
--- a/mrbgems/mruby-task/include/task_hal.h
+++ b/mrbgems/mruby-task/include/task_hal.h
@@ -130,6 +130,24 @@ void mrb_task_disable_irq(void);
 void mrb_hal_task_idle_cpu(mrb_state *mrb);
 
 /**
+ * Called by the scheduler each time it regains control from a task
+ * (timeslice expiry, block, or completion), in thread context.
+ *
+ * Gives the platform a periodic servicing point that fires even when the
+ * ready queue is never empty — mrb_hal_task_idle_cpu() only runs when no
+ * task is ready, so a compute-bound task would otherwise starve anything
+ * the platform services there (e.g. a polled network driver).
+ *
+ * Requirements:
+ * - Must be cheap when there is nothing to service (called every switch)
+ * - Must NOT sleep or wait for interrupts (that is idle_cpu's job)
+ * - Most ports need no servicing: provide an empty implementation
+ *
+ * @param mrb The mruby state (for context, may be unused)
+ */
+void mrb_hal_task_switch_hook(mrb_state *mrb);
+
+/**
  * Sleep for specified microseconds in wall-clock time
  *
  * Called by sleep functions when in root context (not in a task).
--- a/mrbgems/mruby-task/include/task_hal.h
+++ b/mrbgems/mruby-task/include/task_hal.h
@@ -130,22 +130,41 @@ void mrb_task_disable_irq(void);
 void mrb_hal_task_idle_cpu(mrb_state *mrb);
 
 /**
+ * Which scheduler control point is calling mrb_hal_task_switch_hook().
+ *
+ * More reasons may be added as the scheduler grows control points, without
+ * changing the hook's signature.
+ */
+typedef enum mrb_task_switch_reason {
+  /** A task returned control (timeslice expiry, block, or completion) */
+  MRB_TASK_SWITCH_TASK,
+  /** An idle GC step ran in the scheduler's pending-GC drain loop */
+  MRB_TASK_SWITCH_GC_STEP,
+} mrb_task_switch_reason;
+
+/**
  * Called by the scheduler each time it regains control from a task
- * (timeslice expiry, block, or completion), in thread context.
+ * (timeslice expiry, block, or completion) and after each GC step in its
+ * pending-GC drain loop, always in thread context.
  *
  * Gives the platform a periodic servicing point that fires even when the
  * ready queue is never empty — mrb_hal_task_idle_cpu() only runs when no
- * task is ready, so a compute-bound task would otherwise starve anything
- * the platform services there (e.g. a polled network driver).
+ * task is ready, so a compute-bound task (or a long GC drain) would
+ * otherwise starve anything the platform services there (e.g. a polled
+ * network driver).
  *
  * Requirements:
  * - Must be cheap when there is nothing to service (called every switch)
  * - Must NOT sleep or wait for interrupts (that is idle_cpu's job)
  * - Most ports need no servicing: provide an empty implementation
+ * - Ports that don't distinguish control points ignore `reason`; ports
+ *   that branch on it must service unrecognized values as if they were
+ *   MRB_TASK_SWITCH_TASK, so reasons added later never lose servicing
  *
  * @param mrb The mruby state (for context, may be unused)
+ * @param reason The control point being serviced (may be ignored)
  */
-void mrb_hal_task_switch_hook(mrb_state *mrb);
+void mrb_hal_task_switch_hook(mrb_state *mrb, mrb_task_switch_reason reason);
 
 /**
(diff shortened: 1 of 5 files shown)


matz:
Thanks, this is a well-motivated change and the device measurements are convincing. The scheduler-driven GC drain interaction with #6938 is exactly the kind of thing that needed a servicing beat.

One design question before I merge. The hook is called from three points with different semantics (post-`execute_task` in `task_run_body`, the GC-drain branch, and post-`execute_task` in `mrb_task_run_once`), and the callers already treat them differently: `mrb_task_run_once`'s GC branch deliberately skips the hook while `task_run_body`'s does not. Yet the hook itself only receives `mrb`, so a port cannot tell which control point it is servicing from.

For the polled-network use case that is fine, since the port just wants "service now" regardless of reason. My concern is that this is a HAL contract: changing the signature later is a breaking change for every out-of-tree port. If we ever want a port to behave differently per site (for example doing lighter work during a GC drain than after a task switch), we would have to break the contract again to add the discriminator.

Would you consider introducing the hook with a reason argument now, e.g. `void mrb_hal_task_switch_hook(mrb_state *mrb, mrb_task_switch_reason reason)` with an enum like `MRB_TASK_SWITCH_TASK` / `MRB_TASK_SWITCH_GC_STEP`? Ports that do not care just ignore it (`(void)reason;`), and we avoid a future breaking signature change. If you think the beat is fundamentally reason-agnostic and a discriminator would only invite misuse, I am fine keeping it uniform too. What is your take?

sylph01:
@matz Thanks for the review! I agree, and here is the reasoning:

This change in link-time contract keeps out-of-tree ports from breaking. Changing the signature is an obvious break, but adding a second hook function later breaks them just as hard (an unconditional call to a new HAL symbol is an undefined reference for every port that hasn't added it). A new enum value is the only evolution that costs existing ports nothing. And the risk isn't hypothetical — #6938 added exactly such a new control point (the pending-GC drain) while this hook was in flight, which is why the drain-branch call site exists at all.

On your counter-scenario (a discriminator inviting misuse): the header's requirements — cheap when idle, never sleep — apply on every path regardless of reason, so the enum doesn't weaken the contract; it only lets a port specialize. To keep future enum additions genuinely non-breaking, I've also written an explicit rule into the doc comment: ports that branch on the reason must service unrecognized values as if they were `MRB_TASK_SWITCH_TASK`, so a control point added later can never silently lose servicing (which would recreate the exact starvation this hook fixes).

Pushed @ 3a0a3e33f :

- `mrb_task_switch_reason` enum with your two suggested values; task_run_body's drain branch passes `MRB_TASK_SWITCH_GC_STEP`, both post-execute_task sites pass `MRB_TASK_SWITCH_TASK`. `mrb_task_run_once`'s GC branch still deliberately has no call — that asymmetry lives in whether a site calls the hook, and the host loop services between steps there.
- The three in-tree ports gain `(void)reason;` and stay empty.

Tests: gctest config 788/788, host 104/104.

matz:
Merged. Thanks for the thoughtful iteration, and especially for writing the "service unrecognized reasons as MRB_TASK_SWITCH_TASK" rule into the doc comment. That is what makes future enum additions genuinely non-breaking for out-of-tree ports, and it keeps the starvation guarantee intact. Verified locally: gctest 788/788, host mrbtest 2042 OK, bintest 115/115.

Issue #4675, opened 2019-08-29
Replace unsafe functions with safe alternatives?


nsuchy:
I saw recently that Git VCS generates compiler errors if you try to use unsafe C functions. Would a pull request be accepted to add this header and replace the unsafe functions with safe ones?

**banned.h**
```c
# https://github.com/git/git/blob/master/banned.h
#ifndef BANNED_H
#define BANNED_H

/*
 * This header lists functions that have been banned from our code base,
 * because they're too easy to misuse (and even if used correctly,
 * complicate audits). Including this header turns them into compile-time
 * errors.
 */

#define BANNED(func) sorry_##func##_is_a_banned_function

#undef strcpy
#define strcpy(x,y) BANNED(strcpy)
#undef strcat
#define strcat(x,y) BANNED(strcat)
#undef strncpy
#define strncpy(x,y,n) BANNED(strncpy)
#undef strncat
#define strncat(x,y,n) BANNED(strncat)

#undef sprintf
#undef vsprintf
#ifdef HAVE_VARIADIC_MACROS
#define sprintf(...) BANNED(sprintf)
#define vsprintf(...) BANNED(vsprintf)
#else
#define sprintf(buf,fmt,arg) BANNED(sprintf)
#define vsprintf(buf,fmt,arg) BANNED(sprintf)
#endif

#endif /* BANNED_H */
```

matz:
Usually, we ensure the buffer size before using those functions. But we accept PR anyway.

Issue #383, opened 2012-07-18
Reduce strcpy().


monaka:
It's a nitpick.
The place calling strcpy is just one line. It might reduce some bytes to change to memcpy().

```
diff --git a/src/codegen.c b/src/codegen.c
index daa5211..e032ec5 100644
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -553,7 +553,7 @@ attrsym(codegen_scope *s, mrb_sym a)

   name = mrb_sym2name_len(s->mrb, a, &len);
   name2 = codegen_palloc(s, len+1);
-  strcpy(name2, name);
+  memcpy(name2, name, len);
   name2[len] = '=';
   name2[len+1] = '\0';

```

PR #6963, opened 2026-07-23
minor change: Utilize strlcpy and strlcat instead of strcpy and strcat

gemmaro:
This changeset replaces `strcpy` with [`strlcpy`](https://man.openbsd.org/strlcpy) and `strcat` with `strlcat`.

These new functions is size-bounded and eliminate warnings like this (on OpenBSD, for instance):

```
ld: warning: diagnostic.c:85 (mrbgems/mruby-compiler/src/diagnostic.c:85)(diagnostic.o:(mrc_diagnostic_list_append) in archive /home/gemmaro/src/mruby/build/host/lib/libmruby.a): warning: strcpy() is almost always misused, please use strlcpy()
ld: warning: cmdmisc.c:268 (mrbgems/mruby-bin-debugger/tools/mrdb/cmdmisc.c:268)(/home/gemmaro/src/mruby/build/host/mrbgems/mruby-bin-debugger/tools/mrdb/cmdmisc.o:(replace_ext)): warning: strcat() is almost always misused, please use strlcat()
```

Related issues: #383 #4675

Thank you,

### DIFF
--- a/include/mruby/common.h
+++ b/include/mruby/common.h
@@ -122,6 +122,24 @@ MRB_BEGIN_DECL
 # endif
 #endif
 
+/** Declare size-bounded string copying and concatenation functions. */
+// The `__POSIX_VISIBLE` and `__BSD_VISIBLE` conditionals are from
+// OpenBSD's <string.h>.  The former refers to the POSIX.1-2024.
+#if ((defined(__POSIX_VISIBLE) && __POSIX_VISIBLE >= 202405) ||\
+     (defined(__BSD_VISIBLE) && __BSD_VISIBLE)) ||\
+    (defined(__GLIBC__) &&\
+     ((__GLIBC__ == 2 && __GLIBC_MINOR__ >= 38) || __GLIBC__ > 2))
+#define MRB_STRLCPY(dst, src, dstsize) strlcpy((dst), (src), (dstsize))
+#define MRB_STRLCAT(dst, src, dstsize) strlcat((dst), (src), (dstsize))
+#else
+#define MRB_STRLCPY(dst, src, dstsize) snprintf((dst), (dstsize), "%s", (src))
+#define MRB_STRLCAT(dst, src, dstsize) do {\
+    size_t tmp_len = strlen(dst);\
+    if (tmp_len < (dstsize))\
+      snprintf((dst) + tmp_len, (dstsize) - tmp_len, "%s", (src));\
+  } while (0)
+#endif
+
 MRB_END_DECL
 
 #endif  /* MRUBY_COMMON_H */
--- a/mrbgems/mruby-bin-debugger/tools/mrdb/apilist.c
+++ b/mrbgems/mruby-bin-debugger/tools/mrdb/apilist.c
@@ -10,6 +10,7 @@
 #include "apilist.h"
 #include "apistring.h"
 #include <mruby/compile.h>
+#include <mruby/common.h>
 #include <mruby/irep.h>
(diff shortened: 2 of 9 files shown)


matz:
Thank you for the patch, and for caring about bounded string operations. The intent is good, but I cannot merge this as it stands, because `strlcpy` and `strlcat` are not portable enough for mruby's targets.

These two functions are BSD extensions, not part of standard C. They exist on the BSDs and macOS, on musl, and on glibc only since 2.38 (2023). They are absent on older glibc (for example Ubuntu 22.04 ships glibc 2.35), on MSVC, usually on MinGW's msvcrt, and on many embedded libcs. mruby explicitly supports embedded targets, so we cannot assume the C library provides them.

This is not only a link-time issue. mruby compiles C with `-Werror-implicit-function-declaration`, so on any platform whose `<string.h>` does not declare these functions, the build fails at compile time rather than warning. Our CI already covers such platforms (Ubuntu 22.04 gcc/clang, Windows MSVC), and this change touches core files that every build compiles, including embedded ones: `src/numeric.c` (`mrb_float_to_str`) and `mrbgems/mruby-compiler/src/diagnostic.c`.

To take this improvement, mruby first needs its own portable `strlcpy`/`strlcat`: a fallback implementation used when the C library does not provide them, selected by feature detection. Once that is in place, the call sites here can use them safely. Alternatively, the bounded copies can be written with `snprintf`, which is portable.

I would be happy to take a revised version along those lines. Thanks again.

gemmaro:
Thank you for the review!
I added macros `MRB_STRLCPY` and `MRB_STRLCAT` to provide fallbacks via `snprintf`.

Reference links for the conditionals (brief notes in comments):

* `__POSIX_VISIBLE` and `__BSD_VISIBLE`: [OpenBSD's `string.h`](https://github.com/openbsd/src/blob/691001c74c471d66f8f9a8df7f074547fac7c645/include/string.h#L121-L127)
* POSIX.1-2024 issue 8 changes: [the section which has added `strlcpy` and `strlcat`](https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/xrat/V4_xsh_chap01.html#tag_22_01_01_02)
* `__GLIBC__` and `__GLIBC_MINOR__`: [`features.h` of GNU C library](https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=a8aadb9345c2825a7172879e162bd0455c892097;hb=HEAD#l1434)

matz:
Thanks for the report. The linker warnings are legitimate, but I would rather not introduce a `strlcpy`/`strlcat` compatibility layer. The feature-test conditions are fragile (glibc 2.38 only declares them under `_DEFAULT_SOURCE`, so the declaration can still be invisible while the macro takes the native branch), and the `snprintf` fallback pulls the whole formatted-output machinery into builds that are trying to stay small.

Looking at each call site, the destination size is already known at that point, either freshly allocated from a computed length or already length-checked. So a plain `memcpy` with an explicit length removes the warnings, is fully portable, and is cheaper than either alternative. I have pushed that to master in 8ab5fc96d, which covers every site in your patch plus the Windows `wcscat` in mruby-dir.

One of those sites turned out to have a real problem: `mirb_linenoise_completion()` was doing an unchecked `memcpy` into a 1024 byte stack buffer fed from a linenoise line buffer of up to 4096 bytes. Writing the length out explicitly made the missing bound obvious. Thanks for pointing me at this code.

Closing since master now has the fix, but I appreciate you digging into it.

PR #6993, opened 2026-08-02
mruby-regexp: add Symbol#match, #match? and #=~

takumin:
CRuby defines `match`, `match?` and `=~` on Symbol so that a symbol can be
matched against a regexp without spelling out the `to_s`; mruby-regexp
provided none of them, so any regexp use on a symbol raised NoMethodError.

```ruby
:abc.match?(/b/)   # CRuby: true,  mruby: NoMethodError
:abc =~ /b/        # CRuby: 1,     mruby: NoMethodError
```

CRuby implements all three as the String method applied to the symbol's name
(`rb_sym2str` then `rb_str_match_m` / `rb_str_match`), so they delegate to
`to_s` rather than repeat the pattern handling. That inherits the String
pattern compilation, the `pos` argument, the block form, and the `TypeError`
for a String argument to `=~`. `$~` and `$1`-`$9` are set by the engine
itself, so delegating does not lose them.

`Symbol#match` needs the block to survive the delegation, which #6991 made
`String#match` do.

Delegating also inherits one difference from CRuby that is worth naming: for
an argument that is neither a Regexp nor a String, `String#=~` dispatches
`re =~ self`, so `:a =~ nil` raises NoMethodError where CRuby returns nil.
Fixing that belongs to `String#=~`, not to the Symbol wrapper.

The argument type of `match` and `match?` is inherited the same way, and
there it interacts with #6994. `String#match` currently hands `self` to
anything that is not a String, so once `Symbol#match` exists the receiver
and the argument swap places and a symbol pattern quietly matches instead of
raising:

```ruby
:abc.match(:b)   # without #6994: nil       (runs :b.to_s.match("abc"))
                 # with #6994:    TypeError (CRuby raises it too)
:abc.match?(:b)  # likewise false, then TypeError
```

#6994 adds that check in one helper on `String`, which both receivers go
through, so this PR needs no code for it. Merging #6994 first closes the
window; merging this one first opens it for as long as the two are apart.
The tests here therefore assert nothing about the argument type, and the
rows above are covered by #6994's tests instead.

This covers the symbol-on-the-left direction only. The Regexp side still
takes strings only, and rejects symbols in every entry point rather than just
in `#===`:

```ruby
/l/ =~ :hello        # TypeError (CRuby: 2)
/l/.match(:hello)    # TypeError (CRuby: MatchData)
/l/.match?(:hello)   # TypeError (CRuby: true)
/l/ === :hello       # false     (CRuby: true)
[:to_s, :abc].grep(/^to_/)  # [] -- Enumerable#grep goes through Regexp#===
```

`sym[/re/]` is a third gap: `mruby-symbol-ext` already delegates `Symbol#[]`
to `String#slice`, but this gem does not implement the regexp form of
`String#[]` / `#slice`, so `"hello"[/l+/]` does not work either. Both are
fixes on the String and Regexp side, so they are left alone here; the README
Limitations section and the comment in symbol_regexp.rb now spell them out.

`rake test` passes.

## Summary by CodeRabbit

* **New Features**
  * Added regular-expression matching methods to symbols: `match`, `match?`, and `=~`.
  * Supports matching against symbol text, including positions, offsets, blocks, and multibyte symbols.
* **Documentation**
  * Documented supported symbol regular-expression behavior and current limitations.
* **Tests**
  * Added comprehensive coverage for symbol matching, match state, edge cases, and invalid operations.

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -82,6 +82,12 @@ str.gsub(re) { |m| ... }          # replace all with block
 str.scan(re)                      # => array of matches
 str.split(re)                     # => array of parts
 
+# Symbol methods (the String methods applied to the symbol's name)
+sym.match(re)                     # => MatchData or nil
+sym.match(re) { |md| ... }        # => block result, or nil if no match
+sym.match?(re)                    # => true/false
+sym =~ re                         # => index or nil
+
 # Global variables
 $~                                # last MatchData
 ```
@@ -115,6 +121,15 @@ pattern analysis.
   only.
 - **Step limit on backtracking**: Patterns that require the
   backtracking engine are subject to a step limit.
+- **Symbols only on the left of a match**: `sym.match(re)`,
+  `sym.match?(re)` and `sym =~ re` work, but the Regexp side still
+  takes strings only. `re =~ sym`, `re.match(sym)` and
+  `re.match?(sym)` raise TypeError, `re === sym` returns false, and
+  therefore `syms.grep(re)` (which goes through `Regexp#===`) returns
+  `[]`.
+- **No regexp form of `String#[]`**: `str[re]` and `str.slice(re)`
+  are not supported, and neither is `sym[re]`, which delegates to
+  them.
 
 ## Configuration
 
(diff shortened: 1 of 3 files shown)


takumin:
A note on merge order with #6994, which adds a `TypeError` for a non-Regexp
pattern in `String#match` and `#match?`.

The two PRs touch disjoint files, so they merge in either order without a
conflict. The order still matters for behaviour. This PR defines
`Symbol#match`, which gives a symbol pattern a receiver to dispatch to, so
`"abc".match(:b)` starts resolving to `:b.to_s.match("abc")` and quietly
returns nil where it used to raise NoMethodError:

```ruby
"abc".match(:b)  # before:            NoMethodError
                 # after this PR:     nil
                 # after #6994:       TypeError
```

Merging #6994 first avoids that window. Nothing here needs to change either
way, so this is only about which one goes in first.

Issue #7003, opened 2026-08-03
mruby-regexp: String#split dispatches to_int where the rest of mruby does not, and the respond_to? guarding it is redefinable


takumin:
`mruby-regexp`'s `String#split` override converts a non-Integer `limit` by asking the
argument for `to_int`:

```ruby
# mrbgems/mruby-regexp/mrblib/string_regexp.rb:137-144
if limit.respond_to?(:to_int)
  limit = limit.to_int
  unless limit.is_a?(Integer)
    raise TypeError, "no implicit conversion of #{limit.class} to Integer)"
  end
else
  limit = limit.__to_int
end
```

This is the only place in mruby's own Ruby code that dispatches `to_int` as an implicit
conversion. Nothing in core does it, and the C implementation this override replaces does
not either:

```ruby
o = Object.new
def o.to_int
  2
end

"a,b,c".split(",", o)     # ["a", "b,c"]   <- the override
"a,b,c".__split(",", o)   # TypeError      <- the core C split it aliased away
Array.new(o)              # TypeError
[1, 2, 3][o]              # TypeError
"ab" * o                  # TypeError
```

CRuby accepts the object in every one of those. mruby accepts it in exactly one, and that
one is a gem override.

## Why this is worth asking about rather than just fixing

The gem is more permissive than the core it sits on, which is the situation
`mruby/mruby#6994` settled the other way for the *pattern* argument. `String#match` resolves
its pattern with a plain type check rather than `rb_check_string_type`'s `to_str` protocol,
on the grounds that mruby has no implicit String conversion in core at all, and matching
CRuby there would make the gem more permissive than its host. That decision is pinned with a
test (`599856dfe`).

The `to_int` dispatch predates that decision. It arrived in `b61e44b35`
("mruby-regexp: coerce String#split limit argument", 2026-06-25), whose stated purpose was to
stop a non-Integer limit such as `nil` from reaching `limit > 0` and failing with an
unrelated comparison error. Normalizing `nil` is the part that fixed a real bug; accepting
`to_int` came along with it. The `to_str` policy was written afterwards, so nothing was
inconsistent at the time.

## Symptom 1: the guard can be redefined, and only here

`respond_to?` is redefinable, so an argument that claims `to_int` without defining it gets
past the guard and fails inside the conversion:

```ruby
class Liar
  def respond_to?(name, include_all = false)
    true
  end
end

"a,b,c".split(/,/, Liar.new)
[...]

PR #7004, opened 2026-08-03
mruby-regexp: read the real type of `String#split`'s limit

takumin:
Related to #7003, which 8f85a8413 already settled by dropping the `to_int` dispatch. This
pull request does not close that issue. It is the follow-up 8f85a8413 deliberately left in
place: the `is_a?` guard standing in front of the conversion. Rebased onto that commit. The
stray-paren commit is gone, since the branch that raised that message no longer exists.

`String#split`'s `limit` argument is converted in the override this gem installs over the
core method. The conversion is guarded by `limit.is_a?(Integer)`, which asks the argument
for its own type. `is_a?` is redefinable, so an argument claiming to be an Integer skips the
conversion entirely and reaches the split loop as itself.

What happens next depends on which operators that argument answers. With none, the failure
surfaces as a `NoMethodError` naming an internal comparison, which says nothing about the
limit being wrong:

```ruby
class FakeInt
  def is_a?(klass)
    true
  end
end

"a,b,c".split(/,/, FakeInt.new)
# CRuby: TypeError (no implicit conversion of FakeInt into Integer)
# mruby: NoMethodError (undefined method '>' for FakeInt)
```

With `==`, `>` and `-` answered, there is no error at all: the loop runs to completion with
a non-Integer limit and returns a plausible wrong result.

```ruby
class Cmp
  def is_a?(klass)
    true
  end

  def ==(other)
    false
  end

  def >(other)
    true
  end

  def -(other)
    1
  end
end

"a,b,c".split(/,/, Cmp.new)
# CRuby: TypeError (no implicit conversion of Cmp into Integer)
# mruby: ["a", "b,c"]
```

This is the same hole #7001 closed for the *pattern* argument, left open one argument over.

## The two halves of the same method disagree

The conversion sits ahead of the delegation to `__split`, so a String or `nil` pattern runs
through it too. It survives only because the core implementation converts the limit again in
C, where the argument gets no say:

```ruby
"a,b,c".split(",",  FakeInt.new)   # TypeError (FakeInt cannot be converted to Integer)
"a,b,c".split(/,/, FakeInt.new)    # NoMethodError (undefined method '>' for FakeInt)
```

Two calls to one method, one argument, two answers. The regexp path is the one with no C
conversion behind it, and it is the one that gets it wrong.

## Fix

`Module#===` reads the real type and cannot be redefined, substituted for the one `is_a?`
the limit block still has:

```diff
-    if limit_given && !limit.is_a?(Integer)
+    # `is_a?` is redefinable, so a limit claiming to be an Integer would skip
+    # that conversion and reach the arithmetic below as itself. `Module#===`
+    # reads the real type and cannot be redefined.
+    if limit_given && !(Integer === limit)
       limit = limit.__to_int
     end
```

This is the substitution #7001 made for the pattern argument a few lines below, so the two
guards in this method now read the same way. 8f85a8413 collapsed the body behind this guard
to a single `__to_int`, and that call is exactly what a redefined `is_a?` skips.

Dropping the guard altogether and writing `limit = limit.__to_int if limit_given` would
close the same hole with no redefinable call at all. It is not what this does: 8f85a8413
kept that line deliberately, and the substitution above is the shape #7001 already
established for this method. One measured consequence of keeping it is recorded below.

### What keeping the guard leaves behind

A BigInt limit is a real Integer, so it passes the guard under either spelling and never
reaches `__to_int`. The regexp path then runs the loop with it, while the String path
converts it in C and raises. Measured on a plain host build, `mruby-bigint` being part of
`math.gembox` and so of `default.gembox`:

```ruby
big = 2 ** 70
"a,b,c".split(/,/, big)   # mruby: ["a", "b", "c"]   CRuby: RangeError
"a,b,c".split(",",  big)  # mruby: RangeError        CRuby: RangeError
```

This is the same two-halves disagreement described above, on a different argument, and it
predates this pull request: `limit.is_a?(Integer)` answers true for a BigInt exactly as
`Integer === limit` does, so the change neither introduces nor closes it. Removing the guard
would close it, by sending every limit through `mrb_ensure_int_type()`, but that is a
behaviour change on a case no test covers, and a wider question than the redefinable guard
this pull request is about. Recorded here rather than folded in, and happy to file it
separately if it is worth fixing.

## Test

Added to `mrbgems/mruby-regexp/test/regexp.rb`:

```ruby
class StringSplitLimitIsALiar
  def is_a?(klass)
    true
  end
end

class StringSplitLimitComparable
  def is_a?(klass)
    true
  end

  def ==(other)
    false
  end

  def >(other)
    true
  end

  def -(other)
    1
  end
end

assert("String#split limit cannot pose as an Integer") do
  # `is_a?` is redefinable, so a limit claiming to be an Integer used to skip
  # the conversion and reach the split loop as itself.
  assert_raise(TypeError) { "a,b,c".split(/,/, StringSplitLimitIsALiar.new) }
  # The String pattern delegates to __split, which converts the limit again in
  # C, so this one held before the fix too. Asserted so that the two halves of
  # the method stay pinned to the same answer.
  assert_raise(TypeError) { "a,b,c".split(",", StringSplitLimitIsALiar.new) }

  # Answering the operators the loop uses used to produce a wrong result
  # instead of an error.
  assert_raise(TypeError) { "a,b,c".split(/,/, StringSplitLimitComparable.new) }
end
```

The two classes cover distinct pre-fix paths: `StringSplitLimitIsALiar` blew up on
`limit > 0`, `StringSplitLimitComparable` returned a wrong array with no error at all. They
are named top-level classes rather than anonymous ones because that is how this file already
writes an `is_a?` liar, in `StringMatchIsALiar` a few hundred lines above.

The `String#split with regexp limit` test 8f85a8413 rewrote is untouched. The
`respond_to?` liar it added is a different route from the `is_a?` one pinned here.

Verified on a default host build at `8f85a8413`:

- With this commit: 1941 tests, 1923 OK, 0 failures, 0 crashes, plus 105 bintests.
- With the test kept and the guard reverted to `limit.is_a?(Integer)`, `rake test` reports
  `Fail: String#split limit cannot pose as an Integer`, so the test pins the guard rather
  than passing either way.

## Comparison against CRuby 4.0.6

Run side by side on the four cases the guard governs:

| case | mruby at 8f85a8413 | mruby after | CRuby 4.0.6 |
|---|---|---|---|
| `split(/,/, FakeInt.new)` | `NoMethodError` | `TypeError` | `TypeError` |
| `split(",", FakeInt.new)` | `TypeError` | `TypeError` | `TypeError` |
| `split(/,/, Cmp.new)` | `["a", "b,c"]` | `TypeError` | `TypeError` |
| `split(",", Cmp.new)` | `TypeError` | `TypeError` | `TypeError` |

After the change the exception class agrees in every case and the return value agrees in
every case. What is left is wording, from `mrb_ensure_integer_type()`:

```
mruby: TypeError: FakeInt cannot be converted to Integer
CRuby: TypeError: no implicit conversion of FakeInt into Integer
```

That is the wording 8f85a8413 chose deliberately for every non-Integer limit, and this
change does not touch it.

Unaffected and green: an ordinary Integer limit, a Float limit, `nil`, and limits of 0, 1
and -1.

### DIFF
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -138,7 +138,10 @@ class String
     # `ary[obj]` and `"s" * obj` all reject an object that only defines
     # `to_int`; dispatching it here would leave this the one place in the tree
     # that accepts one, as the same reasoning keeps `match` off `to_str`.
-    if limit_given && !limit.is_a?(Integer)
+    # `is_a?` is redefinable, so a limit claiming to be an Integer would skip
+    # that conversion and reach the arithmetic below as itself. `Module#===`
+    # reads the real type and cannot be redefined.
+    if limit_given && !(Integer === limit)
       limit = limit.__to_int
     end
     # `nil?` and `is_a?` are redefinable, so an argument answering either one
--- a/mrbgems/mruby-regexp/test/regexp.rb
+++ b/mrbgems/mruby-regexp/test/regexp.rb
@@ -1082,6 +1082,44 @@ assert("String#split with regexp limit") do
   assert_raise(TypeError) { "a,b".split(/,/, limit) }
 end
 
+class StringSplitLimitIsALiar
+  def is_a?(klass)
+    true
+  end
+end
+
+class StringSplitLimitComparable
+  def is_a?(klass)
+    true
+  end
+
+  def ==(other)
+    false
+  end
+
+  def >(other)
+    true
+  end
+
+  def -(other)
+    1
+  end
+end
+
+assert("String#split limit cannot pose as an Integer") do
+  # `is_a?` is redefinable, so a limit claiming to be an Integer used to skip
+  # the conversion and reach the split loop as itself.
+  assert_raise(TypeError) { "a,b,c".split(/,/, StringSplitLimitIsALiar.new) }
+  # The String pattern delegates to __split, which converts the limit again in
+  # C, so this one held before the fix too. Asserted so that the two halves of
+  # the method stay pinned to the same answer.
+  assert_raise(TypeError) { "a,b,c".split(",", StringSplitLimitIsALiar.new) }
+
+  # Answering the operators the loop uses used to produce a wrong result
+  # instead of an error.
+  assert_raise(TypeError) { "a,b,c".split(/,/, StringSplitLimitComparable.new) }
+end
+
 assert("String#split with empty regexp") do
   assert_equal ["a", "b", "c"], "abc".split(//)
   assert_equal ["a", "bc"], "abc".split(//, 2)


matz:
This needs a rebase, and the conflict is my doing rather than yours.

I answered #7003 with option 1 and pushed it as 8f85a8413, which removes the `to_int` branch entirely. This PR is based on 0426f7d7b2, from before that, and two of its three hunks edit lines that no longer exist.

I also told you in #7003 that "the `is_a?` one still applies, I left that line alone on purpose so your patch lands on it unchanged". The line does survive, but I was looking at the line and not at your patch, which also touches the block above and below it. That was my mistake, and you had already pushed this before my comment went up.

What is left after the rebase is one line:

```ruby
-    if limit_given && !limit.is_a?(Integer)
+    if limit_given && !(Integer === limit)
```

The other two hunks are gone with the branch that held them: the inner `unless limit.is_a?(Integer)` no longer exists, and the stray paren went with the `raise` beside it.

The fix itself I want. I checked both of your cases on current master and they both still reproduce:

```
FakeInt : NoMethodError: undefined method '>' for FakeInt
Cmp     : ["a", "b,c"]
```

The second one is the one that concerns me. No exception at all, a non-Integer limit driving the loop to completion, and a plausible wrong answer coming back. That is worse than the `NoMethodError`, and it is the same shape as the `sub(:sym) { }` case in #7001: an argument that answers enough of a protocol to keep going, where the honest answer is a `TypeError`. `Module#===` is the right instrument here for the same reason it was there.

Please keep the tests. The `FakeInt` and `Cmp` classes cover something nothing else in the tree does, and they carry across the rebase untouched.

matz:
Merged. Please disregard my previous comment: you had already rebased before I wrote it.

Your rebased commit is dated 08:29 and my comment went up at 09:40. I read a stale head SHA and a stale diff, saw the pre-rebase content, and wrote a whole comment asking you to do something you had finished over an hour earlier. I should have re-fetched before posting; there was no excuse for it, since checking is one command.

The one thing the comment got right is that the rebase left exactly one line, and that is what you had already pushed:

```ruby
-    if limit_given && !limit.is_a?(Integer)
+    if limit_given && !(Integer === limit)
```

Verified on your head (d09f62652): all three cases raise now, and the suite is green (2108 OK, 0 KO).

```
FakeInt(/,/)   -> TypeError: FakeInt cannot be converted to Integer
FakeInt(",")   -> TypeError: FakeInt cannot be converted to Integer
Cmp(/,/)       -> TypeError: Cmp cannot be converted to Integer
```

Thank you for the extra `split(",", ...)` row. Pinning both halves of the method to the same answer is the right instinct, since the String pattern takes the `__split` path and would drift away silently otherwise.

Between #7001, #7003 and this one, redefinable guards have come up three times in a day, each with `Module#===` as the answer. That is worth writing down rather than rediscovering, so I will look at whether the C conventions section is the right place for it.

takumin:
Thank you for reviewing, verifying the behavior on the rebased head, and merging the PR.

I also appreciate the clarification about the stale diff. No problem at all.

Your point about documenting this pattern is especially helpful. Using `Module#===` for guards that must not be influenced by redefined methods seems like a valuable convention to record, given that the same issue appeared across #7001, #7003, and this PR.

PR #7006, opened 2026-08-03
mruby-regexp: add the regexp form of `String#[]` and `String#slice`

takumin:
## Problem

`String#[]` and its alias `String#slice` accept an Integer, a Range and a
String, but not a Regexp. Every regexp form fails with a `TypeError` naming
`Integer`, which does not hint at what is actually missing.

```ruby
"hello"[/l+/]            # CRuby: "ll",  mruby: TypeError (Regexp cannot be converted to Integer)
"hello".slice(/l+/)      # CRuby: "ll",  mruby: TypeError
"hello"[/(l+)(o)/, 1]    # CRuby: "ll",  mruby: TypeError
"hello"[/(?<x>l+)/, :x]  # CRuby: "ll",  mruby: TypeError
"hello"[/z/]             # CRuby: nil,   mruby: TypeError
:hello[/l+/]             # CRuby: "ll",  mruby: TypeError
```

`mrb_str_aref()` funnels every non-String, non-Range argument through an
Integer conversion, and `mruby-regexp` never overrode `[]` or `slice`.

## Solution

Override `[]` in `mrbgems/mruby-regexp/mrblib/string_regexp.rb` and alias
`slice` to it, in the shape `split` already uses in that file: capture the
C-defined method as `__aref`, check the arity, then delegate every argument
list that does not start with a Regexp straight back to C.

The type test is `Regexp === args[0]` rather than `is_a?`, for the reason
the `split` and `=~` comments already give: `is_a?` is redefinable, so an
argument could deny its own type and be read as an index instead, or claim
a type it does not have. `Module#===` reads the real type.

## Behaviour notes

- `$~` is set through `Regexp#match`, including to nil on a failed match,
  which is why `match?` is not used even in the no-capture case.
- The capture argument reaches `MatchData#[]` unchanged. A negative index
  is normalized, an index past the last group answers nil, and a name that
  resolves to no group raises `IndexError`. This matches CRuby, where
  `rb_reg_nth_match()` answers nil and `rb_reg_backref_number()` raises.
  #7000 is what made this delegation correct on its own.
- A failed match answers nil without inspecting the capture argument, as
  `rb_str_subpat()` does.
- The arity check comes before the argument is inspected, so
  `"a"[/a/, 1, 2]` raises `ArgumentError (given 3, expected 1..2)` rather
  than quietly ignoring the extra argument.
- One known divergence remains: a Float capture argument is accepted, since
  `MatchData#[]` reads it through `mrb_as_int()`. This is the same latitude
  the surrounding gem already takes.

## `OP_GETIDX` interaction

`vm_op_getidx()` answers an Integer, String or Range index for a receiver
whose class is exactly String directly from `mrb_str_aref()`, without
consulting the method table, so an override of `String#[]` is not visible
for those three forms. A Regexp index falls through to a send and does
reach the override.

This is safe here because the override only delegates those forms, and the
method comment records the constraint for whoever changes it next.

`str[0]` is the exception, and the one place this change costs anything.
A literal zero index compiles to `OP_GETIDX0`, which has a fast path for
Array and Hash and none for String, so it always sends and therefore does
reach the override. Measured over 1000000 iterations on the built binary,
`s[0]` goes from 48 ms to 193 ms, which is the Ruby frame and the argument
array; 48 ms is what `s.__aref(0)` still costs, and `__aref` is the same C
method `str[0]` reached before. Every other form is unaffected: `str[i]`
with a non-zero or non-literal index is answered by `OP_GETIDX` from C as
before, and `str[i, len]`, `str.slice(...)`, an explicit send and a String
subclass receiver were already sends.

The loops in `gsub` and `split` read one character with `str[0]` on every
iteration, so they now spell it `__aref(0)` and stay on the old path.

## Symbol

`Symbol#[]` and `Symbol#slice` are defined in `mruby-symbol-ext`, not in
core, and delegate to the String methods, so `sym[re]` follows from this
change with no implementation of its own. `mruby-regexp` does not depend on
`mruby-symbol-ext`, so the Symbol test skips itself where that gem is
absent, the same way the `to_enum` test in this gem skips without
`mruby-enumerator`.

## The write side is left alone

`str[re] = repl` and `str.slice!(re)` still raise the `TypeError` this PR
removes from the read side. `String#[]=` is core (`mrb_str_aset_m`) and
`String#slice!` is mruby-string-ext (`mrb_str_slice_bang`, registered as
`MRB_SYM_B(slice)`). Covering the capture form of the assignment would mean
adding C, since `MatchData#__byte_begin` takes an Integer and the name to
group resolution is private to `matchdata_aref()`, and the destructive side
brings the frozen check and the live reference a MatchData keeps to the
string it matched. Both belong with the rest of the destructive-method work,
so the README says so rather than claiming the regexp form is complete.

## Tests

`mrbgems/mruby-regexp/test/regexp.rb` covers the regexp form and its
`slice` twin, a subclass receiver, capture access by index and by name,
negative and out of range indices, an unknown name, a failed match, the
`$~` and `$1` globals on success and on failure, the delegated non-regexp
forms including the subclass path that always goes through the override,
the arity errors, and an argument that lies in `is_a?`.
`mrbgems/mruby-regexp/test/symbol_regexp.rb` covers `sym[re]` behind the
`mruby-symbol-ext` guard.

`rake test` passes.

## Documentation

Narrows the **No regexp form of `String#[]`** item in the gem README's
Limitations section to the write side, adds the forms to the usage examples,
and rewrites the clause in the `symbol_regexp.rb` header comment that
recorded `sym[/re/]` as still missing.

## Summary by CodeRabbit

* **New Features**
  * Added regular expression support to `String#[]` and `String#slice`, including capture and named-capture extraction.
  * Added regular expression slicing support for symbols where symbol slicing is available.
  * Preserved existing behavior for non-regexp indexing and slicing.

* **Documentation**
  * Documented regexp-based string and symbol slicing, including supported reads and unsupported writes.

* **Tests**
  * Added coverage for captures, match globals, validation, subclasses, unmatched patterns, and edge cases.

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -127,6 +127,8 @@ pattern analysis.
   only.
 - **Step limit on backtracking**: Patterns that require the
   backtracking engine are subject to a step limit.
+- **No regexp form on the write side**: `str[re] = repl` and
+  `str.slice!(re)` are not supported. Reading with `str[re]` is.
 
 ## Configuration
 
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -82,12 +82,17 @@ str.gsub(re, replacement)         # replace all occurrences
 str.gsub(re) { |m| ... }          # replace all with block
 str.scan(re)                      # => array of matches
 str.split(re)                     # => array of parts
+str[re]                           # => matched substring or nil
+str[re, capture]                  # => capture by index or name
+str.slice(re)                     # => same as str[re]
 
 # Symbol methods (the String methods applied to the symbol's name)
 sym.match(re)                     # => MatchData or nil
 sym.match(re) { |md| ... }        # => block result, or nil if no match
 sym.match?(re)                    # => true/false
 sym =~ re                         # => index or nil
+sym[re]                           # => matched substring or nil
+                                  #    (Symbol#[] comes from mruby-symbol-ext)
 
 # Global variables
 $~                                # last MatchData
@@ -122,9 +127,6 @@ pattern analysis.
   only.
 - **Step limit on backtracking**: Patterns that require the
   backtracking engine are subject to a step limit.
-- **No regexp form of `String#[]`**: `str[re]` and `str.slice(re)`
-  are not supported, and neither is `sym[re]`, which delegates to
-  them.
 
 ## Configuration
 
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -138,6 +138,12 @@ class String
   # an override is not visible there.  That is harmless as long as this method
   # only delegates them, which is why the regexp branch is the only behaviour
   # added here.
+  #
+  # `str[0]` is the exception: a literal zero index compiles to OP_GETIDX0,
+  # which has no String branch at all and always sends, so it does arrive here
+  # and pays a Ruby frame and an argument array for what used to be a direct
+  # call into C, roughly four times the cost.  That is why the loops in `gsub`
+  # and `split` above read their one character with `__aref(0)`.
   def [](*args)
     # Checked before the argument is inspected, so the non-regexp forms keep
     # reporting the CRuby arity rather than silently ignoring an extra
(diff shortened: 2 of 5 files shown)


takumin:
Thank you for taking a look. I’m going to close this PR for now. I’ll reorganize the individual changes, make their dependencies and intended submission order explicit, and then resubmit them as appropriately scoped PRs. Sorry for the churn.

takumin:
Resubmitted as #7054, with the `OP_GETIDX0` performance concern split out and merged separately as #7040.

PR #7074, opened 2026-08-10
mruby-regexp: read the \u escape

takumin:
The regexp engine has no `\u` escape. `parse_escape()`
(`re_compile.c:339`) knows `\n`, `\t`, `\r`, `\f`, `\v`, `\a`, `\e`, `\b`, the
octal form `\NNN` and the hex form `\xHH`, and returns every other escaped
character as itself (`re_compile.c:389`). `\u` is not among them, so the
backslash is dropped and the `u` becomes an ordinary literal, along with
whatever follows it.

```ruby
"µ"     =~ /\u00b5/     # CRuby: 0,   mruby: nil
"u00b5" =~ /\u00b5/     # CRuby: nil, mruby: 0
"µ"     =~ /\u{b5}/     # CRuby: 0,   mruby: nil
"u{b5}" =~ /\u{b5}/     # CRuby: nil, mruby: 0
"µ"     =~ /[\u00b5]/   # CRuby: 0,   mruby: nil
"あ"    =~ /\u{3042}/   # CRuby: 0,   mruby: nil
"ab"    =~ /\u{61 62}/  # CRuby: 0,   mruby: nil
```

The parser is not involved. `/\u00b5/.source` is `"\\u00b5"` in both, so the
escape reaches the engine intact and the engine is where it is lost.

## Cause

What the leftover text means depends on the spelling, and neither reading is
the one that was written.

`\uXXXX` leaves five ASCII characters. A pattern that means one codepoint
quietly means `u00b5`, and a subject that happens to carry that text matches. A
short `\uXX`, which is a `RegexpError` in CRuby, likewise matches the text
`uXX`.

`\u{...}` leaves a `u` followed by a brace group, so a body that reads as a
repetition count becomes a quantifier on the `u`. `/\u{3042}/` is `u{3042}`,
which matches 3042 `u`s and nothing else, and `/\u{110000}/` raises
`RegexpError: quantifier too large`, naming something the pattern never said. A
body that is not a count stays literal, so `/\u{b5}/` matches `u{b5}`, and
`/\u{}/` matches `u{}` where CRuby raises `invalid Unicode list`.

```ruby
("u" * 3042) =~ /\u{3042}/  # CRuby: nil, mruby: 0
"uXX"        =~ /\uXX/      # CRuby: RegexpError (invalid Unicode escape),
                            # mruby: 0
/\u{110000}/                # CRuby: RegexpError (invalid Unicode range),
                            # mruby: RegexpError (quantifier too large)
```

So the quiet wrong answer is the common case, but it is not the only one: the
`\u{...}` spelling can also fail loudly under a message about a quantifier the
author never wrote.

Neither the README's `Pattern Syntax` list nor its `Limitations` section said
anything about `\u`, and the character escapes that do work (`\n`, `\t`,
`\xHH`, `\NNN`) were not listed either, so the gap was not written down
anywhere.

## Fix

Read the escape. `unicode_escape_first()` takes `\uXXXX` as exactly four hex
digits, or opens a `\u{...}` list whose codepoints `unicode_escape_next()`
yields one at a time, one to six hex digits each and whitespace between them.

A codepoint above 127 has to become its UTF-8 byte sequence, which is a run of
`RE_CHAR`, the same shape `emit_char_bytes()` already emits for a multibyte
character written literally in the pattern (`re_compile.c:681`), and for the
same reason: the run has to be one atom, or a following quantifier binds to the
last byte alone. `emit_codepoint()` is that function for a codepoint the
pattern names rather than spells, so the bytes come from
`mrb_re_utf8_encode()`, the encoding counterpart of the
`mrb_re_utf8_decode()` the gem already has.

The `\u` branch sits ahead of the `ch >= 0xC0` branch `0e96b2c2d` added, which
routes a backslash before a multibyte character away from `parse_escape()`.
`u` is below `0xC0`, so the order is what keeps `\u` from reaching
`parse_escape()` as a literal `u`; the same holds for the `peek(c) < 0xC0`
guard in `read_class_atom()`.

The list form is where the two contexts part.

Outside a character class `/\u{61 62}/` is a sequence of atoms rather than one,
and a quantifier after it repeats the last codepoint only, so `/\u{61 62}+/` is
`a` followed by `b+`. `compile_quantified()` used to take the atom's start from
the position it saved before calling `compile_atom()`; it now takes it from
`c->atom_start`, which the `\u` case moves past every codepoint but the last.
The field is saved and restored around the call, so a nested
`compile_quantified()` inside a group leaves nothing behind for the outer one.

Inside a class each codepoint is a member of its own, and `read_class_atom()`
adds all but the last to the class itself and returns the last, so it can still
open a range: `/[\u{61 62}-z]/` is `a` plus `b-z`, which is how CRuby reads it.
`read_class_atom()` takes the class for that, and `class_add_member()` routes a
member to the ASCII bitmap or the codepoint range list.

The malformed spellings raise rather than fall back on a shorter codepoint or
on literal text, with CRuby's messages: `invalid Unicode escape` for a `\uXXXX`
with fewer than four hex digits, `too short escape sequence` for a bare `\u` at
the end of the pattern, `invalid Unicode list` for `\u{}`, for a separator
CRuby does not take and for an unterminated group, and `invalid Unicode range`
for more than six digits, for a surrogate and for anything above U+10FFFF. A
surrogate and an out of range value have no UTF-8 encoding, so rejecting them
is also what keeps `mrb_re_utf8_encode()` total.

One more pass had to learn about the escape. Whitespace separates the
codepoints of a list, and `preprocess_pattern()` strips whitespace for `/x`
before the parser runs, so it now copies a `\u{...}` group whole; without that
`/\u{61 62}/x` would join into the single codepoint `\u{6162}`.

The executors are not touched. They see the `RE_CHAR` run and the class
members any other pattern produces.

The change comes in two commits: the first lifts out the three pieces the
escape reuses (`emit_char()`, `class_add_member()`, `hex_value()`) with no
behaviour change, the second reads the escape. `emit_char()` sits beside the
`emit_char_bytes()` that `0e96b2c2d` extracted and names the same job for a
character that is a single byte.

## Tests

Three cases in `mrbgems/mruby-regexp/test/regexp.rb`:

- `Regexp - \u escapes`: both spellings against the character and against the
  text they used to match, a codepoint in each UTF-8 length, the quantifier
  binding for a single codepoint and for a list, the `/x` case that the
  free-spacing pass would otherwise join, and the `/i` folding of an ASCII
  letter reached through `\u`.
- `Regexp - \u escapes in a character class`: single members, a codepoint
  range, a list, a list that opens a range, and a negated class.
- `Regexp - malformed \u escapes`: the eleven spellings CRuby refuses, with the
  message asserted for three of them, `\u{110000}` among them, since that is
  the one that used to raise about a quantifier instead.

Patterns are always parsed as UTF-8, so the subjects are written as explicit
bytes and the assertions hold in both `MRB_UTF8_STRING` and byte-string builds.

## Verified on x86_64-linux

- A differential sweep against CRuby 4.0.6 over 273 pattern and flag
  combinations and 4700 answers: 57 patterns under no flag, `/i` and `/x`
  against 20 subjects; 26 patterns exercising the interaction with groups,
  repetition counts, alternation and backreferences against the same subjects;
  27 malformed or boundary patterns under no flag and `/x` against 10 subjects;
  and 22 patterns crossing the escape with the escaped multibyte literal
  `0e96b2c2d` added, with the surrogate boundary and the astral plane, against
  10 subjects. Every answer and every error class and message agrees with
  CRuby. The one difference left is that mruby's `RegexpError` message does not
  carry the flag suffix (`/\uXX/` where CRuby writes `/\uXX/i`), which is how
  `compile_error()` has always formatted a pattern and is unrelated to this
  change.
- `rake test`: 2020 total, 2002 OK, 0 KO, 0 crash, and bintest 105 OK.
- `MRUBY_CONFIG=build_config/asan.rb rake test` (clang, ASAN and UBSan): 2197
  total, 2195 OK, 0 KO, 0 crash, no sanitizer report, and bintest 78 OK.
- A `full-core` build with `MRB_UTF8_STRING`: 2197 total, 2195 OK, 0 KO,
  0 crash, and bintest 78 OK. The sweep above runs there.
- `clang -Wall -Wextra` over `re_compile.c` and `re_utf8.c`: no new warning.

## Summary by CodeRabbit

* **New Features**
  * Added support for Unicode escape sequences in regular expressions, including fixed-width and braced formats.
  * Supports Unicode codepoint lists, character classes, quantifiers, extended mode, and case-insensitive matching.
  * Added UTF-8 encoding for valid Unicode codepoints.

* **Bug Fixes**
  * Invalid, malformed, surrogate, and out-of-range Unicode escapes now produce clear regular expression errors.

* **Documentation**
  * Documented supported character escapes, examples, and hexadecimal escape limitations.

* **Tests**
  * Added comprehensive coverage for Unicode escapes, case folding, and invalid input handling.

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -26,6 +26,21 @@ simulation) with backtracking fallback.
 - `(?<=...)` positive lookbehind (fixed-length only)
 - `(?<!...)` negative lookbehind (fixed-length only)
 
+### Character Escapes
+
+- `\n`, `\t`, `\r`, `\f`, `\v`, `\a`, `\e` control characters
+- `\NNN` octal, one to three digits
+- `\xHH` hex, one or two digits
+- `\uXXXX` Unicode codepoint, exactly four hex digits
+- `\u{...}` Unicode codepoints, one to six hex digits each, several of
+  them separated by spaces: `/\u{61 62}/` is `ab`
+
+Outside a character class the list form is a sequence rather than one
+atom, so a quantifier after it repeats the last codepoint only:
+`/\u{61 62}+/` is `ab+`. Inside a class every codepoint is a member of
+its own, and the last one can still open a range: `/[\u{61 62}-z]/` is
+`a` plus `b-z`.
+
 ### Anchors
 
 - `^` beginning of line
@@ -148,6 +163,8 @@ pattern analysis.
   Maximum 255 bytes.
 - **No Unicode properties**: `\p{Alpha}`, `\p{L}`, etc. are not
   supported.
+- **No `\x{...}` hex escape**: the hex escape is `\xHH`, so it reaches
+  `0xff` at most. Write `\u{...}` for a codepoint above that.
 - **ASCII case folding by default**: The `i` flag handles ASCII letters
   only unless the build defines `MRB_REGEXP_UNICODE_CASE`, which adds the
   Unicode foldings that pair one codepoint with one other. Without the
--- a/mrbgems/mruby-regexp/include/re_internal.h
+++ b/mrbgems/mruby-regexp/include/re_internal.h
@@ -161,6 +161,7 @@ void mrb_re_free(mrb_state *mrb, mrb_regexp_pattern *pat);
 /* UTF-8 helpers */
 int mrb_re_utf8_charlen(const char *s, const char *end);
 uint32_t mrb_re_utf8_decode(const char *s, const char *end, int *len);
+int mrb_re_utf8_encode(uint32_t cp, char *buf);
 mrb_bool mrb_re_is_word_char(uint32_t c);
 
 /* The two foldings whose result is an ASCII letter. Every build carries them,
--- a/mrbgems/mruby-regexp/src/re_compile.c
+++ b/mrbgems/mruby-regexp/src/re_compile.c
@@ -380,6 +380,17 @@ posix_class_bits(uint8_t *bits, const char *name, size_t len)
 #undef BRANGE
 }
 
+/* Value of one hex digit, or -1 for anything else (including the -1 that
+   peek() returns at the end of the pattern). */
+static int
+hex_value(int ch)
+{
+  if (ch >= '0' && ch <= '9') return ch - '0';
+  if (ch >= 'a' && ch <= 'f') return ch - 'a' + 10;
+  if (ch >= 'A' && ch <= 'F') return ch - 'A' + 10;
+  return -1;
+}
+
 static int
 parse_escape(re_compiler *c)
 {
@@ -419,12 +430,8 @@ parse_escape(re_compiler *c)
     int val = 0;
     int n = 0;
     while (n < 2) {
-      int d = peek(c);
-      int v;
-      if (d >= '0' && d <= '9') v = d - '0';
-      else if (d >= 'a' && d <= 'f') v = d - 'a' + 10;
-      else if (d >= 'A' && d <= 'F') v = d - 'A' + 10;
-      else break;
+      int v = hex_value(peek(c));
+      if (v < 0) break;
       val = val * 16 + v;
       next_char(c);
       n++;
@@ -433,6 +440,16 @@ parse_escape(re_compiler *c)
   }
   default: return ch;  /* literal: \., \\, \/, \(, etc. */
   }
+}
+
+/* Add one codepoint to the class: the ASCII bitmap and the codepoint range
+   list each hold one side of 128, and class_match() picks the side to read
+   from the codepoint alone. */
+static void
+class_add_member(re_compiler *c, re_charclass *cc, uint32_t cp)
+{
+  if (cp < 128) class_set_bit(cc, (uint8_t)cp);
+  else class_add_codepoint(c, cc, cp);
 }
 
 /* Read one character class atom: either an ASCII byte (0-127), a
@@ -537,8 +554,7 @@ compile_charclass(re_compiler *c)
       }
     }
     else {
-      if (cp < 128) class_set_bit(cc, (uint8_t)cp);
-      else class_add_codepoint(c, cc, cp);
+      class_add_member(c, cc, cp);
     }
   }
   next_char(c);  /* skip ']' */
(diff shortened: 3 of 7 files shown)


matz:
This needs a rebase: #7058 landed as 3bc4b8884 and it adds `emit_char_folded()` right where this adds `emit_codepoint()`. Three hunks in `re_compile.c` and one in the README.

The textual part is mechanical, both sides being new functions side by side. The part I would rather you decide than guess at is what `\u` should do under `/i`, because the two changes answer it differently today.

On your branch as it stands:

```ruby
Regexp.new("\\u00c0", Regexp::IGNORECASE) =~ "à"   # nil
```

On master, the same character written literally:

```ruby
Regexp.new("À", Regexp::IGNORECASE) =~ "à"         # RegexpError
```

A naive merge keeps both, and then one character answers "silently no match" when it is named by codepoint and "refused" when it is spelled, in the same build. That is the split #7058 exists to remove, so whichever way it goes it should go the same way for both spellings.

I have no view on where `\u` should hook in, only that the two should agree. Presumably the escape wants to reach `emit_char_folded()` the way a literal does, so that a codepoint the build cannot fold is refused whichever way it was written, and one it can fold is folded. Your call, and it may be simpler than that once the two are in the same file.

The rest I checked before running into this. All seven rows in your table reproduce on master, `/\u{3042}/`, `/\u{61 62}/` and `/[µ]/` all answer correctly on your branch, and `/A/i` matching "a" shows the ASCII side is unaffected. It is only the non-ASCII `/i` case that the rebase has to settle.

takumin:
Rebased, onto `beefa1e02` rather than `3bc4b8884` since #7075 and #7076 landed in the meantime. Neither touches `re_compile.c`, so the only overlap with them was one README section and one test file, both automatic.

On the question you left to me: the escape reaches the folding the way a literal does, as you guessed. `emit_char_folded()` had the decode and the folding in one function, and only the decode wants the pattern, so the folding half is now `emit_cp_folded()` and both callers reach it: `emit_char_folded()` after decoding what the pattern spells, and `emit_codepoint()` with the codepoint the escape named. No logic moved with it.

Both spellings answer alike now, and both spellings of the escape do too:

```ruby
# Build without MRB_REGEXP_UNICODE_CASE
Regexp.new("À",         Regexp::IGNORECASE)  # RegexpError
Regexp.new("\\u00c0",   Regexp::IGNORECASE)  # RegexpError
Regexp.new("\\u{c0}",   Regexp::IGNORECASE)  # RegexpError
Regexp.new("[\\u{c0}]", Regexp::IGNORECASE)  # RegexpError

# Build with MRB_REGEXP_UNICODE_CASE
Regexp.new("À",       Regexp::IGNORECASE) =~ "à"  # 0
Regexp.new("\\u00c0", Regexp::IGNORECASE) =~ "à"  # 0
```

A character class was already consistent before this, since a class member is a codepoint whichever way it was written and the closure runs over the finished class. The literal path is what needed the change.

The tests for it sit with the rest of the case folding tests rather than with the `\u` ones, since what they assert depends on the build: `test/ascii_case.rb` takes the refusals and the U+212A case that folds to ASCII without the table, `test/unicode_case.rb` takes the foldings. Each covers the plain form, the class, the negated class and a range.

Checked both builds, and checked that the tests fail without the change: without it the option build answers `nil` where it now matches, so the two paths were not agreeing by accident.

takumin:
The CodeRabbit nitpick on `test/ascii_case.rb` has no inline thread to answer
in, so: fixed in 0b665e3ec.

The point was fair. The `\u` cases asserted `RegexpError` and nothing else, and
a `\u` pattern raises that class for its own spelling too, so a regression in
the escape parser could report a malformed escape for `\u{100}` and the test
would still pass. They now assert the message through
`assert_raise_with_message`, whose tail (`for this character` against `for this
character class`) also pins which of the two paths refused.

The `begin`/`rescue` that checked the literal path's message went the same way,
since it asserted nothing at all when the pattern compiled.

`rake test`: 2037 total, 2019 OK, 0 KO, 0 crash, and bintest 105 OK. Breaking
the expected message on purpose turns the test red, so the assertions do run.

PR #7086, opened 2026-08-10
style/performance fixes

leviongit:
the code under `$repo/mrblib/` has been changed to use `yield` instead of `&.call`
due to the `OP_BLKCALL` optimization

similarly, checks of the form `var.{is_a?, kind_of?}(Class)` have been changed into `Class === var` (aside from test, and prism source)

## Summary by CodeRabbit

- **Refactor**
  - Standardized internal type validation across array, numeric, range, string, symbol, I/O, socket, and collection operations.
  - Streamlined block execution across enumerable, hash, numeric, range, array, and string methods.
  - Preserved existing method signatures, iteration behavior, return values, formatting, comparisons, and error handling.

- **Bug Fixes**
  - Improved consistency of type handling and callback execution without changing expected application behavior.

### DIFF
--- a/mrbgems/mruby-array-ext/mrblib/array.rb
+++ b/mrbgems/mruby-array-ext/mrblib/array.rb
@@ -498,7 +498,7 @@ class Array
 
     column_count = nil
     self.each do |row|
-      raise TypeError unless row.is_a?(Array)
+      raise TypeError unless Array === row
       column_count ||= row.size
       raise IndexError, 'element size differs' unless column_count == row.size
     end
--- a/mrbgems/mruby-compar-ext/mrblib/compar.rb
+++ b/mrbgems/mruby-compar-ext/mrblib/compar.rb
@@ -43,7 +43,7 @@ module Comparable
   def clamp(min, max=nil)
 
     if max.nil?
-      if min.kind_of?(Range)
+      if Range === min
         max = min.end
         if !max.nil? && min.exclude_end?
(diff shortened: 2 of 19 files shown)


matz:
Merged, thank you. Both halves are changes I want, and the second one is a convention I wrote into `CONTRIBUTING.md` a few days ago that nothing had gone back and applied.

What I checked before merging, since a sweep across nineteen files is the kind where a single bad substitution hides easily:

- The 39 block changes are all `block.call(...)`, none of them `&.call`, so nothing that was guarding against a missing block lost its guard. `map`, `select`, `each_with_index`, `collect!` and friends still return an Enumerator when called without one, and `max`/`min` still fall back to `<=>`.
- The 28 type checks all kept the class on the left and the value on the right. `Array === row`, not the reverse.
- Twenty rows against CRuby 4.0.6, block-present and block-absent, all agreeing. Full suite green, ASan and UBSan clean, no new warning.

One request for next time, and it is about the shape rather than the content: **please split a sweep like this into one pull request per kind of change.** `block.call` to `yield` and `is_a?` to `Module#===` are independent, and each would stand alone.

The reason is collision surface rather than review effort. This landed alongside a run of unrelated work in `mrblib` and `mrbgems/*/mrblib`, and a nineteen-file diff has to be rebased whole every time any of it moves. Two smaller ones would each have moved through on their own. It also lets one be reverted without the other if something turns up later.

Nothing to change here, and I would rather have the sweep than not have it. Thank you for doing the unglamorous half of a convention.

PR #7109, opened 2026-08-12
Share UTF-8 byte lengths and character heads between core and mruby-regexp

takumin:
mruby-regexp carries its own answers to two questions core already answers:
how many bytes the character at a pointer takes, and where the character
covering a pointer starts. `mrb_re_utf8_charlen()` and core's `mrb_utf8len()`
agree byte for byte, and `mrb_re_utf8_interior_p()` applies the same rule as the
`str_char_head()` that #7107 arrived at.

They are separate because core keeps `mrb_utf8len()` behind `MRB_UTF8_STRING`,
and the regexp engine has no such guard: it reads UTF-8 whatever a build's
strings index by. That is not a hypothetical configuration. mruby-regexp is in
the stdlib gembox and mruby-encoding is not, so the stock **default gembox
already builds the engine without `MRB_UTF8_STRING`**, decoding UTF-8 with its
own copy while core's sits compiled out.

### What moves

`MRB_UTF8_STRING` says a string indexes by character. What a run of bytes spells
is a question apart from that, so the two functions that only answer it come out
from under the guard:

- `mrb_utf8len()`, unchanged
- `mrb_utf8_char_head()`, which is #7107's `str_char_head()` renamed and
  declared in internal.h beside it

Nothing else follows. `mrb_utf8_strlen()`, the character and byte conversions
and the searches stay behind the guard, because those are the indexing.

### What the engine drops

- `mrb_re_utf8_charlen()` is deleted; `mrb_re_charlen()` and
  `emit_char_bytes()` call `mrb_utf8len()`
- `mrb_re_utf8_interior_p()` becomes `mrb_utf8_char_head(...) != s`
- `mrb_re_utf8_decode()` stays, since core offers no decoder, but takes its
  byte length from `mrb_utf8len()` too

The predicate keeps its own answer for a byte that starts a character, so most
of the positions the matcher tries stay a branch rather than a call. It also
stops reading `*s` at the end of the subject, where it used to lean on the
terminating NUL to answer the same way.

### Equivalence

Checked by brute force rather than argued:

- byte length, `mrb_re_utf8_charlen()` against `mrb_utf8len()`: 2,359,296 cases,
  both first bytes exhaustively, by nine representative tail bytes, by every
  available length from 1 to 4
- the predicate, before against after: 33,338,661 positions, every buffer of up
  to five bytes over a 23-value alphabet covering each lead, continuation and
  discriminant class, at every position

No differences in either. The same run checks each head against a forward walk
from the start of the buffer, which is what makes the three-byte lookback exact
rather than an approximation of one: if a character covers `p` and starts `back`
bytes earlier, every byte in between is a continuation byte, so the nearest
non-continuation byte going back is that character's lead, and that lead cannot
belong to an earlier character in turn.

### Size

Deleting one copy costs less than exposing the other, in both directions:

| build | before | after |
| --- | ---: | ---: |
| full-core | 1808085 | 1806989 |
| default gembox (no `MRB_UTF8_STRING`) | 1722079 | 1721479 |

`size(1)` text of `bin/mruby`, gcc `-O2` on x86-64.

### Base

Branched off #7107, which is still open, because `str_char_head()` arrives
there. The first two commits here are that PR and drop out once it merges.

### Verified

- `rake test` on a full-core build: 2248 tests, all green
- `rake test` on the default gembox, which is the build where the engine reads
  UTF-8 without `MRB_UTF8_STRING`: 2057 tests, all green
- each commit is green on its own

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTF-8 character boundary handling for string indexing and reverse searches.
  * Corrected `String#rindex` behavior with negative positions, multibyte characters, invalid UTF-8 bytes, and boundary limits.
  * Improved regular-expression processing of valid and invalid UTF-8 sequences.
* **Tests**
  * Added coverage for UTF-8-aware `String#rindex` behavior and edge cases.

### DIFF
--- a/include/mruby/internal.h
+++ b/include/mruby/internal.h
@@ -190,8 +190,18 @@ mrb_int mrb_str_byte_to_char(mrb_state *mrb, mrb_value str, mrb_int bi);
 mrb_bool mrb_str_valid_encoding_p(mrb_state *mrb, mrb_value str);
 
 mrb_int mrb_utf8_to_buf(char *buf, uint32_t cp);
-#ifdef MRB_UTF8_STRING
+
+/* The byte length of the character at `str`, which has to be a byte of the
+   string rather than `end` itself, and 1 for a run of bytes that spells no
+   character. See the definition in string.c for what it rejects. */
 mrb_int mrb_utf8len(const char *str, const char *end);
+
+/* The byte the character covering `p` starts at, or `p` itself when `p` is
+   already a character boundary. A continuation byte that no lead byte reaches
+   is a boundary too. */
+const char *mrb_utf8_char_head(const char *beg, const char *p, const char *end);
+
+#ifdef MRB_UTF8_STRING
 mrb_int mrb_utf8_strlen(const char *str, mrb_int byte_len);
 #endif
 
--- a/mrbgems/mruby-regexp/include/re_internal.h
+++ b/mrbgems/mruby-regexp/include/re_internal.h
@@ -8,6 +8,7 @@
 #define MRB_RE_INTERNAL_H
 
 #include <mruby.h>
+#include <mruby/internal.h>
 #include <stdint.h>
 
 /* Bytecode instructions for the NFA engine */
@@ -173,7 +174,6 @@ mrb_regexp_pattern* mrb_re_compile(mrb_state *mrb, const char *pattern, mrb_int 
 void mrb_re_free(mrb_state *mrb, mrb_regexp_pattern *pat);
 
 /* UTF-8 helpers */
-int mrb_re_utf8_charlen(const char *s, const char *end);
 uint32_t mrb_re_utf8_decode(const char *s, const char *end, int *len);
 mrb_bool mrb_re_is_word_char(uint32_t c);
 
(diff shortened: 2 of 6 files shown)


takumin:
@coderabbitai Both applied in e33179b, which is also rebased onto master now that #7107 has merged.

**`mrb_utf8len`'s precondition.** Right that the comment left it unsaid. It now names the byte the function reads:

> `mrb_utf8len` returns the byte length of the character at `str`, which has to be a byte of the string rather than `end` itself, and 1 for anything that spells no character: [...]

Every caller in the tree already holds to it: the walks test `p < e` before they step, and `io.c` returns early on an empty buffer.

**The truncated sequence.** Added, though not with the expected values the diff proposed. `"a\xE3\x81".length` is 3, not 2: `\xE3` claims three bytes and the string end leaves it two, so `mrb_utf8len` answers 1 for it, and `\x81` is then a continuation byte that no lead byte reaches, which is 1 as well.

```ruby
  # A lead byte the string end cuts short reaches none of the bytes that
  # follow it, so those stand alone too.
  assert_equal 3, "a\xE3\x81".length
  assert_equal 1, "a\xE3\x81".rindex("\xE3")
  assert_equal 2, "a\xE3\x81".rindex("\x81")
```

It is worth having, since it is the case that pins `end` down to the end of the subject: measure that lead byte against a longer buffer and the length would come out 2.

Green after the change on both builds, 2248 tests on full-core and 2057 on the default gembox, and the first commit is green on its own.

PR #7110, opened 2026-08-12
mruby-regexp: refuse a subject whose bytes are not UTF-8

takumin:
CRuby raises `ArgumentError` when a search is given a subject holding a byte
that spells no character. mruby answered for it, so the same program took a
result CRuby would not have produced:

```ruby
```

Refuse the search instead, for `=~`, `match`, `match?`, `===`, `index`,
`rindex`, `byteindex`, `byterindex`, `[]`, `[]=`, `sub`, `gsub`, `scan`,
`split`, `partition`, `rpartition`, `start_with?` and `slice!`. `String#scrub` is how a
subject like this becomes matchable.

### Two commits

The first moves the tests that ask what the engine does with a byte standing
for no character onto a byte-indexed subject. What each of them pins belongs to
the pattern, and a byte-indexed subject puts the same question to the engine,
so they keep asking it after the second commit refuses the UTF-8 subject they
used to go through. That commit changes no behavior and is green on its own.

The second is the refusal itself, which leaves it holding the check, the tests
whose answer it changes, and nothing else.

### What goes through untouched

A binary string is indexed by byte from end to end, so its bytes make no claim
that could be broken.

A quoted String pattern is exempt on the grounds CRuby exempts it: a literal is
searched for byte by byte, and the subject is read as UTF-8 nowhere along the
way.

```ruby
"あ\x80b".sub("b", "!")  #=> "あ\x80!" in both, where /b/ is refused in both
"あ\x80b".scan("b")      # ArgumentError in both: `scan` refuses a literal too
```

`sub`, `sub!`, `gsub` and `gsub!` quote a String pattern into a `Regexp` before
they search, so each carries the fact that it was a literal down to the search
rather than letting the quoting hide it.

A string derived from a binary one does not carry `MRB_STR_BINARY` with it, so
`b + ""` and `b.sub(...)` are refused where `b` itself is not. That gap predates
this change and is left as it is.

### Where the check runs

`re_check_encoding()` walks the whole subject, so it runs once per search a
method makes rather than once per match it finds.

The entry points that take a subject from Ruby check it themselves:
`Regexp#match`, `#=~`, `#===`, `#match?`, and the class methods `__search`,
`__search_p`, `__sub_str`, `__gsub_str` and `__scan`, the last three of which
loop in C over a subject already checked.

The two entry points an mrblib loop drives leave the check to that loop, which
runs it once on the subject it holds fixed. `Regexp.__check_encoding` at the top
of `gsub` with a block, of `split` and of `byteindex` covers their
`__byte_search` calls, and `__regexp_rsearch`, which backs `rindex`,
`byterindex` and `rpartition`, runs it and passes the flag that says so to
`__search`.

Checking inside those loops instead walks the subject once per match, which is
quadratic in the number of matches: `("あ," * 20000).split(/,/)` took 8 times as
long that way.

### Cost

What remains is one walk per search, which a whole subject pays as well. The
flag `String#length` leaves behind cannot stand in for the walk, since a string
of stray bytes has one byte per character too, so a subject searched twice is
walked twice.

The walk skips a run of ASCII a word at a time and decodes only where a byte
leaves that range, so the cost follows how much of the subject is multi-byte.
Measured at `-O3` on a full-core build, three runs each:

| | before | after |
|---|---|---|
| `("あ" * 100000 + "z") =~ /z/`, 200 times | 62 ms | 117 ms |
| the same over an ASCII subject of the same size | 11.8 ms | 12.0 ms |
| `"hello world" =~ /world/`, 100000 times | 139 ms | 140 ms |

### One known difference

An out of range positive `pos` still answers nil rather than raising, since
`re_char_to_byte()` rejects it before the check runs. CRuby raises there for
`Regexp#match`; for `index`, `rindex` and `byteindex` it answers nil as this
does.

### Tests

What a match position inside a whole character is remains pinned on subjects
that are whole UTF-8. A build without `MRB_UTF8_STRING` reads no encoding for
the bytes to break, so the cases that turn on one answer positions there and
branch on `__ENCODING__`.

The engine reads a byte-indexed subject through a branch of its own, so the
tests the first commit moved stop covering the UTF-8 one. `Regexp - truncated
UTF-8 at subject end` is the case where that mattered: it was a fuzz-derived
regression test for a read past the end of the string buffer, and the code path
it covered is no longer reachable from Ruby under this rule.

### Verified

  2246 tests at the first commit and 2247 at the second, all green
- `rake test` on the default gembox (no `MRB_UTF8_STRING`): 2057 and 2058 tests,
  all green
- 30 String-argument methods compared against CRuby 4.0.6 on a broken subject.
  The two remaining differences are `split("b")`, which CRuby refuses and the
  aliased C `__split` still answers, and `slice!("b")`, which is wrong on whole
  UTF-8 too and is not this change

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved UTF-8 validation across regular expression search, matching, splitting, scanning, and substitution operations.
  - Invalid UTF-8 subjects now raise `ArgumentError` consistently in UTF-8 builds.
  - Preserved byte-oriented behavior for literal string operations and binary data.
  - Improved handling of malformed, truncated, and overlong UTF-8 sequences, including character-class matching.

### DIFF
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -115,12 +115,17 @@ class String
     # get_pat.  Only the quoting is taken from it: get_pat_quoted also accepts
     # anything answering `to_str`, where `__check_pattern` keeps to a real
     # String, as `match` already does.
-    pattern = Regexp.new(Regexp.escape(pattern)) if String === pattern
+    literal = String === pattern
+    # CRuby searches for a literal byte by byte and never reads the subject as
+    # UTF-8 on the way, so quoting one into a Regexp here must not put the
+    # subject through a check CRuby does not make: `"a\x80b".sub("b", "!")`
+    # answers there, where the same call with `/b/` is refused.
+    pattern = Regexp.new(Regexp.escape(pattern)) if literal
     # A replacement argument wins over the block, as in CRuby.
     if args.length == 2
-      return Regexp.__sub_str(pattern, self, replacement.to_s)
+      return Regexp.__sub_str(pattern, self, replacement.to_s, literal)
     end
-    md = Regexp.__search(pattern, self)
+    md = Regexp.__search(pattern, self, 0, literal)
     return self.dup unless md
     md.pre_match + block.call(md[0]).to_s + md.post_match
   end
@@ -141,19 +146,23 @@ class String
     # Resolved here rather than left to `sub` because the match below decides
     # the return value, and a String pattern is a literal on both paths.
     pattern = Regexp.__check_pattern(args[0])
-    pattern = Regexp.new(Regexp.escape(pattern)) if String === pattern
+    literal = String === pattern
+    pattern = Regexp.new(Regexp.escape(pattern)) if literal
     raise FrozenError, "can't modify frozen String" if frozen?
     # Whether a substitution happened is a question about the match, not about
     # the result: `"aaa".sub!(/a/, "a")` returns self even though the string is
     # unchanged.  A full search and not `match?`, so a failed match clears $~.
-    return nil unless Regexp.__search(pattern, self)
+    return nil unless Regexp.__search(pattern, self, 0, literal)
     # `sub` matches again and publishes its own $~ over this one, leaving the
     # caller the match `sub` would have left, a block's own matches included.
     # The resolved pattern takes the place of the original argument so that a
-    # String is not quoted and compiled a second time.  Overwriting `self`
-    # afterwards is safe: a MatchData snapshots its subject, so $~ keeps
-    # describing the string as it was matched.
-    str = args.length == 2 ? self.sub(pattern, args[1], &block) : self.sub(pattern, &block)
+    # String is not quoted and compiled a second time; a literal goes down as
+    # the String it was instead, since that is what tells `sub` to leave the
+    # subject unread, and quoting it twice is the price of saying so.
+    # Overwriting `self` afterwards is safe: a MatchData snapshots its subject,
+    # so $~ keeps describing the string as it was matched.
+    down = literal ? args[0] : pattern
+    str = args.length == 2 ? self.sub(down, args[1], &block) : self.sub(down, &block)
     self.replace(str)
   end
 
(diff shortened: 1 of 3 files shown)


takumin:
@coderabbitai Added, thanks. `String#[]=` does reach `Regexp.__search`, and it
refuses a broken subject the way CRuby does:

```ruby
s = "あ\x80b"
s[/b/] = "!"   # ArgumentError: invalid byte sequence in UTF-8, in both
```

The assertion now sits next to the `slice!` one in the second commit, and
`[]=` is named alongside the other methods in the commit message and in the
description, which had left it out.

takumin:
Closing this in favour of four smaller pull requests. As it stood, a reviewer
had to read a behaviour change and a rewrite of most of `regexp_utf8.rb` at the
same time, and the two argue for each other in a way that is hard to check in
one pass.

Split as follows, each green on its own:

1. #7115 asks the engine's byte questions of a byte-indexed subject too. Test
   only, purely additive, no behaviour change. Independent of the rest.
2. #7116 does the same for the malformed sequence cases (overlong, surrogate,
   above U+10FFFF, truncated at the subject end, truncated in a class). Test
   only, purely additive, independent of #7115.
3. The refusal itself, wherever the search runs in C: `=~`, `match`, `match?`,
   `===`, `index`, `rindex`, `byterindex`, `[]`, `[]=`, `sub`, `sub!`, `gsub`
   with a replacement, `gsub!`, `scan`, `partition`, `rpartition`,
   `start_with?` and `slice!`, with the quoted String pattern exempt as CRuby
   exempts it. With 1 and 2 in, the tests whose subject this refuses keep their
   question and lose only the subject that no longer reaches the engine, so this
   piece is the check, the tests it changes, and nothing else.
4. `Regexp.__check_encoding` for the three entry points an mrblib loop drives
   (`gsub` with a block, `split`, `byteindex`), plus hoisting the check out of
   the backward walk that backs `rindex`, `byterindex` and `rpartition`. One
   walk per search rather than one per match: measured at 105 ms against
   1404 ms for `("あ," * 20000).split(/,/)`.

3 and 4 touch the same lines of `regexp_utf8.rb` as 1 and 2, so I will open them
once those two are settled rather than post a diff that carries their content
along. The behaviour, the exemptions and the cost are the same as described
here; nothing has been dropped from the plan.

takumin:
#7115 and #7116 are in, so the third piece is #7119. The fourth follows once that one is settled, since the two touch the same lines of `regexp_utf8.rb`.

PR #7120, opened 2026-08-12
mruby-regexp: ask the byte cases of a byte-indexed subject alone

takumin:
Third of the pieces #7110 was split into, on top of #7115 and #7116, and test
only. Deleting assertions deserves its own pull request rather than a paragraph
inside one that changes behaviour, so here it is on its own.

Each of these cases asks what the engine does with a byte that stands for no
character, and since #7115 and #7116 it asks twice: once through a subject read
as UTF-8 that carries the byte, and once through a byte-indexed subject.

### Only one of the two subjects can answer

Read as UTF-8, the subject spells no character where that byte is. What a
quantifier binds to there, whether a class holds the byte, and where a match may
start are then answers about a decoder walking input no rule covers.

Byte-indexed, the same byte is a byte from end to end, every position reported is
a byte offset the subject's own indexing agrees with, and the engine reaches it
through the branch it keeps for exactly that: `mrb_re_charlen()` and
`mrb_re_decode_char()` advance a byte at a time, and the seeding guard skips
`mrb_re_utf8_interior_p()`, so no position is the interior of anything.

### What this does

Drops the UTF-8 subject from those cases and keeps the byte-indexed one. The
comment that explained each case moves to the assertion that still makes it,
which is most of what the diff shows as added.

Nothing a whole UTF-8 subject can ask is touched:

- a match position inside a character (`"あ".match(Regexp.new("\x81"))` and the
  `"ĵ"` cases) keeps its subject
- a class that holds a character rather than a byte (`mu =~ /[\xB5]/` and the
  `\u{B5}` spellings) keeps its subject
- the pattern side of an overlong sequence
  (`Regexp.new("[\xC0\xBC]").match?("<")`) keeps its subject
- `Regexp - match positions on malformed UTF-8 agree with string indexing` is
  left exactly as it is: what it pins is what mruby knows about the string
  rather than what the engine does with the bytes, so it belongs with the piece
  that changes its answer

The fuzz-derived buffer cases (`Regexp - invalid UTF-8 byte near pattern end`
and `Regexp - truncated UTF-8 at subject end`) keep a walk to the end of each
buffer, in the byte-indexed form #7116 added.

### Verified

  2253 tests, all green
- `rake test` on the default gembox (no `MRB_UTF8_STRING`): 2060 tests, all
  green

## Summary by CodeRabbit

* **Tests**
  * Expanded UTF-8 regular expression coverage for malformed, truncated, overlong, and invalid sequences.
  * Added validation for byte-based match positions, extraction offsets, character classes, and quantifier boundaries.
  * Improved coverage of decoding boundaries in both patterns and input text.

### DIFF
--- a/mrbgems/mruby-regexp/test/regexp_utf8.rb
+++ b/mrbgems/mruby-regexp/test/regexp_utf8.rb
@@ -56,16 +56,16 @@ assert("Regexp - /i does not read a byte above 127 as a character") do
   # path would take a lone 0xB5 for U+00B5 and answer /i for a character the
   # pattern does not hold. A literal compares bytes, with or without /i.
   micro = "\xB5"        # U+00B5 is "\xC2\xB5"; the byte on its own is not it
-  assert_equal 0, (Regexp.new(micro, Regexp::IGNORECASE) =~ micro)
   assert_nil (Regexp.new(micro, Regexp::IGNORECASE) =~ "\u00B5")
   # A sequence cut short by the end of the pattern reads the same way.
   lead = "\xC3"         # starts a two byte character and never completes one
-  assert_equal 0, (Regexp.new(lead, Regexp::IGNORECASE) =~ lead)
   assert_nil (Regexp.new(lead, Regexp::IGNORECASE) =~ "\u00E3")
   # What /i folds is settled when the pattern is compiled, so a byte-indexed
   # subject puts the same question to the engine, which reads one through a
   # branch of its own: `mrb_re_exec` takes the flag and every step it drives,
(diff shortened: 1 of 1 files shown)


takumin:
Closing in favour of #7121. This diff showed 21 added assertion lines, which are the same assertions moved rather than new ones, and a diff that deletes tests should not need the reader to work that out. #7121 does the same deletion with no reordering, so its only added lines are comments (`+39/-79`, every added line beginning with `#`), and it says in each block, and in a table in the description, which assertions go and why.

PR #7126, opened 2026-08-12
mruby-regexp: refuse a subject whose bytes are not UTF-8

takumin:
CRuby raises `ArgumentError` when a search is given a subject holding a byte that spells no character. mruby answers for it instead, so the same program takes a result CRuby would not have produced:

```ruby
```

This follows CRuby wherever a search reads the subject: `=~`, `match`, `match?`, `===`, `index`, `rindex`, `byteindex`, `byterindex`, `[]`, `[]=`, `sub`, `sub!`, `gsub`, `gsub!`, `scan`, `split`, `partition`, `rpartition`, `start_with?` and `slice!`. `String#scrub` is how a subject like this becomes matchable.

A binary string goes through untouched, since it is indexed by byte from end to end and its bytes make no claim that could be broken.

### What this leaves for the next PR

A quoted String pattern is refused here as well, where CRuby answers for it:

```ruby
"あ\x80b".sub("b", "!")  # CRuby: "あ\x80!", here: ArgumentError
```

CRuby searches for a literal byte by byte and reads the subject as UTF-8 nowhere along the way. Exempting it means carrying the fact that the pattern was a literal from `sub` down to the search it makes, which is a change of its own; it is the PR that follows this one.

### Where the check goes, and what it costs

The check goes in each search rather than at the entry to the methods that drive one, because core remembers a string it has read as valid UTF-8. Most entry points search once for what arrives from Ruby, but `__byte_search`, which the mrblib loops of `gsub`, `split` and `byteindex` drive, is called once per match, and `__regexp_rsearch` steps `__search` over every match position. Those turns read the flag core left on the subject after the first one, so what they cost is a flag test and not a walk. A `gsub` block that modifies the subject clears the flag and the next turn walks what is left; mruby allows such a block where CRuby raises `RuntimeError: string modified`, and this leaves that as it is.

The walk skips a run of ASCII a word at a time and decodes only where a byte leaves that range, so what a subject nothing has walked yet pays follows how much of it is multi-byte. At `-O3` on a full-core build, the fastest of fifteen runs each, with `mb` for `"あ" * 100000` and `cs` for `"あ," * 20000`:

| | before | after |
|---|---|---|
| `(mb + "z") =~ /z/`, 200 times, less the cost of building the subjects | 52.1 ms | 103.8 ms |
| the same with an ASCII subject of the same byte size | 1.2 ms | 0.7 ms |
| `"hello world" =~ /world/`, 100000 times | 133.6 ms | 133.0 ms |
| `cs.split(/,/)`, one walk and 20000 searches | 71.1 ms | 61.4 ms |
| `cs.gsub(/,/) { ";" }`, the same | 75.1 ms | 72.9 ms |
| `cs.rindex(/,/)`, one walk and 20000 match positions | 2105.9 ms | 2114.6 ms |

The first row builds a subject for every search so that every search walks one. The rest reuse a subject, which is why the loop rows show what the flag costs rather than what the walk does, and they and the ASCII row are inside the noise of the machine they were taken on.

### What changed since #7110

#7110 was all of this in one PR. Its test moves are in master already (#7115, #7116, #7120, #7121), so what is left here is the refusal itself, and the quoted String pattern it exempted is the PR that follows this one. Three things it said do not hold, which is why the shape is different:

- It ran the check at the entry to each mrblib loop, through a new `Regexp.__check_encoding` class method and a flag passed into `__search` out of `__regexp_rsearch`, on the grounds that checking inside `__byte_search` walks the subject once per match, quadratic in the number of matches, and that `("あ," * 20000).split(/,/)` took 8 times as long that way. Measured again on master it does not: the flag core sets makes every turn after the first a flag test, and neither placement is distinguishable from the other or from master. The class method and the hoist are gone with it, and `__byte_search` holds the check.
- It said that a subject searched twice is walked twice, so a walk fell on every search. The flag survives the search and a copy of the string, so the walk falls on the string.
- It said an out of range `pos` answers nil for `index`, `rindex` and `byteindex`. `rindex` clamps a position past the end to the end rather than rejecting it, so it reaches the check and refuses, which is what CRuby answers too.

### Positions

A `pos` outside the subject still answers nil rather than raising, since `re_char_to_byte()` rejects it before the check runs. `String#index` answers nil for one in CRuby too; `Regexp#match` raises there, where this answers nil. `rindex` clamps a position past the end to the end rather than rejecting it, so it reaches the check and refuses, which is what CRuby answers as well.

### What the lookbehind tests ask now

#7125 measures a lookbehind in the characters its bytes spell, and its tests put the question to a subject whose bytes spell no character, read as UTF-8:

```ruby
"\x80ab" =~ /(?<=\x80a)b/          # 2
"\xE3\x81ab" =~ /(?<=\xE3\x81a)b/  # 3
```

A search like that is refused here, so the block branches on `__ENCODING__` and keeps both answers: a UTF-8 build asserts the refusal, and a build that reads no encoding for the bytes to break asserts the widths. The byte rewind the same block asserts is reached in both, since a byte-indexed subject is exempt.

That leaves the character measurement reachable only where `MRB_UTF8_STRING` is off. A subject that spells characters throughout never matches a lookbehind holding a byte that spells none, at either width, so no other subject puts the question.

### Checked

- `rake test` on the default gembox: 2064 assertions, 0 failures; bintest 105, 0 failures
- `rake test` on a full-core build (`MRB_UTF8_STRING`): 2259 assertions, 0 failures
- `rake test` on a build without `MRB_UTF8_STRING`, holding mruby-regexp and mruby-string-ext: 1060 assertions, 0 failures
- 20 calls over a broken subject compared against CRuby 4.0.6, and the position cases above

## Summary by CodeRabbit

* **Bug Fixes**
  * Regular expression operations now reject malformed UTF-8 subjects with `ArgumentError`.
  * Validation consistently applies to matching, searching, replacements, scanning, splitting, and related operations.
  * Binary strings remain supported without UTF-8 validation.
  * Valid UTF-8 behavior and position handling remain unchanged.

* **Tests**
  * Added coverage for malformed UTF-8 across regular expression and string APIs, including lookbehind and mutation operations.

### DIFF
--- a/mrbgems/mruby-regexp/src/regexp.c
+++ b/mrbgems/mruby-regexp/src/regexp.c
@@ -228,6 +228,28 @@ re_binary_string_p(mrb_value str)
   return RSTR_BINARY_P(RSTRING(str));
 }
 
+/* CRuby refuses a search whose subject holds a byte that spells no character,
+   and mruby answers for it. Refuse it here too, so that a program moved from
+   one to the other is told about the subject rather than handed a result the
+   other would not have produced.
+
+   A binary string is exempt because it is indexed by byte throughout, so its
+   bytes make no claim that could be broken.
+
+   The check walks the whole subject, so every entry point below runs it on the
+   subject it is handed and the C loops over a subject run it before the first
+   turn. Two searches are driven from mrblib once per match rather than once per
+   call, `__byte_search` and the `__search` the backward search steps, and they
+   check too: core remembers a string it has read as valid UTF-8, so every turn
+   after the first costs a flag test and not a walk. */
+static void
+re_check_encoding(mrb_state *mrb, mrb_value str)
+{
+  if (!mrb_str_valid_encoding_p(mrb, str)) {
+    mrb_raise(mrb, E_ARGUMENT_ERROR, "invalid byte sequence in UTF-8");
+  }
+}
+
 static mrb_value
 regexp_binary_string_p(mrb_state *mrb, mrb_value self)
 {
@@ -362,6 +384,7 @@ regexp_match(mrb_state *mrb, mrb_value self)
     return mrb_nil_value();
   }
 
+  re_check_encoding(mrb, str);
   md = exec_match(mrb, self, str, pos);
   if (!mrb_nil_p(md) && !mrb_nil_p(block)) {
     return mrb_yield(mrb, block, md);
@@ -410,6 +433,7 @@ regexp_s_search(mrb_state *mrb, mrb_value klass)
     clear_match_globals(mrb);
     return mrb_nil_value();
   }
+  re_check_encoding(mrb, str);
   return exec_match(mrb, re, str, pos);
 }
 
@@ -419,7 +443,8 @@ regexp_s_search(mrb_state *mrb, mrb_value klass)
  * Internal: the byte-offset search the mrblib loops of `gsub`, `split` and
  * `byteindex` drive themselves. No position normalization, because the
  * callers already work in byte space, and no operand conversion, because
- * they always pass a String.
+ * they always pass a String. The subject is the one the loop holds fixed, so
+ * the check reads the flag core left on it after the first turn.
  */
 static mrb_value
 regexp_s_byte_search(mrb_state *mrb, mrb_value klass)
@@ -429,6 +454,7 @@ regexp_s_byte_search(mrb_state *mrb, mrb_value klass)
 
   mrb_get_args(mrb, "oS|i", &re, &str, &pos);
   check_regexp_arg(mrb, re);
+  re_check_encoding(mrb, str);
   return exec_match(mrb, re, str, pos);
 }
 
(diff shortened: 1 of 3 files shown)


matz:
Merged, and thank you for the split. The commits are in master as 80c781b6c and 456a708bd; I took them locally rather than through the button, so these stay open and I am closing them by hand.

Two things happened on the way in that are worth recording.

**The lookbehind assertions.** #7125 landed before these, and its subjects are the ones this refuses, so the block crashed once both were in. I resolved it by dropping those assertions to the byte-indexed forms. You had solved the same problem better in the rebase: asserting the refusal on a build that reads an encoding, and keeping the widths on the one that does not. I have taken your version (c2b7b3a27). Mine recorded that the question could no longer be asked; yours asks the new one.

**A mistake of mine, since it touched your work.** While measuring whether the lookbehind fix was still reachable after the refusal, I reverted `re_compile.c` with `git checkout <commit> -- <path>`, which stages as well as writes the file. Restoring the working tree with `cp` left the old version in the index, and the next commit took it. That put #7125's `compute_fixed_len` change back to counting lead bytes, and it was pushed. Restored in 1e6a2e68a, with the history left alone.

What I checked before merging, on top of the verification in your description:

- 95 subject and method combinations against CRuby 4.0.6: 37 differences on master, 1 with both of these in. The one left is `start_with?` on a binary receiver, where CRuby raises `Encoding::CompatibilityError` and this gem has no encoding to compare.
- The pair has to land together. #7126 alone refuses `"あ\x80b".sub("b", "!")`, which CRuby answers, and #7127 is what exempts the quoted literal.
- No measurable cost. `("あいうえお" * 2000) =~ /お/` twenty thousand times runs at master's speed, because the walk is remembered on the string (5a9f2bd2a) rather than repeated per search.

On the piece still to come: `split`, `byteindex` and `gsub` with a block are the three that still answer, and hoisting the check to the entry of each is the right shape for them. No hurry from me.

PR #7127, opened 2026-08-12
mruby-regexp: exempt a quoted String pattern from the subject check

takumin:
Builds on #7126, which refuses a search whose subject holds a byte that spells no character. That PR refuses one made with a String pattern as well, where CRuby answers: a literal is searched for byte by byte and the subject is read as UTF-8 nowhere along the way.

```ruby
"あ\x80b".sub("b", "!")    # CRuby: "あ\x80!", after #7126: ArgumentError
"あ\x80b".scan("b")        # ArgumentError in both
```

`sub`, `sub!`, `gsub` and `gsub!` quote a String pattern into a `Regexp` before they search, which hides what it was. This takes note of it before the quoting and carries the note down to the search: `Regexp.__search`, `__byte_search`, `__sub_str` and `__gsub_str` each take a `checked` argument saying the caller has settled the encoding question and the search must not ask it again.

`sub!` and `gsub!` search once themselves and then call `sub` and `gsub`, which search again. The resolved pattern used to go down to that second call so that a String was not quoted twice; a literal goes down as the String it was instead, since that is what tells `sub` to leave the subject unread. Quoting it a second time is the price of saying so.

### What stays as it is

`scan` is not exempt: CRuby refuses a literal there, so it keeps the check.

`index`, `partition`, `start_with?`, `slice!` and `[]` search a String pattern in core without reaching a Regexp at all, so they already answer what CRuby answers. So does `split`, which is where the two part company: `"あ\x80b".split("b")` answers here and raises in CRuby. That divergence predates both PRs and is left where it was.

This exemption was part of #7110, which is where it was argued from; it is split out here so that #7126 is the refusal alone. What #7110 carried and this does not is the `Regexp.__check_encoding` class method, which #7126 does without.

The first commit here is #7126, so this shows two until that one is merged.

### Checked

- `rake test` on the default gembox: 2063 assertions, 0 failures; bintest 105, 0 failures
- `rake test` on a full-core build (`MRB_UTF8_STRING`): 2257 assertions, 0 failures
- 14 calls with a String pattern over a broken subject compared against CRuby 4.0.6: every one agrees except `split`, which is the divergence named above

## Summary by CodeRabbit

- **Bug Fixes**
  - Regexp matching and substitution now consistently reject invalid UTF-8 text with `ArgumentError`.
  - Literal string patterns preserve byte-oriented behavior, including for substitution and global substitution.
  - Binary strings continue to support byte-based matching without UTF-8 validation.
  - Matching behavior is now consistent across search, scanning, predicates, and replacement operations.

- **Tests**
  - Expanded coverage for malformed UTF-8, encoding-specific behavior, and literal string patterns.

### DIFF
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -115,12 +115,17 @@ class String
     # get_pat.  Only the quoting is taken from it: get_pat_quoted also accepts
     # anything answering `to_str`, where `__check_pattern` keeps to a real
     # String, as `match` already does.
-    pattern = Regexp.new(Regexp.escape(pattern)) if String === pattern
+    literal = String === pattern
+    # CRuby searches for a literal byte by byte and never reads the subject as
+    # UTF-8 on the way, so quoting one into a Regexp here must not put the
+    # subject through a check CRuby does not make: `"a\x80b".sub("b", "!")`
+    # answers there, where the same call with `/b/` is refused.
+    pattern = Regexp.new(Regexp.escape(pattern)) if literal
     # A replacement argument wins over the block, as in CRuby.
     if args.length == 2
-      return Regexp.__sub_str(pattern, self, replacement.to_s)
+      return Regexp.__sub_str(pattern, self, replacement.to_s, literal)
     end
-    md = Regexp.__search(pattern, self)
+    md = Regexp.__search(pattern, self, 0, literal)
     return self.dup unless md
     md.pre_match + block.call(md[0]).to_s + md.post_match
   end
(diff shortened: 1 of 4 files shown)


matz:
Merged, and thank you for the split. The commits are in master as 80c781b6c and 456a708bd; I took them locally rather than through the button, so these stay open and I am closing them by hand.

Two things happened on the way in that are worth recording.

**The lookbehind assertions.** #7125 landed before these, and its subjects are the ones this refuses, so the block crashed once both were in. I resolved it by dropping those assertions to the byte-indexed forms. You had solved the same problem better in the rebase: asserting the refusal on a build that reads an encoding, and keeping the widths on the one that does not. I have taken your version (c2b7b3a27). Mine recorded that the question could no longer be asked; yours asks the new one.

**A mistake of mine, since it touched your work.** While measuring whether the lookbehind fix was still reachable after the refusal, I reverted `re_compile.c` with `git checkout <commit> -- <path>`, which stages as well as writes the file. Restoring the working tree with `cp` left the old version in the index, and the next commit took it. That put #7125's `compute_fixed_len` change back to counting lead bytes, and it was pushed. Restored in 1e6a2e68a, with the history left alone.

What I checked before merging, on top of the verification in your description:

- 95 subject and method combinations against CRuby 4.0.6: 37 differences on master, 1 with both of these in. The one left is `start_with?` on a binary receiver, where CRuby raises `Encoding::CompatibilityError` and this gem has no encoding to compare.
- The pair has to land together. #7126 alone refuses `"あ\x80b".sub("b", "!")`, which CRuby answers, and #7127 is what exempts the quoted literal.
- No measurable cost. `("あいうえお" * 2000) =~ /お/` twenty thousand times runs at master's speed, because the walk is remembered on the string (5a9f2bd2a) rather than repeated per search.

On the piece still to come: `split`, `byteindex` and `gsub` with a block are the three that still answer, and hoisting the check to the entry of each is the right shape for them. No hurry from me.

takumin:
Thank you for taking the rebase version, and for the restore.

On the collision: #7126 has a section on it (*What the lookbehind tests ask now*) carrying the same resolution you arrived at, but #7127 is the page the two commits were shown on and it says nothing about the lookbehind. Read from the top of the stack there was nothing to find. I will repeat an interaction like that in the description of every PR in a stack rather than only in the one that introduces it.

On the piece to come, I put the three to CRuby 4.0.6 and to a full-core build with `MRB_UTF8_STRING`:

```ruby
"あ\x80b".split("b")        # CRuby: ArgumentError, here: ["あ\x80"]
"あ\x80b".byteindex("b")    # CRuby: 4, here: 4
"あ\x80b".gsub("b") { "!" } # CRuby: "あ\x80!", here: "あ\x80!"
```

Only `split` parts company, as it did in the 14 calls measured for #7127. `byteindex` with a String, and `gsub` with a block over a String pattern, are the literal case that PR exempts, and this gem answers them the way CRuby does, so a check at their entry would carry them away from CRuby rather than toward it. Their Regexp forms already refuse: `byteindex(/b/)` and `gsub(/b/) { }` reach `__byte_search` with no `checked`.

That leaves `split`, which reaches no Regexp when the pattern is a String or nil: `String#split` hands both to the C `__split`, and the awk form goes there too, so all three answer where CRuby raises.

```ruby
"あ\x80b".split      # CRuby: ArgumentError, here: ["あ\x80b"]
"あ\x80b".split(" ") # CRuby: ArgumentError, here: ["あ\x80b"]
```

The check goes at the entry to `__split`, the one place all three pass through. I will send it on its own.

PR #7130, opened 2026-08-12
add example .clang-format

leviongit:
we propose implementing a `.clang-format` standard, for the purpose of unifying the formatting across the whole codebase, the proposed config is not exhaustive, and should be extended

## Summary by CodeRabbit

* **Style**
  * Added comprehensive, consistent formatting rules for C code.
  * Standardized indentation, spacing, braces, alignment, line breaks, comments, includes, macros, namespaces, and whitespace.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,219 @@
+---
+Language:        C
+AlignAfterOpenBracket: true
+AlignArrayOfStructures: Right
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: Consecutive
+AlignConsecutiveDeclarations: None
+AlignConsecutiveMacros: Consecutive
+AlignConsecutiveShortCaseStatements:
+  Enabled: false
+AlignEscapedNewlines: LeftWithLastLine
+AlignOperands:   Align
+AlignTrailingComments:
+  AlignPPAndNotPP: true
+  Kind:            Always
+  OverEmptyLines:  0
+AllowAllArgumentsOnNextLine: true
+AllowShortBlocksOnASingleLine: Always
+AllowShortCaseExpressionOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine:
+  Empty:           true
+  Inline:          true
+  Other:           true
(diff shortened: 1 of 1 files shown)


matz:
Thank you for writing this up, and sorry to turn it down. I am closing it, for a reason about the file rather than about the settings you chose.

Two things weigh against carrying a `.clang-format` at all.

**It would rewrite code away from the style the project documents.** `CONTRIBUTING.md` asks for a newline before `else`, and the config sets `BeforeElse: false`, which attaches it. That is not a small corner: across `src/`, `include/mruby/` and the gems, 966 `else` keywords sit on their own line and 14 are attached. `ColumnLimit: 80` is the same shape of question, with 3.2% of the lines in `src/*.c` over that today. Whichever way each of those goes, the file decides it for every future patch rather than the guide doing so.

**A formatter config in the root reformats on save.** Most editors pick one up automatically, so a contributor editing three lines gets a diff touching a hundred, and the review has to separate the change from the churn. We hit exactly this with `prettier`, which globbed the whole worktree and rewrote a vendored submodule during a single-file run; it took a `.prettierignore` to contain it. I would rather not open the same door for C, where the blast radius is the whole core.

What the project relies on instead is `.editorconfig` for the mechanical part (2-space indent, LF, trailing whitespace, final newline) and `CONTRIBUTING.md` for the judgment calls. That split is deliberate: the parts a tool can settle without argument are automated, and the rest is written down for a person to apply. Consistency here has held up well without a formatter, which is why I do not think the gap you are aiming at is a missing config file.

None of this is a criticism of the config itself, and the motivation is a fair one to raise. If you find places where the code disagrees with `CONTRIBUTING.md`, a patch fixing those directly is welcome, and easier to review than a sweep.

leviongit:
re: "[a `.clang-format`] would rewrite code away from the style the project documents
while yes, we made a mistake with the `BraceWrapping.BeforeElse` setting, it is configurable, which means that setting it to the value of `true` would make it align with the style in `CONTRIBUTING.md`, same with the line width setting.

this is however a point which is based on an assumption that the `.clang-format` configuration file would never be tweaked, that file would, long term, *become* the project style, since it would define the target formatting.

re: "a formatter config in the root reformats on save"
aside from this pull request not introducing a pre-commit hook running `clang-format`, it is reasonable to only run it on *changed* files/lines

PR #7267, opened 2026-08-18
mruby-regexp: follow a block that changes the receiver in `gsub`, `sub!` and `scan`

takumin:
Rebased onto master now that #7274, which moves the block walk of `gsub` into C, is merged: the two commits here are all this PR is. The semantics, the tests and the differential run are the ones reviewed above, with the `gsub` part living in the C loop, where the cost question of that review does not arise. Every figure below is measured against master as it stands after that merge.

The block forms of `String#gsub`, `gsub!`, `sub!` and `scan` in mruby-regexp did not answer for a block that writes to the receiver the way CRuby's `str_gsub`, `rb_str_sub_bang` and `rb_str_scan` do. Measured against CRuby 4.0.6:

```ruby
s = "hello"; s.gsub(/l/) { s.tr!("h", "H"); "X" }
# CRuby: "HeXXo"   mruby before: "heXXo"   mruby after: "HeXXo"

s = "abc"; s.gsub(/b/) { s.replace("xyz"); "!" }
# CRuby: "x!z"     mruby before: "a!z"     mruby after: "x!z"

t = "hello"; n = 0
t.gsub(/l/) { n += 1; t.upcase! if n == 2; "X" }
[$&, $`, $']
# CRuby: [nil, nil, nil]   mruby before: ["l", "hel", "o"]   mruby after: [nil, nil, nil]

s = "abc"; s.gsub(/b/) { s << "zz"; "!" }
# CRuby: RuntimeError (string modified)   mruby before: "a!czz"   mruby after: RuntimeError (string modified)

s = "hello"; s.sub!(/l/) { s.upcase!; "X" }
# CRuby: "HEXLO"   mruby before: "heXlo"   mruby after: "HEXLO"

s = "hello"; s.scan(/l/) { s.upcase! }; $~ && $~[0]
# CRuby: nil       mruby before: "l"       mruby after: nil
```

Three things were off in what the loops read and when. The `gsub` loop copied the stretch before each match before the block ran, so a change the block made there was lost from the answer while the next match was already searched for in the changed string. A block that changed the length was let through in every loop, and the offsets of the match then named other bytes than the ones it matched; CRuby refuses one in `str_mod_check` right after the block returns. And the match left in `$~` was the MatchData of the last match the loop had, published again; CRuby searches once more when the loop is over, from the offset the last match was found from and on the receiver as it stands then, which is nil where the block wrote the match away and a fresh match on the changed string otherwise. `sub!` went down to `sub`, which builds the answer from the snapshot the MatchData holds, where `rb_str_sub_bang` splices the replacement into the receiver as the block left it.

## Fix

Two commits.

The first takes `gsub` and `sub!`. `Regexp.__gsub_block`, the C loop #7274 put on master, calls the block first and copies the stretch before the match afterwards, compares the receiver's byte length after each call against what it was when the loop began and raises `RuntimeError` where it differs (`str_mod_check` compares the buffer pointer too, which is no test here: mruby answers a write into a shared string with a buffer of its own, which is also why the bytes, their reading and the encoding check are taken afresh from the receiver the block left rather than through the pointer the search was given), keeps the offset each match was found from and the match itself, and ends on a search from that offset where the receiver no longer reads as it did when that match was made. Where it still does, the search could only find that match again, so the loop publishes it again: a MatchData holds a frozen copy of its subject, and what a search reads of a subject is its bytes and whether they are read by byte, so the receiver is compared against that copy on both counts. A change of length fails the comparison, so do `tr!` and `upcase!`, which keep the length and change the bytes, and so does `s.replace(s.b)`, which keeps every byte and changes the reading; where the receiver still shares its buffer with the copy, the pointer settles it. In C the length check is one comparison per turn and the copy test a pointer compare or a `memcmp` in the same function. `gsub!` inherits all of it. The block form of `sub!` runs the block itself in mrblib, applies the same length check, and splices the replacement into the receiver by the byte offsets of the match. `sub` is unchanged: CRuby's `rb_str_sub` works on a copy, so the block reaches nothing it reads.

The second takes `scan`, whose block loop stays in mrblib. It hands the length it began with to every search it makes after the block, and `Regexp.__byte_search` takes it as a fourth argument and raises `RuntimeError` before it looks where the receiver no longer has it, so the loop pays one argument per search rather than one `bytesize` call per match; it keeps the offset each match was found from and the match itself, and ends on a search from that offset unless `MatchData#__republish(str)`, the copy test above asked from mrblib, publishes the match again. `MatchData#__set_globals`, which republished the last match whatever the block had done to the receiver, loses its last caller there and goes.

## Testing

- `mrbgems/mruby-regexp/test/match_data.rb`: the test that pinned the republished MatchData after a `gsub` block mutated the subject now pins CRuby's answer (`$~` nil where the match was written away, a match on the changed string where it was left in place, and the closing search running from the offset the last match was found from).
- `mrbgems/mruby-regexp/test/string_regexp.rb`: return values of `gsub` / `gsub!` / `sub!` / `scan` under a block that changes the receiver in place, `RuntimeError` for a change of length (in bytes, and against the length at the start of the loop, so a change undone inside the block passes), the receiver keeping what the block did to it, `sub` unaffected, `FrozenError` for a receiver the block freezes, a receiver long enough that the write moves it off the buffer the match was made on, and, for the empty-match step, `s.clear` and `s.chop!` under `gsub(/x*/)` and `s.clear` under `scan(/x*/)`, which raise `RuntimeError` on CRuby. The assertion master gained in ba8a7c82d, a `replace` of the same length under `gsub(/(?=a)/)` that tells reading the receiver's bytes again from reading the pointer the search was made with, stands inside this rewritten test as it was written.
- `mrbgems/mruby-regexp/test/regexp_utf8.rb`: the match left behind after a block that changed the reading of the receiver without changing a byte (`s.replace(s.b)`) counts its offsets in bytes, as CRuby's does after a `force_encoding` in the block; a comparison of bytes alone would take that receiver for unchanged.

Every value in the tests was measured on CRuby 4.0.6 first. The differential run from the earlier review, the four block forms over 7 subjects, 8 patterns and 16 kinds of in-block change, 3584 cases printed the same way on both sides and diffed against CRuby 4.0.6: master differs on 992, the first commit here on 223 (all of them `scan`), the tip on 0.

Full suite green at every commit: `rake -m test` on the default configuration at each commit, and `MRUBY_CONFIG=build_config/ci/gcc-clang.rb rake -m test` and `MRUBY_CONFIG=build_config/gcc-asan.rb rake -m test` at the tip, no sanitizer report.

| Build | Total | KO | Crash |
| --- | --- | --- | --- |
| full-debug | 2377 | 0 | 0 |
| bintest | 2377 (+122 bintest) | 0 | 0 |
| cxx_abi | 2377 | 0 | 0 |
| byte-string | 2306 | 0 | 0 |
| ascii-ctype | 2373 | 0 | 0 |
| gcc-asan | 2377 (+84 bintest) | 0 | 0 |
| default | 2152 | 0 | 0 |

## Cost

Wall clock, the cases of the earlier review plus the String pattern block forms and a `scan` over the long subject, minimum of 5 alternating runs, `-O3`, default configuration. master has the C walk of #7274, so the block rows start from where that PR left them; the percentages are against master.

```ruby
s480 = "hello world foo bar " * 24
30000.times { s480.gsub(/o/) { } }        # gsub480_blk
100000.times { "abc".gsub(/b/) { } }      # gsub_abc_blk
100000.times { "abc".scan(/b/) { } }      # scan_abc_blk
100000.times { "abc".sub!(/b/) { } }      # sub!_abc_blk
30000.times { s480.gsub(/o/, "0") }       # gsub480_str
100000.times { "abc".scan(/b/) }          # scan_abc
100000.times { "abc".sub!(/b/, "0") }     # sub!_abc_str
30000.times { s480.gsub(/z/) { } }        # gsub480_none
30000.times { s480.scan(/o/) { } }        # scan480_blk
30000.times { s480.gsub("o") { } }        # lit_gsub480_blk
100000.times { "abc".gsub("b") { } }      # lit_gsub_abc_blk
100000.times { "abc".sub!("b") { } }      # lit_sub!_abc_blk
100000.times { s = "hello"; s.gsub(/l/) { s.tr!("h", "H"); "X" } }        # gsub_tr
30000.times { s = s480.dup; s.gsub(/o/) { s.tr!("h", "H"); "0" } }        # gsub480_tr
100000.times { s = "hello"; s.scan(/l/) { s.tr!("h", "H") } }             # scan_tr
```

| case | master | commit 1 | this PR |
| --- | --- | --- | --- |
| gsub480_blk | 1207ms | 1231ms (+2%) | 1207ms |
| gsub_abc_blk | 160ms | 164ms (+3%) | 164ms (+3%) |
| scan_abc_blk | 174ms | 172ms (-1%) | 182ms (+5%) |
| sub!_abc_blk | 212ms | 172ms (-19%) | 170ms (-20%) |
| gsub480_str | 118ms | 118ms | 118ms |
| scan_abc | 117ms | 118ms (+1%) | 117ms |
| sub!_abc_str | 193ms | 190ms (-2%) | 192ms (-1%) |
| gsub480_none | 37ms | 38ms (+3%) | 38ms (+3%) |
| scan480_blk | 1592ms | 1580ms (-1%) | 1612ms (+1%) |
| lit_gsub480_blk | 1189ms | 1216ms (+2%) | 1201ms (+1%) |
| lit_gsub_abc_blk | 157ms | 158ms (+1%) | 157ms |
| lit_sub!_abc_blk | 266ms | 168ms (-37%) | 167ms (-37%) |
| gsub_tr | 222ms | 221ms | 223ms |
| gsub480_tr | 4286ms | 4293ms | 4318ms (+1%) |
| scan_tr | 260ms | 257ms (-1%) | 266ms (+2%) |

The `gsub` rows: the length comparison, the reading and the encoding flag read per turn cost about 2% on the 480 byte subject in the first commit (24ms over 2.16 million turns), a spread the tip's run of the same code lands inside, and 3% at most on the short ones; the closing test is the pointer alone there, since the receiver still shares its buffer with the snapshot. `sub!` with a block no longer goes down to `sub`, which is the -20% and -37% on its two rows. The `scan` rows carry the argument the search now parses and the `__republish` call, 1% on the long subject and 5% on the short one, and `scan` on `"abc"` without a block, `gsub(/o/, "0")` and `sub!(/b/, "0")` are unchanged. The mutating rows (`gsub_tr`, `gsub480_tr`, `scan_tr`) are where the closing search runs, and they sit on master to within 2%.

## Size

`.text` of `bin/mruby`, `build_config/ci/gcc-clang.rb`, each side from a clean
build directory. `regexp.o` is the object that changes, and it accounts for the
whole delta in every build (to within 5 bytes of alignment in `full-debug`):
the copy test and the closing search in `__gsub_block`, `__republish` and the
length argument of `__byte_search`. The mrblib side, the block form of `sub!`
and the closing search of `scan`, is bytecode in `mruby-regexp/gem_init.o`'s
`.rodata`, which grows by 128 (160 in `full-debug`).

| build | master | commit 1 | this PR | delta |
|---|---:|---:|---:|---:|
| `bintest` | 1,287,414 | 1,287,718 | 1,288,134 | +720 |
| `ascii-ctype` | 1,275,302 | 1,275,606 | 1,276,022 | +720 |
| `byte-string` | 1,257,014 | 1,257,318 | 1,257,734 | +720 |
| `cxx_abi` | 1,313,017 | 1,313,353 | 1,313,593 | +576 |
| `full-debug` (`-O0`) | 1,886,710 | 1,887,062 | 1,887,318 | +608 |

On the default configuration, the one the wall clock was measured on, `.text` is
1,204,174 on master, 1,204,478 after the first commit and 1,204,894 at the tip
(+720).

### Environment

<details>
<summary>Machine, toolchain, and the compile line of every build</summary>

| Item | Value |
| --- | --- |
| OS | Ubuntu 24.04.4 LTS |
| Kernel | 7.0.0-29-generic |
| CRuby (reference) | ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [x86_64-linux] |

Actual compile line of `mrbgems/mruby-regexp/src/regexp.c` in each build (`-MMD -c`, `-I`, and `-o` dropped). `full-debug` and `gcc-asan` are `-O0` because `enable_debug` appends `-g3 -O0` after the toolchain's `-g -O3`; `cxx_abi` compiles C as C++ with `gcc -x c++ -std=gnu++03`, g++ only links.

```console
# full-debug
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -g3 -O0 -DMRB_GC_STRESS -DMRB_USE_DEBUG_HOOK -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_DEBUG -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/regexp.c
# bintest
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRB_GC_FIXED_ARENA -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER -DMRB_USE_DEBUG_HOOK mrbgems/mruby-regexp/src/regexp.c
# cxx_abi
gcc -g -O3 -Wall -Wundef -Wwrite-strings -x c++ -std=gnu++03 -DMRB_GC_FIXED_ARENA -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_CXX_EXCEPTION -DMRB_USE_CXX_ABI -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/regexp.c
# byte-string
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/regexp.c
# ascii-ctype
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRB_USE_ASCII_CTYPE -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/regexp.c
# gcc-asan
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -fsanitize=address,undefined -g3 -O0 -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_DEBUG -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/regexp.c
# default (no MRUBY_CONFIG), the build every figure above was measured on
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DMRB_USE_COMPLEX -DMRB_USE_BIGINT -DMRB_USE_DEBUG_HOOK mrbgems/mruby-regexp/src/regexp.c
```

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `String#sub`, `gsub`, `scan`, and bang variants when the source string is modified during processing.
  * Added validation to detect source-length changes and prevent inconsistent results.
  * Preserved match information more reliably after substitutions and callbacks.
  * Improved handling of literal string patterns, replacement expansions, empty patterns, frozen strings, and invalid byte sequences.
  * Corrected UTF-8 and byte-based match indexing after block substitutions.

### DIFF
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -169,18 +169,29 @@ class String
     end
     pattern = Regexp.new(Regexp.escape(pattern)) if literal
     # A full search and not `match?`, so a failed match clears $~.
-    return nil unless Regexp.__search(pattern, self, 0, literal)
-    # `sub` matches again and publishes its own $~ over this one, leaving the
-    # caller the match `sub` would have left, a block's own matches included.
-    # The resolved pattern takes the place of the original argument so that a
-    # String is not quoted and compiled a second time; a literal goes down as
-    # the String it was instead, since that is what tells `sub` to leave the
-    # subject unread, and quoting it twice is the price of saying so.
-    # Overwriting `self` afterwards is safe: a MatchData snapshots its subject,
-    # so $~ keeps describing the string as it was matched.
-    down = literal ? args[0] : pattern
-    str = argc == 2 ? self.sub(down, args[1], &block) : self.sub(down, &block)
-    self.replace(str)
+    md = Regexp.__search(pattern, self, 0, literal)
+    return nil unless md
+    if argc == 2
+      # `sub` matches again and publishes its own $~ over this one, leaving
+      # the caller the match `sub` would have left.  The resolved pattern
+      # goes down so that it is not compiled a second time; a literal with a
+      # replacement never reaches here, having been answered by `__sub_lit`
+      # above.  Overwriting `self` afterwards is safe: a MatchData snapshots
+      # its subject, so $~ keeps describing the string as it was matched.
+      return self.replace(self.sub(pattern, args[1]))
+    end
+    # The block form does not go down to `sub`, which builds its answer from
+    # the snapshot the MatchData holds, because CRuby's `rb_str_sub_bang`
+    # builds it from the receiver as the block left it: `s = "hello";
+    # s.sub!(/l/) { s.upcase!; "X" }` is "HEXLO" there, where `sub` on the
+    # same receiver is "heXlo".  It refuses a block that changed the length
+    # first, as `gsub` does, so the offsets of the match still name the bytes
+    # they named.  $~ stays what the search above published, or whatever the
+    # block put there.
+    len = self.bytesize
+    val = block.call(md[0]).to_s
+    raise RuntimeError, "string modified" if self.bytesize != len
+    self.replace(self.byteslice(0, md.__byte_begin(0)) + val + self.byteslice(md.__byte_end(0)..-1))
   end
 
   def gsub(*args, &block)
--- a/mrbgems/mruby-regexp/mrblib/string_regexp.rb
+++ b/mrbgems/mruby-regexp/mrblib/string_regexp.rb
@@ -272,20 +272,30 @@ class String
     # a character on its own.
     pos = 0
     len = self.bytesize
-    # The loop ends on a failed search, which clears the globals. CRuby leaves
-    # the last match behind, so keep it and republish it below, the way `gsub`
-    # does. A scan that matched nothing keeps the cleared state.
+    # A block that changes the receiver is answered for as `rb_str_scan`
+    # answers for it, which is the way `__gsub_block` does: one that changed
+    # the length is refused with `RuntimeError` by the next search, which takes
+    # `len` for that, the next match is searched for in the string it left,
+    # and the match left in $~ is a search once more from the offset the last
+    # match was found from, on the string as it stands at the end. That search
+    # also republishes what the failed one that ends the loop clears; a scan
+    # that matched nothing keeps the cleared state. And as in `gsub`, a
+    # receiver that still reads as it did when the last match was made gets
+    # that match republished, and the search runs only where `__republish`
+    # finds it reading differently.
     last = nil
+    last_md = nil
     while pos <= len
-      md = Regexp.__byte_search(pattern, self, pos)
+      md = Regexp.__byte_search(pattern, self, pos, len)
       break unless md
-      last = md
+      last = pos
+      last_md = md
       yield(md.size == 1 ? md[0] : md.captures)
       match_start = md.__byte_begin(0)
       match_end = md.__byte_end(0)
       pos = match_start == match_end ? match_end + 1 : match_end
     end
-    last.__set_globals if last
+    Regexp.__byte_search(pattern, self, last, len) if last && !last_md.__republish(self)
     self
   end
 
--- a/mrbgems/mruby-regexp/src/regexp.c
+++ b/mrbgems/mruby-regexp/src/regexp.c
@@ -541,23 +541,32 @@ regexp_s_search(mrb_state *mrb, mrb_value klass)
 }
 
 /*
- * Regexp.__byte_search(re, str, pos = 0)
+ * Regexp.__byte_search(re, str, pos = 0, len = -1)
  *
  * Internal: the byte-offset search the mrblib loops of `scan`, `split` and
  * `byteindex` drive themselves. No position normalization, because the
  * callers already work in byte space, and no operand conversion, because
- * they always pass a String. The subject is the one the loop holds fixed, so
- * the check reads the flag core left on it after the first turn. Nor is there
- * a `checked` any more: `gsub` was the caller that set it, and its loop is
- * `__gsub_block` now, which takes the flag itself.
+ * they always pass a String. The subject is the one the loop holds, so the
+ * check reads the flag core left on it after the first turn, and walks it
+ * again only where a block has written to it in between. Nor is there a
+ * `checked` any more: `gsub` was the caller that set it, and its loop is
+ * `__gsub_block` now, which takes the flag itself. `len`, where the caller
+ * gives one, is the byte length its loop began with, and a subject that no
+ * longer has it is refused before the search: this is `str_mod_check` for
+ * the block loop of `scan`, asked here so that the loop pays one argument
+ * per search rather than one `bytesize` call per match.
  */
 static mrb_value
 regexp_s_byte_search(mrb_state *mrb, mrb_value klass)
 {
   mrb_value re, str;
   mrb_int pos = 0;
+  mrb_int len = -1;
 
-  mrb_get_args(mrb, "oS|i", &re, &str, &pos);
+  mrb_get_args(mrb, "oS|ii", &re, &str, &pos, &len);
+  if (len >= 0 && RSTRING_LEN(str) != len) {
+    mrb_raise(mrb, E_RUNTIME_ERROR, "string modified");
+  }
   check_regexp_arg(mrb, re);
   /* Every mrblib loop enters at zero or at an offset a match answered with,
      so a position before the subject reaches here only from a direct call.
@@ -1098,12 +1107,12 @@ matchdata_byte_end(mrb_state *mrb, mrb_value self)
   return mrb_int_value(mrb, pos);
 }
 
-/* Whether `str` still reads as the subject `md` was made on. The block loop
-   of `gsub` ends on a search from the offset its last match was found from,
-   on the receiver as the block left it, the way CRuby's `str_gsub` does; on a
-   receiver that reads as it did when that match was made, the search can only
-   find that match again, and the loop republishes it instead. What a search
-   reads of a subject is its bytes
+/* Whether `str` still reads as the subject `md` was made on. The block loops
+   of `gsub` and `scan` end on a search from the offset their last match was
+   found from, on the receiver as the block left it, the way CRuby's
+   `str_gsub` and `rb_str_scan` do; on a receiver that reads as it did when
+   that match was made, the search can only find that match again, and the
+   loops republish it instead. What a search reads of a subject is its bytes
    and whether they are read by byte, and `source` is a frozen copy of both as
    they stood at match time, so comparing the two is the whole of the test:
    where CRuby's `str_mod_check` reads the buffer pointer and the length, this
@@ -1122,18 +1131,24 @@ re_subject_reads_as(mrb_state *mrb, mrb_value str, mrb_value mdv)
   return p == q || memcmp(p, q, (size_t)len) == 0;
 }
 
-/* Private: republish $~ and the thirteen names derived from it. Used by the
-   mrblib loops that drive Regexp.__byte_search themselves, where the failing
-   call that ends the loop clears the match the loop is supposed to leave behind.
-   The names other than $~ are not assignable from Ruby, so restoring them
-   has to come from here. */
+/* Private: publish this match once more, if `str` still reads as the subject
+   it was made on, and say whether it did. The mrblib loop of `scan` ends on
+   a search from the offset its last match was found from, on the receiver as
+   the block left it, the way `rb_str_scan` does; `re_subject_reads_as()`
+   above is the test that spares that search, and this is it asked from
+   mrblib. The names other than $~ are not assignable from Ruby, so publishing
+   them again has to come from here. Returns false where `str` reads
+   differently, and the caller searches. */
 static mrb_value
-matchdata_set_globals(mrb_state *mrb, mrb_value self)
+matchdata_republish(mrb_state *mrb, mrb_value self)
 {
+  mrb_value str;
+  mrb_get_args(mrb, "S", &str);
   mrb_match_data *md = DATA_GET_PTR(mrb, self, &matchdata_type, mrb_match_data);
-  if (!md) return mrb_nil_value();
+  if (!md) return mrb_false_value();
+  if (!re_subject_reads_as(mrb, str, self)) return mrb_false_value();
   set_match_globals(mrb, self, md->source, md->captures, md->num_captures);
-  return self;
+  return mrb_true_value();
 }
 
 /*
@@ -2108,7 +2123,7 @@ mrb_mruby_regexp_gem_init(mrb_state *mrb)
   mrb_define_class_method(mrb, re, "__check_pattern", regexp_check_pattern, MRB_ARGS_REQ(1));
   mrb_define_class_method(mrb, re, "__check_byte_pos", regexp_check_byte_pos, MRB_ARGS_REQ(2));
   mrb_define_class_method(mrb, re, "__search", regexp_s_search, MRB_ARGS_ARG(2, 2));
-  mrb_define_class_method(mrb, re, "__byte_search", regexp_s_byte_search, MRB_ARGS_ARG(2, 1));
+  mrb_define_class_method(mrb, re, "__byte_search", regexp_s_byte_search, MRB_ARGS_ARG(2, 2));
   mrb_define_class_method(mrb, re, "__byte_rsearch", regexp_s_byte_rsearch, MRB_ARGS_REQ(3));
   mrb_define_class_method(mrb, re, "__search_p", regexp_s_search_p, MRB_ARGS_ARG(2, 1));
 
@@ -2177,7 +2192,7 @@ mrb_mruby_regexp_gem_init(mrb_state *mrb)
   mrb_define_method(mrb, md, "end", matchdata_end, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, md, "__byte_begin", matchdata_byte_begin, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, md, "__byte_end", matchdata_byte_end, MRB_ARGS_REQ(1));
-  mrb_define_method(mrb, md, "__set_globals", matchdata_set_globals, MRB_ARGS_NONE());
+  mrb_define_method(mrb, md, "__republish", matchdata_republish, MRB_ARGS_REQ(1));
   mrb_define_method(mrb, md, "pre_match", matchdata_pre, MRB_ARGS_NONE());
   mrb_define_method(mrb, md, "post_match", matchdata_post, MRB_ARGS_NONE());
   mrb_define_method(mrb, md, "named_captures", matchdata_named_captures, MRB_ARGS_NONE());
--- a/mrbgems/mruby-regexp/src/regexp.c
+++ b/mrbgems/mruby-regexp/src/regexp.c
@@ -1098,6 +1098,30 @@ matchdata_byte_end(mrb_state *mrb, mrb_value self)
   return mrb_int_value(mrb, pos);
 }
 
+/* Whether `str` still reads as the subject `md` was made on. The block loop
+   of `gsub` ends on a search from the offset its last match was found from,
+   on the receiver as the block left it, the way CRuby's `str_gsub` does; on a
+   receiver that reads as it did when that match was made, the search can only
+   find that match again, and the loop republishes it instead. What a search
+   reads of a subject is its bytes
+   and whether they are read by byte, and `source` is a frozen copy of both as
+   they stood at match time, so comparing the two is the whole of the test:
+   where CRuby's `str_mod_check` reads the buffer pointer and the length, this
+   reads the bytes themselves, and a change of length, of contents or of
+   reading each fail it. A receiver that still shares its buffer with the copy
+   is told apart by the pointer alone. */
+static mrb_bool
+re_subject_reads_as(mrb_state *mrb, mrb_value str, mrb_value mdv)
+{
+  mrb_match_data *md = DATA_GET_PTR(mrb, mdv, &matchdata_type, mrb_match_data);
+  mrb_value src = md->source;
+  mrb_int len = RSTRING_LEN(src);
+  if (RSTRING_LEN(str) != len) return FALSE;
+  if (re_binary_string_p(str) != re_binary_string_p(src)) return FALSE;
+  const char *p = RSTRING_PTR(str), *q = RSTRING_PTR(src);
+  return p == q || memcmp(p, q, (size_t)len) == 0;
+}
+
 /* Private: republish $~ and the thirteen names derived from it. Used by the
    mrblib loops that drive Regexp.__byte_search themselves, where the failing
    call that ends the loop clears the match the loop is supposed to leave behind.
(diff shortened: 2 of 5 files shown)


takumin:
@coderabbitai review

matz:
The behavior is right and I reproduced all of it. A random comparison of 500 cases (the five methods over 14 kinds of in-block mutation) against CRuby 3.2.3 has master differing on 70 and this branch on 0. I also planted each of the five changes back one at a time, and the new tests caught every one. `host-debug`, the default configuration, all six `ci/gcc-clang` builds and `clang-asan` are green.

Before I merge I would like to ask about the closing `Regexp.__byte_search`. It is one extra search per call that matched, and it lands only on the block forms. Minimum of three runs, `-O3`, default configuration:

| case | master | this branch |
| --- | --- | --- |
| `s.gsub(/o/) { }`, 480 byte subject, 30k calls | 1748ms | 1826ms (+4%) |
| `"abc".gsub(/b/) { }`, 100k calls | 193ms | 220ms (+14%) |
| `"abc".scan(/b/) { }`, 100k calls | 151ms | 175ms (+16%) |
| `"abc".sub!(/b/) { }`, 100k calls | 194ms | **148ms (-24%)** |

`gsub(/o/, "0")`, `scan` without a block, `sub!(pat, str)` and a `gsub` that matched nothing all stay within noise, so `sub!` with a block gets faster by not going down to `sub`, and the cost is confined to the two loops, which is where the semantics changed.

I am not asking you to trade the semantics away. My question is whether the search can be skipped where it cannot tell us anything new: if the receiver is what it was when the loop started, a search from `last` finds the match the loop already had, and republishing that one would do. Comparing `bytesize` does not establish that, since `tr!` and `upcase!` keep the length, so I have no test for it I trust. If you see one, I would take it.

There is a second angle. Every search copies the subject into `` $` `` and `$'` eagerly, which is why the long subject pays 4% even though it has many matches, and making those lazy is already on my list from #7148 and #7149. If you think that work would absorb this cost, say so and I will merge as it stands and leave the search alone.

takumin:
There is a test I trust, and the loop was holding it all along. `create_matchdata` snapshots the subject with `mrb_str_dup_frozen`, so the last MatchData the loop has carries a frozen copy of the receiver as it stood when that match was made. What a search reads of a subject is its bytes and whether they are read by byte, so comparing the receiver against that copy is the whole of the question: a change of length fails it, `tr!` and `upcase!` fail it because they change the bytes, and so does a change of reading with no change of bytes (`s.replace(s.b)`), which the search would answer differently and `str_mod_check` would let through. Where the receiver still shares its buffer with the copy, the pointer settles it; otherwise it is a `memcmp` over the receiver's length, which is less than a search.

The reference is the copy the last match holds, not one taken when the loop started, so nothing is taken at loop start and it is the tighter question: the closing search runs from `last` on the receiver as it stands, and where the receiver reads as it did when the match from `last` was made, that search can only find that match again. A block that changed the receiver in an early turn and left it alone afterwards passes.

I have reworked the branch with this. `MatchData#__republish(str)` takes the place of `MatchData#__set_globals`: it publishes the match again and returns true where `str` reads as its subject, and false where the caller has to search, so both loops end on `Regexp.__byte_search(...) if last && !last_md.__republish(self)`. A test in regexp_utf8.rb pins the reading, since bytes alone would take `s.replace(s.b)` for unchanged and republish a match that counts in characters where CRuby's counts in bytes. The differential run is 3584 cases (the four block forms, 7 subjects, 8 patterns, 16 kinds of in-block change) against CRuby 4.0.6: master differs on 992, the branch on 0 before and after. Full suite green at every commit on `ci/gcc-clang` and the default configuration.

Timings the way you took them, minimum of 7, `-O3`, default configuration; `.text` of `bin/mruby` is 1198526 on master, 1197374 as pushed, 1197822 now:

| case | master | as pushed | now |
| --- | --- | --- | --- |
| `s.gsub(/o/) { }`, 480 byte subject, 30k calls | 2755ms | 2904ms (+5%) | 2921ms (+6%) |
| `"abc".gsub(/b/) { }`, 100k calls | 268ms | 303ms (+13%) | 276ms (+3%) |
| `"abc".scan(/b/) { }`, 100k calls | 204ms | 239ms (+17%) | 212ms (+4%) |
| `"abc".sub!(/b/) { }`, 100k calls | 263ms | 203ms (-23%) | 201ms (-24%) |
| `s.gsub(/l/) { s.tr!("h", "H"); "X" }`, 100k calls | 379ms | 425ms | 395ms |

The republish is what closes the two short cases. What is left on the 480 byte subject is not the closing search but the length check itself: one `bytesize` call per match, 72 per call there, which is what `str_mod_check` costs when it is written in mrblib. A build with that line taken out lands on master in every row (2784ms, 271ms, 207ms), so the 6% is the check and nothing else. Folding it into the next `__byte_search` would spare the call, but it moves the check past the `byteslice` that reads the stretch before the match, and I would rather it stay where CRuby's is.

On the second angle: since 8bd5bd0ea, `` $` `` and `$'` are windows on the subject for anything past the embed length, so what a publish still pays on a long subject is the objects, the MatchData with its snapshot and the four RStrings, and thirteen `mrb_gv_set`. Lazy publishing would trim part of that per search; here the republish removes the whole closing search wherever the receiver stands, which is every row above but the mutating one, so I would not leave it to that work.

takumin:
One more measurement, so that the "I would rather it stay" above has a number behind it. The length check can be folded into the next `Regexp.__byte_search` (a fifth argument, the raise from C, a `bytesize` left only in the empty-match step, which reads the receiver before that search), and that recovers about half of what the check costs on a long subject, since the search then parses one more argument per match. Same setup as above, minimum of 7:

| case | master | this branch | check taken out | check folded |
| `s.gsub(/o/) { }`, 480 byte subject, 30k calls | 2749ms | 2940ms (+7%) | 2776ms (+1%) | 2836ms (+3%) |
| `s.gsub(/o/) { }`, 4800 byte subject, 3k calls | 2796ms | 2954ms (+6%) | 2804ms (+0%) | 2837ms (+1.5%) |
| `s.scan(/o/) { }`, 480 byte subject, 30k calls | 1963ms | 2103ms (+7%) | 1991ms (+1%) | 2040ms (+4%) |
| `"abc".gsub(/b/) { }`, 100k calls | 260ms | 268ms | 264ms | 266ms |

The differential run is 0 for the folded variant too. The patch, the bench and the differential script are in https://gist.github.com/takumin/52b77788594a38040b42c64d0e2d3748. I have left the branch as it is: what the fold buys is 3 to 4% on a subject with many matches, and it puts the check in two places, gives `__byte_search` an argument only the loops use, and raises from C. If you would rather have it, it is a small change and I will fold it in.

takumin:
I have folded the length check in after all, as a fourth commit (`584cd522a`); the first three are as you reviewed them. I measured again in one session, your four cases the way you took them, minimum of 3, `-O3`, default configuration, on master `c6a3fcb9a`, the branch as it stood, and the branch with the check folded into the next `__byte_search`, in that order, and the fold keeps its share on every row where the check was what remained, so leaving it out was only a preference for the check standing where CRuby's does, and that is not observable.

| case | master | before the fourth commit | this branch |
| --- | --- | --- | --- |
| `s.gsub(/o/) { }`, 480 byte subject, 30k calls | 2767ms | 2940ms (+6%) | 2907ms (+5%) |
| `"abc".gsub(/b/) { }`, 100k calls | 261ms | 273ms (+5%) | 268ms (+3%) |
| `"abc".scan(/b/) { }`, 100k calls | 207ms | 218ms (+5%) | 213ms (+3%) |
| `"abc".sub!(/b/) { }`, 100k calls | 263ms | **200ms (-24%)** | **201ms (-24%)** |

`gsub(/o/, "0")`, `scan` without a block, `sub!(pat, str)` and a `gsub` that matched nothing stay within noise (124/123ms, 156/155ms, 236/238ms, 156/159ms, master and branch). The PR body now carries the minimum of 7 from the same session with the 4800 byte subject and the `scan` loop added, where the 480 byte `gsub` is +6% before the fold and +2% after, and the `scan` loop +10% and +5%.

What the commit does: the loops hand the length they began with to every search they make after the block, the next one and the closing one, and `Regexp.__byte_search` takes it as a fifth argument and raises `RuntimeError` before it looks. Nothing between the block and that search reads a byte the offsets could misname, except the step over an empty match, which reads the receiver for the character it steps over and keeps a `bytesize` check of its own (a shrunk receiver would otherwise make a `NoMethodError` of it, which is what master answers for `s = "ab"; s.gsub(/x*/) { s.clear; "!" }`); `sub!` has no search after its block, so its check stays in mrblib. Three tests pin that step, and taking that check out fails them. Full suite green at the tip on the five `ci/gcc-clang` builds and the default configuration, and the 3584 case differential against CRuby 4.0.6 is still 0.

takumin:
I have moved this back to draft, and here is why.

The closing search and the length check cost what they cost because the loop is in mrblib: every check is a method call and every search a frame. I have the block walk of `gsub` running in C, around one `mrb_yield`, and there your cases sit far below master rather than above it (`s.gsub(/o/) { }` on the 480 byte subject 2324ms to 1200ms, `"abc".gsub(/b/) { }` 233ms to 162ms, minimum of 5, `-O3`, default configuration). In that loop the length check is one comparison per turn, and the closing question, whether the receiver still reads as the snapshot the last match holds, is a pointer compare or a `memcmp` in the same function, so the cost you asked about goes away rather than being trimmed. Given that, I would rather not ask you to review a check in mrblib that the C loop then replaces.

The plan is two steps. First a PR of its own that moves the `gsub` block walk into C, a pure speed-up that changes no answer except one: `$~.regexp` after a String pattern becomes the same object for the same literal asked for twice running, as it is in CRuby. Then this PR reworked on top of it: the same five behaviours and the same tests, the 3584 case differential kept at 0, with the `gsub` check and republish living in the C loop and `scan`, `sub` and `sub!` as they are here. Nothing about the semantics you reproduced changes; only where the `gsub` part sits. I will take this out of draft once the first PR is up.

takumin:
The C loop is up as #7274, and this PR is now stacked on it: the first commit here is that PR's, the two after it are this one, and the body is rewritten for the new shape.

What changed against what you reviewed is where the `gsub` part sits, not what it does. The length check, the copy of the stretch before the match after the block, the closing search from the offset the last match was found from and the republish where the receiver still reads as the snapshot are all inside `Regexp.__gsub_block` now, so the check is one comparison per turn and the copy test a pointer compare or a `memcmp` in the same function; `__byte_search` takes the length as its fourth argument for `scan`, whose loop stays in mrblib with `MatchData#__republish`, and `sub!` is as before. Same tests, and the 3584 case differential against CRuby 4.0.6 is 992 on master, 992 on #7274, 0 at the tip.

On your cost table, against master, minimum of 5: `s.gsub(/o/) { }` over 480 bytes 2416ms to 1241ms, `"abc".gsub(/b/) { }` 235ms to 165ms, `"abc".scan(/b/) { }` 180ms to 185ms, `"abc".sub!(/b/) { }` 231ms to 174ms. The full table with the #7274 column between is in the body.

takumin:
Rebased (`7a8a865ab` → `9775cf0f2`) onto the amended #7274 (`29d33be97`, which itself now sits on master `48b986b92`). The two commits here are unchanged apart from the base; the size figures in the description are re-measured against it (+720 over the base as before), and the suite is green again on the default configuration at each commit and on `ci/gcc-clang.rb` and `gcc-asan.rb` at the tip.

takumin:
@coderabbitai review

takumin:
@coderabbitai resume

takumin:
Rebased onto master now that #7274 is merged. The branch is the two commits of this PR (`8445f1cc6` for `gsub` and `sub!`, `1a5e26b46` for `scan`); the code is unchanged.

The one conflict was in `test/string_regexp.rb`: ba8a7c82d added an assertion to the test block this PR rewrites, the same-length `replace` under `gsub(/(?=a)/)` that tells reading the receiver's bytes again from reading the pointer the search was made with. That assertion and its comment now stand inside the rewritten test, as written there.

The body is updated to measure against master as it stands: the differential run (master 992, first commit 223, tip 0 of 3584), the test totals, the wall clock (now three columns, master / commit 1 / this PR) and the `.text` sizes (+720 on `bintest`, `ascii-ctype` and `byte-string`, +576 on `cxx_abi`, +608 on `full-debug`, the same deltas as before).

PR #7286, opened 2026-08-19
Simplify some control flow in `vm.c`

leviongit:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when returning from strict procedures, blocks, and self-invoked routines.
  - Improved handling of calls between virtual machines, helping prevent unexpected execution-flow issues.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -3101,8 +3101,7 @@ RETRY_TRY_BLOCK:
         mrb_raisef(mrb, E_TYPE_ERROR, "wrong type %T (expected Proc)", regs[a]);
       }
       const struct RProc *p = mrb_proc_ptr(regs[a]);
-      ci = cipush(mrb, a, CINFO_DIRECT, NULL, NULL, NULL, 0, b);
-      ci->cci = CINFO_NONE;  /* mark as VM-to-VM call for proper break handling */
+      ci = cipush(mrb, a, CINFO_NONE, NULL, NULL, NULL, 0, b);
       int r = vm_call_proc(mrb, p, b+1, &irep, ai);
       ci = mrb->c->ci;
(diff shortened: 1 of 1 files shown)


matz:
Thanks for looking at this, but the return opcodes cannot take their value out of a register.

`OP_RETNIL`, `OP_RETTRUE` and `OP_RETFALSE` now write into `regs[1]`, and `regs[1]` is not scratch: it is the first argument or local variable of the frame that is returning. Those registers are also the storage of a block's environment while the frame is alive, and the environment is moved to the heap only when the frame is popped, a few lines further down. So whatever the return wrote is what the block reads back afterwards:

```ruby
def m(x)
  $p = Proc.new { x }
  return
end
m(42)
p $p.call            # master: 42, this PR: nil

def m2(x, y)
  $q = Proc.new { [x, y] }
  return false
end
m2(1, 2)
p $q.call            # master: [1, 2], this PR: [nil, 2]
```

`return true` corrupts the same slot. Nothing in the suite read a captured variable back after one of those returns, which is why CI is green here; I have added that test on master (3c583f2b9), and with this branch on top of it the suite reports `KO: 1`.

The two other parts do hold. Passing `CINFO_NONE` straight to `cipush()` in `OP_BLKCALL` is the same store the next line made, and reaching `L_OP_RETURN_BODY` instead of a second label inside the same case is what `OP_CALL` and `OP_BLKCALL` already do above. If you want to keep those, the value still has to come from a local, the way `v = mrb_nil_value()` does today, so the three labels have to stay.

leviongit:
thank you for rectifying our misunderstanding of the handling of the return opcodes, for some reason we managed to miss their the broken behavior, we'll remove that patch shortly

PR #7307, opened 2026-08-22
mruby-regexp: move the backtracking stack off the C stack

takumin:
## Summary

`bt_match()` recursed at every fork, so the C stack *was* the backtracking stack and `MRB_REGEXP_RECURSION_LIMIT` was what kept a long subject from overflowing it. A greedy repetition forks once per iteration whatever its body holds, so what reached the limit was the length of the run and not the nesting of the pattern: `/(?:(?>a))*/` gave up after 332 iterations and `/(?:a)*(b)\1/` after 996. The state moves onto two stacks of the engine's own, on the heap; `bt_match()` becomes one loop, and a search spends one C frame however long the subject is. What the limit counts moves with it: entries a search would have to put back, rather than C frames or calls made.

## Changes

| File | What |
| --- | --- |
| `src/re_exec.c` | added `re_cpoint`, `re_undo`, `bt_room()`, `bt_grow_capa()`, `bt_push()`, `bt_log()`, `bt_undo_to()`, `bt_barrier_find()`, `bt_iter_begin()`; removed `bt_iter()`, `bt_look()`, `BT_CUT()` and `bt_match()`'s `depth` parameter; `bt_log()` records a write only where the slot changes |
| `include/re_internal.h` | `MRB_REGEXP_RECURSION_LIMIT` (1000) becomes `MRB_REGEXP_STACK_LIMIT` (2048), the old name an `#error`; `RE_OVER_RECURSION_LIMIT` becomes `RE_OVER_STACK_LIMIT`, and `RE_NOMEM` joins it; an `#error` bounds what the new macro may be set to |
| `src/regexp.c` | `Regexp::RECURSION_LIMIT` becomes `Regexp::STACK_LIMIT`; `re_check_over_limit()` becomes `re_check_exec_error()`, which raises `NoMemoryError` on `RE_NOMEM` and `RegexpError` naming the knob on either limit |
| `README.md` | the two limits described as the work a search may do and the state it may hold, with what the stack limit does and does not bound |
| `test/regexp_syntax.rb` | five asserts added for what no longer costs C stack; the limit asserts re-sized from the new constant, and skipped on a build whose two limits leave them nothing to pin; the step limit pinned by a pattern that holds little while spending much; the blocks that mixed parser checks with engine checks split so that only the latter carry the skip |
| `test/string_index.rb`, `test/string_regexp.rb` | the limit subjects re-sized from the new constant; the same split |
| `test/backtracking_stack.rb` | new: what a test that reaches the backtracking engine asks of the build's limit (48), and the skip where it stands below that |

### Two stacks rather than one

A choice point is a branch not taken: where the input stood (`sp`), which instruction takes the branch (`pc`), and how tall the undo log was when it was pushed. An undo record is one write to take back: the slot, and what stood in it. Backtracking pops a choice point, unwinds the log to its height, and goes on from there.

They are separate stacks because what a cut does to each differs. The captures a positive lookaround or an atomic group wrote outlive it, so its cut drops the choice points above its barrier and leaves the log alone, where a negative lookaround unwinds both. One mixed stack could not truncate at all: it would have to walk the region above the barrier and keep the undo records while dropping the choice points.

### A cut is a truncation

Entering an atomic group or a lookaround pushes a barrier onto the choice point stack. Reaching one while backtracking is that group failing, so it resumes nothing and the search goes on popping; its end drops the barrier and everything above it. A negative lookaround's barrier is the one that resumes rather than keeps being popped, its sub-pattern running out of alternatives being the assertion holding. That is what `BT_CUT` was, and it goes with the recursion.

Two things ride on the barrier rather than on the undo log. The pass a lookaround was entered from is one, since a positive lookaround has to come back to it without unwinding anything; and a pass is numbered from a counter, so that no two runs of a sub-pattern share one where the frame depth that numbered them could.

### The limit is renamed, and the old name is refused

What the limit counted was C frames. What a search may hold now is the height of a stack of the engine's own, on the heap, so the name says so:

| was | is |
| --- | --- |
| `MRB_REGEXP_RECURSION_LIMIT` (1000) | `MRB_REGEXP_STACK_LIMIT` (2048) |
| `Regexp::RECURSION_LIMIT` | `Regexp::STACK_LIMIT` |
| `recursion limit over (MRB_REGEXP_RECURSION_LIMIT)` | `stack limit over (MRB_REGEXP_STACK_LIMIT)` |

The old macro is not aliased to the new one. The two count different things, so a value chosen for C frames does not carry over, and a build that had `-DMRB_REGEXP_RECURSION_LIMIT=500` for a reason has to choose a new number rather than have the header guess one for it. It is not ignored either, which would leave that build running on the default under a restriction it believes it still has: defining it is an `#error` naming the replacement.

### What the limit bounds

An entry is a choice point or an undo record, and `MRB_REGEXP_STACK_LIMIT` counts the two stacks together. What a repetition spends per iteration is what it holds: one choice point where it captures nothing, and undo records on top of that for a capture (up to two writes to open a group, since the end slot is cleared with the start, and one to close it) and for the record of an iteration whose body can match empty. A write of the value already in the slot is not one of them (see the next section). A subject longer than the limit still raises.

Counting live entries does not on its own bound the memory, since the arrays behind them grow geometrically and keep their capacity for the rest of the search: a search that fills one, backtracks, and then fills the other holds both high-water marks, and doubling alone would let each of them stand at up to twice the limit. So neither array is grown past the limit. At most that many entries stand in each, an entry being 32 and 16 bytes on a 64-bit ABI and 24 and 8 on a 32-bit one, which is 98,304 bytes together at the default on a 64-bit build and 65,536 on a 32-bit one, and halving the limit halves the ceiling. The capture slots and the per-instruction iteration records are sized by the pattern and lie outside it.

What a build may set the macro to is bounded where it is defined, at 1 and 16777216, and a build outside that range is an `#error` naming it. The range is the engine's rather than the test suite's: the count the limit is compared against is unsigned, so `-1` would stand for the largest ceiling there is rather than be refused; a capacity is doubled in 32 bits and multiplied by an entry's width to ask the allocator for bytes, which is 32 bits wide as well on a 32-bit ABI, so a value near the top of that range stops sizing anything; and at 0 no search could hold a single entry, which is a limit that has stopped being one.

A low limit inside that range is a build's to choose, and nothing here reads it as a mistake: what it buys is a smaller ceiling on what one search may ask the allocator for, at the price of the patterns the engine will match. An ordinary pattern holds a handful of entries whatever its subject, ten groups and a backreference holding twenty-two of them before the first repetition adds any, so a limit below about 48 answers `RegexpError` to patterns the tests take for granted. That is the engine doing what the build asked, and the tests say so rather than fail (see Testing).

The two limits are set apart from one another as well, which no `#error` can do. Filling the stack costs a handful of steps an entry, so a build that turns this one up far enough puts it out of the step limit's reach: a search that was to stop here stops there. Nothing in the engine reads that as an error, the two being answers to different questions, but the tests that pin the stack limit size their subjects from it and have nothing to pin on such a build (see Testing).

### The limit counts what has to be put back

`bt_log()` took an entry off the budget for every write it was handed, whether or not the write changed anything. `RE_SAVE` is where that showed: opening a group logs the start and then clears the end slot with it, so that a backreference reads a group the repetition has just re-entered as unmatched, and the end slot already holds `-1` wherever the group has not matched in this attempt yet. That second call logged `-1` over `-1`: a record `bt_undo_to()` walks to write `-1` back over `-1`, and an entry `bt_room()` counted on the way in. So the limit bounded the writes a search recorded rather than the state that differs, and two builds with the same limit held different amounts of restorable state depending on how many of their writes were no-ops.

A write of the value already in the slot is not logged now. Backtracking unwinds to a height, and a slot the log does not name is a slot no record has moved, so the value the unwind leaves is the one that stood there; a search at the limit goes on through such a write rather than stopping at one that would have restored nothing. `bt_iter_begin()` is covered by the same line, going through the same function.

The clear is a no-op only on the group's first iteration in an attempt, so a repetition's per-iteration cost is what it was and the run a pattern crosses moves by a character or two. What moves is the constant a fresh attempt pays, one entry per group, and a walk that tries a start position per byte pays it at every one of them: that is where the instruction count falls (see Speed).

### The default is where the move costs no pattern its subject

Taking the state off the C stack is one change and letting a search hold more of it is another, so the default is set where the first costs nothing and stops there: no pattern crosses a shorter run than it did on 1,000 C frames, and none crosses much more than it has to.

The old number does not carry over, a frame not being an entry. A fork was one frame and is one choice point; a capture was one frame and is up to three undo records. So the ratio differs by shape, and the default is read off the tightest of them. The longest run each pattern crosses whole:

| pattern | on 1,000 frames (master) | on 1,024 entries | on 2,048 entries (this PR) |
| --- | --- | --- | --- |
| `/(a)*?b/` | 498 | 340 | 682 |
| `/(a)*(b)\2/` | 332 | 255 | 511 |
| `/(?<x>a)*\k<x>/` | 331 | 254 | 510 |
| `/(?:(a)(b))*(c)\3/` | 399 | 293 | 585 |
| `/((a)*)*(b)\3/` | 330 | 253 | 509 |
| `/(?:(a)b?)*(c)\2/` | 332 | 255 | 511 |
| `/(?:a)*(b)\1/` | 996 | 1,020 | 2,044 |
| `/(?:(?>a))*/` | 332 | 1,021 | 2,045 |
| `/(?:(?=a)a)*/` | 332 | 1,021 | 2,045 |
| `/(?:a(?<=a))*/` | 332 | 1,022 | 2,046 |
| `/(?:(?>(a)))*(b)\2/` | 199 | 255 | 511 |

2,048 is the smallest power of two that clears every row: at 1,024 the six capture-heavy shapes cross less than they did on 1,000 C frames. A chain of atomic groups or of lookarounds, which spent two frames a link, now spends none once each has closed and is bounded by the pattern rather than by this limit either way.

### A refused allocation is not a limit

`mrb_realloc_simple()` is what grows the arrays, a raising allocator longjmping past the `mrb_free()` that ends a search. Reading its NULL as the stack limit would answer `stack limit over (MRB_REGEXP_STACK_LIMIT)` for an allocator that had nothing left, and what a build does about that message is turn the limit up, which is the worst thing it could do about the actual problem.

So the two travel apart. `bt_push()`, `bt_log()` and `bt_iter_begin()` answer `BT_OK`, `BT_LIMIT` or `BT_NOMEM` rather than a truth value, `bt_match()` hands the last two up unchanged the way it hands up a match, and `backtrack_exec()` turns `BT_NOMEM` into `RE_NOMEM` once its buffers are freed. `re_check_exec_error()` raises `NoMemoryError` on it, thrown from the object mruby keeps for that, so the raise itself asks for no memory. There is no automatic test for it. A refused allocation is not reachable from Ruby, and `mrb_open_allocf()` hands an allocator to a state at its birth, so a refusal injected that way is the whole VM's: reaching this engine's two growths and nothing else would take either a hook in the engine that only a test build compiles, or an allocator that guesses at request sizes and the order they arrive in, and neither is worth what it would pin. It was exercised by hand instead, on a build whose two `mrb_realloc_simple()` calls go through an injection that refuses either the first k growths or every one of them, and that can narrow the refusal to one of the two stacks. Neither stack begins with a capacity, so a refusal that stands is met by the first search that needs one:

- refusing every growth, a search that needs either stack raises `NoMemoryError: Out of memory`, which `rescue Exception` catches (it is a direct subclass, as in CRuby), and one that needs neither answers as it did;
- refusing the choice points alone, `/(?:(?>a))*z/` raises where `/(a)\1/` answers, the latter writing an undo record and pushing no choice point; refusing the undo log alone, both raise, a repetition logging where its iteration began;
- refusing the first growth alone, the search that meets it raises and every search after it answers, the buffers of the failed one having been freed and nothing of it left behind.

### The seven commits

Each builds and passes the whole suite on its own at the default limit, as C and as C++, so the migration can be read one mechanism at a time. What a build reads from a limit set low is the tip's answer rather than each step's: the tests that pin the limit size their subjects from it, and a mechanism that has not moved yet is bounded by the C stack instead, so those subjects reach no limit until the sixth commit has landed.

| commit | what it moves off the C stack |
| --- | --- |
| `back the backtracking engine's forks with a heap stack` | the unmarked `RE_SPLIT` / `RE_SPLITNG`, the two stacks, the limit |
| `log a capture's write instead of recursing over it` | `RE_SAVE` |
| `log where an iteration began instead of recursing over it` | `bt_iter()`, `entered_at` / `entered_in` |
| `cut an atomic group by truncating the choice point stack` | `RE_ATOMIC` / `RE_ATOMIC_END` |
| `cut a lookaround by truncating the choice point stack` | `bt_look()`, `RE_LOOK_END`, and with it `BT_CUT` |
| `drop what the frame-per-fork engine left behind` | `depth`, the transitional C-stack bound, the comments |
| `log a write only where there is something to take back` | nothing further; it settles what the limit counts once the state is on the heap |

While the migration is under way the mechanisms that still recurse keep a bound of their own (`RE_FRAME_LIMIT`), which the sixth commit removes. The seventh moves nothing further and is separable from the six: what it settles is what an entry is, which only has an answer once the state is on the heap.

## Behaviour

The run a repetition can cross no longer depends on what its body holds. On the default build, the longest subject each pattern matches whole:

| pattern | master | this PR | CRuby |
| --- | --- | --- | --- |
| `/(?:(?>a))*/` | 332 | 2,045 | bounded by `Regexp.timeout` |
| `/(?:(?=a)a)*/` | 332 | 2,045 | bounded by `Regexp.timeout` |
| `/(?:a)*(b)\1/` | 996 | 2,044 | bounded by `Regexp.timeout` |
| `/(a)*?b/` | 498 | 682 | bounded by `Regexp.timeout` |

Master spent three C frames an iteration on the first two, one on the third and two on the fourth; this PR spends one entry an iteration on the first three and three on the fourth, with `MRB_REGEXP_STACK_LIMIT` at 2,048. The fourth is the tightest of the shapes measured, and is where the default is read off (see Changes). So subjects that raised `RegexpError` on master now answer as CRuby does:

```ruby
s = "a" * 2000
s.match(/(?:(?>a))*/)[0].size              # 2000
s.match(/(?:(?=a)a)*/)[0].size             # 2000
s.match(/(?:(?>a))*\z/).begin(0)           # 0
(s + "bb").match(/(?:a)*(b)\1/).begin(0)   # 0

t = "a" * 600                              # a capture costs three entries an iteration
(t + "b").match(/(a)*?b/).begin(0)         # 0
```

Each of the five raised `recursion limit over (MRB_REGEXP_RECURSION_LIMIT)` on master, and each answers here what CRuby 4.0.6 answers. A chain of groups is bounded by the pattern rather than by the limit now, since a group that has closed holds nothing: `Regexp.new("(?>a)" * 1001 + "a")` and `Regexp.new("(?=a)" * 1001 + "a")` both raised on master and both match here.

A search the allocator refuses raises `NoMemoryError` rather than `RegexpError`, which is new: there was no allocation on this path to refuse before.

What a search costs the process falls with the frames. On the longest runs both builds cross, `/(?:(?>a))*/` over 330 characters and `/(?:a)*(b)\1/` over 990, peak RSS is 3,616 KB on master and 3,348 KB here, taken with address-space randomisation off as the median of 21 runs. 3,348 KB is what this build holds before the search, so its peak is the interpreter's own, where master's stands 256 KB above its. Massif, counting the C stack, puts those two searches at 285,104 bytes against 165,608 and 286,776 against 177,032: the stack a search touches goes from about 146,000 bytes to 7,208, and the two arrays hold 16,508 at their widest on the longer run.

## Speed

Callgrind, default configuration (`-O3`), `Ir(400 iterations) - Ir(200 iterations)` so that start-up cancels, per iteration:

```ruby
s480  = "hello world foo bar " * 24
words = (["alpha1", "beta", "gamma22", "delta", "eps3"] * 20).join(" ")   # 20 words

n.times { s480.gsub(/o/) { } }          # gsub480_blk, the Pike VM
n.times { words.scan(/\w+(?=\d)/) }     # scan_lookahead
n.times { words.scan(/(?<=a)\d+/) }     # scan_lookbehind
n.times { words.scan(/(?>\w+)\d/) }     # scan_atomic
n.times { words.scan(/(\w)\1/) }        # scan_backref
n.times { words.scan(/\w+?\d/) }        # scan_nongreedy
n.times { "aaaaaaaaaaaaaaaaaaaab".match(/(a+)+b/) }   # match_nested
```

| case | master | this PR | delta |
| --- | --- | --- | --- |
| gsub480_blk | 402,530 | 402,617 | +0.0% |
| scan_lookahead | 323,621 | 331,930 | +2.6% |
| scan_lookbehind | 242,033 | 232,842 | -3.8% |
| scan_atomic | 382,387 | 331,449 | -13.3% |
| scan_backref | 278,824 | 260,100 | -6.7% |
| scan_nongreedy | 240,498 | 258,411 | +7.4% |
| match_nested | 37,495 | 37,744 | +0.7% |

An atomic group is where the frames were most expensive, two per iteration, and a barrier push is one write; a non-greedy repetition is where they were cheapest, its iterations sharing a frame, so the heap push is what it pays now. What moves the three `scan` rows the other way is the last commit: a walk tries a start position per byte, and every attempt used to spend an entry per group on clearing an end slot that already held `-1`. `gsub480_blk` is the Pike VM, which this does not touch, and the 87 instructions between the two builds are the layout of code neither of them ran.

The last commit on its own, measured against the six before it:

| case | six commits | with the last | delta |
| --- | --- | --- | --- |
| gsub480_blk | 402,617 | 402,617 | +0.0% |
| scan_lookahead | 338,663 | 331,930 | -2.0% |
| scan_lookbehind | 246,911 | 232,842 | -5.7% |
| scan_atomic | 343,942 | 331,449 | -3.6% |
| scan_backref | 282,927 | 260,100 | -8.1% |
| scan_nongreedy | 264,184 | 258,411 | -2.2% |
| match_nested | 37,744 | 37,744 | +0.0% |
| a scan over sixteen groups | 860,163 | 761,956 | -11.4% |

`match_nested` is one search from one start position, so it pays the constant once and reads the same either way.

Wall clock, minimum of 15 alternating runs, same build and the cases of the review on #7267 alongside the seven above:

```ruby
s480  = "hello world foo bar " * 24
words = (["alpha1", "beta", "gamma22", "delta", "eps3"] * 200).join(" ")  # 200 words

30000.times  { s480.gsub(/o/) { } }        # gsub480_blk
100000.times { "abc".gsub(/b/) { } }       # gsub_abc_blk
100000.times { "abc".scan(/b/) { } }       # scan_abc_blk
100000.times { "abc".sub!(/b/) { } }       # sub!_abc_blk
30000.times  { s480.gsub(/o/, "0") }       # gsub480_str
100000.times { "abc".scan(/b/) }           # scan_abc
100000.times { "abc".sub!(/b/, "0") }      # sub!_abc_str
30000.times  { s480.gsub(/z/) { } }        # gsub480_none
2000.times   { words.scan(/\w+(?=\d)/) }   # scan_lookahead
2000.times   { words.scan(/(?<=a)\d+/) }   # scan_lookbehind
2000.times   { words.scan(/(?>\w+)\d/) }   # scan_atomic
2000.times   { words.scan(/(\w)\1/) }      # scan_backref
2000.times   { words.scan(/\w+?\d/) }      # scan_nongreedy
20000.times  { "aaaaaaaaaaaaaaaaaaaab".match(/(a+)+b/) }   # match_nested
```

| case | master | this PR | ratio |
| --- | --- | --- | --- |
| gsub480_blk | 754ms | 748ms | 0.99x |
| gsub_abc_blk | 132ms | 133ms | 1.01x |
| scan_abc_blk | 141ms | 144ms | 1.02x |
| sub!_abc_blk | 155ms | 156ms | 1.01x |
| gsub480_str | 110ms | 111ms | 1.01x |
| scan_abc | 100ms | 101ms | 1.01x |
| sub!_abc_str | 156ms | 157ms | 1.01x |
| gsub480_none | 34ms | 34ms | 1.00x |
| scan_lookahead | 321ms | 307ms | 0.96x |
| scan_lookbehind | 245ms | 227ms | 0.93x |
| scan_atomic | 367ms | 293ms | 0.80x |
| scan_backref | 292ms | 266ms | 0.91x |
| scan_nongreedy | 237ms | 245ms | 1.03x |
| match_nested | 48ms | 48ms | 1.00x |

Each figure is a whole process, so a few milliseconds of start-up stand in every row.

The two tables agree on the rows that move by more than a couple of percent: the atomic group and the backreference gain, the non-greedy repetition pays, and the eight Pike VM cases stand still. `scan_lookahead` is where they part, +2.6% of the instruction count against 0.96x of the clock, and 14ms on 321ms is what a shared machine moves by between runs. Those are the counts to read rather than the milliseconds beside them.

## Size

`.text` of `bin/mruby`, `size -A`:

| --- | --- | --- | --- |
| full-debug (-O0) | 1,905,478 | 1,906,150 | +672 |
| bintest | 1,301,926 | 1,304,310 | +2,384 |
| cxx_abi | 1,326,969 | 1,329,033 | +2,064 |
| byte-string | 1,267,446 | 1,269,734 | +2,288 |
| ascii-ctype | 1,288,262 | 1,290,198 | +1,936 |
| default | 1,214,430 | 1,216,718 | +2,288 |

All of it is the gem: in the default build `re_exec.o` `.text` goes 10,514 to 12,578 and `regexp.o` 26,933 to 27,157, which is that build's +2,288 exactly. `re_compile.o` (27,025) is unchanged. The last commit is 32 of those bytes back, a comparison being cheaper than the push it skips.

## Testing

`rake test` and `rake MRUBY_CONFIG=ci/gcc-clang test` pass on every one of the seven commits. The totals below are the tip:

| Build | Total | KO | Crash | Skip |
| full-debug | 2509 | 0 | 0 | 6 |
| bintest | 2509 (+124 bintest) | 0 | 0 | 14 |
| cxx_abi | 2509 | 0 | 0 | 14 |
| byte-string | 2435 | 0 | 0 | 54 |
| ascii-ctype | 2502 | 0 | 0 | 14 |
| default | 2279 (+113 bintest) | 0 | 0 | 54 |

`rake MRUBY_CONFIG=build_config/gcc-asan.rb test` and the clang counterpart pass on the tip as well, both `full-core` under `address,undefined`: 2509 assertions, no failure and no diagnostic from either sanitizer. The two stacks are grown with `mrb_realloc_simple()` and read through raw slot pointers and indices, so the arithmetic behind them is worth running under a sanitizer rather than only under the tests.

The `assert_raise(RegexpError)` assertions that pinned the limits either stay or become `assert_equal`; none goes the other way. The subjects behind them are re-sized once, in the first commit, since that is where the counting rule moves.

A block that pins the stack limit sizes its subjects from `Regexp::STACK_LIMIT`, and the chain of cuts it spells as well, so it asks two things of the build before it runs: that filling the stack cost fewer steps than `MRB_REGEXP_STEP_LIMIT` allows, and that a chain that long be a pattern one may spell. Where either fails it skips, the second by handing the chain to `Regexp.new` and reading the refusal rather than by writing the compiler's instruction count into a test. The step limit is pinned on its own, by `(?:a|a|a|a)+(z)\1`, which spends 4**m steps while holding about 3*m entries: that reads the same however low the stack limit stands, where the `(a+)+\1b` it replaces would fill the stack first and pin the wrong limit.

A test that reaches the backtracking engine without reading the limit asks for one of its own, `need_backtracking_stack` (see `test/backtracking_stack.rb`), and skips where the build stands below it. The engine refuses more patterns as the limit falls, so a build that set it low answers `RegexpError` to a backreference or a lookaround that these files take for granted, and 51 of them lose their subject that way. 48 is where every one matches again, counted down from there: 44 leaves one, 36 two, 1 fifty-one. The last commit is what brings that floor down from 49, an attempt no longer spending an entry per group on a write that restores nothing. The floor is the same on the boxes that run more of these files than `rake test` does: `full-core` and the byte-indexed box both leave one test at 47 and none at 48.

What asks for it is a test and not a file. An assertion that reaches no further than the parser, and one whose pattern the Pike VM runs, hold none of this state and answer the same whatever the limit is, so they stand in an assert of their own and a build with a low limit goes on running them; where a block held both kinds, it is split in two and the two names say which is which. That leaves the guard over 325 assertions rather than over the 640 a block-at-a-time reading of it would cover, and the split is checked the only way it can be: with `need_backtracking_stack` neutered, a build at 1 leaves exactly the 51 tests that carry it, and none of the rest.

There is no assertion for the accounting itself. What a build's limit buys is not readable from Ruby beyond `Regexp::STACK_LIMIT`, and the runs above move by less than a test sized from that constant can pin without writing down what an instruction costs. The floor coming down, and the suite staying green at every limit in the range, is what says the log still restores what it has to.

`rake test` was run at `MRB_REGEXP_STACK_LIMIT` of 1, 2, 8, 16, 32, 47, 48, 64, 1024, 2048, 125000, 125001 and 16777216, and is KO 0 and Crash 0 at every one: the two ends of the range, the default, the floor the tests ask for and the point below it, and the points on either side of where each skip begins. What moves is the skip count, 54 at the default and 105 below 48.

Each commit adds the tests for what it makes match, plus what its mechanism has to keep answering: captures surviving an atomic group's cut, a negative lookaround's captures not surviving, and a start position leaving no iteration record for the next one to read.

Differential runs against CRuby 4.0.6. Patterns are generated over the features this engine carries, each is run under CRuby, under master and under this branch, and what each answers is compared as `MatchData#to_a` inspected; a run whose address space or clock runs out is recorded rather than lost. Master standing beside the branch is what separates a difference this branch introduced from one the gem already had.

Over 10,000 patterns paired with a subject of at most four characters, CRuby answered 9,998: 9,967 of them both builds answer as CRuby does, 27 are cases both already answered differently, and 4 are cases both stop at a limit on. There is no case the two builds answer differently.

A subject of four characters reaches no limit either build sets, so a second run pairs 2,000 patterns with subjects that are runs of 100, 1,000 and 3,000 characters. CRuby answered 1,982: 1,843 both builds answer as CRuby does, 39 are cases master stops at its limit on and this branch answers as CRuby does, 6 are cases both already answered differently, and 94 are cases both stop at a limit on. Here too the two builds part on nothing, and nothing this branch refuses is something master answered: what a search gives up on goes from 148 cases to 109.

A corpus generated in the gem's own terms was run on the tip against master as well: five sets of 4,000 patterns over the 31 subjects over `ab` of length 4 or less, 187,670 lines counting the diagnostics of the patterns that do not compile, written identically by both.

## Environment

<details>
<summary>Machine, toolchain, and the compile line of every build</summary>

| Item | Value |
| --- | --- |
| OS | Ubuntu 24.04.4 LTS |
| Kernel | 7.0.0-30-generic |
| valgrind | valgrind-3.27.1 |
| CRuby (reference) | ruby 4.0.6 (2026-07-14 revision 03b6d3f889) +PRISM [x86_64-linux] |

Actual compile line of `mrbgems/mruby-regexp/src/re_exec.c` in each build (`-MMD -c`, `-I`, and `-o` dropped). `full-debug` is `-O0` because `enable_debug` appends `-g3 -O0` after the toolchain's `-g -O3`; `cxx_abi` compiles C as C++ with `gcc -x c++ -std=gnu++03`, g++ only links.

```console
# full-debug
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -g3 -O0 -DMRB_GC_STRESS -DMRB_USE_DEBUG_HOOK -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_DEBUG -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/re_exec.c
# bintest
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRB_GC_FIXED_ARENA -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER -DMRB_USE_DEBUG_HOOK mrbgems/mruby-regexp/src/re_exec.c
# cxx_abi
gcc -g -O3 -Wall -Wundef -Wwrite-strings -x c++ -std=gnu++03 -DMRB_GC_FIXED_ARENA -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_CXX_EXCEPTION -DMRB_USE_CXX_ABI -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/re_exec.c
# byte-string
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/re_exec.c
# ascii-ctype
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRB_USE_ASCII_CTYPE -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER mrbgems/mruby-regexp/src/re_exec.c
# default (no MRUBY_CONFIG), the build every figure above was measured on
gcc -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration -Wwrite-strings -DMRBGEM_MRUBY_REGEXP_VERSION=0.0.0 -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DMRB_USE_COMPLEX -DMRB_USE_BIGINT -DMRB_USE_DEBUG_HOOK mrbgems/mruby-regexp/src/re_exec.c
```

</details>

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -301,20 +301,22 @@ A search that reaches either limit raises `RegexpError`, `step limit over
 rather than answer with what it had found by then. The step limit bounds
 the work one search may do; the stack limit bounds the state it holds while
 doing it: the branches it has not taken yet and the writes it has not taken
-back, an entry each. What a repetition spends per iteration is what it
-holds: one choice point where it captures nothing, and undo records on top
-of that for a capture (two writes to open a group and one to close it) and
-for the record of an iteration that may match empty. A run longer than the
-limit reaches it on a pattern that is not pathological; a build with the
-memory for it can set it higher, and one that wants a smaller ceiling can
-set it lower.
+back, an entry each. A write of the value already in the slot leaves nothing
+to take back and is not counted. What a repetition spends per iteration is
+what it holds: one choice point where it captures nothing, and undo records
+on top of that for a capture (up to two writes to open a group and one to
+close it, an iteration that opens one the attempt has not closed yet paying
+for one of the two) and for the record of an iteration that may match empty.
+A run longer than the limit reaches it on a pattern that is not pathological;
+a build with the memory for it can set it higher, and one that wants a
+smaller ceiling can set it lower.
 
 The default is where the state moving off the C stack costs no pattern the
 subject it used to match, and no higher. The limit that preceded it counted
 C frames, and a frame is not an entry: a fork was one frame and is one
-choice point, while a capture was one frame and is three undo records. Of
-the shapes measured across that change the tightest is `(a)*?b`, which
-crossed 498 characters on the old 1,000 frames and crosses 681 on 2,048
+choice point, while a capture was one frame and is up to three undo records.
+Of the shapes measured across that change the tightest is `(a)*?b`, which
+crossed 498 characters on the old 1,000 frames and crosses 682 on 2,048
 entries; a chain of atomic groups or of lookarounds, which spent two frames
 a link and now spends none once each has closed, is bounded by the pattern
 rather than by this limit either way.
@@ -323,7 +325,7 @@ The limit stands between 1 and 16,777,216, and a build that sets it outside
 that fails to compile: at 0 no search could hold one entry, and above the
 ceiling the arithmetic that sizes the arrays stops holding on a 32-bit ABI.
 A low limit is a build's to choose, and what it buys is memory at the price
-of the patterns the engine will match: the gem's tests ask for 49, which is
+of the patterns the engine will match: the gem's tests ask for 48, which is
 where every pattern they take for granted matches, and the assertions that
 reach the engine skip below it while the rest go on running. The two limits
 are set apart from one another as well: a build that turns this one up far
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -176,11 +176,15 @@ ReDoS attacks.
 **Backtracking engine**: Used when patterns contain `\1`-`\9`
 backreferences, non-greedy quantifiers (`*?`, `+?`, `??`),
 lookaround assertions (`(?=...)`, `(?!...)`, `(?<=...)`, `(?<!...)`)
-or atomic groups (`(?>...)`). Bounded by a configurable step limit
-(`MRB_REGEXP_STEP_LIMIT`, default 1M) against excessive backtracking and
-by a recursion limit (`MRB_REGEXP_RECURSION_LIMIT`, default 1000) on
-the C stack. A search that reaches either raises `RegexpError` naming
-the limit, since what it had found by then is not the answer.
+or atomic groups (`(?>...)`). It backtracks on a stack of its own on the
+heap, so a search spends a constant amount of C stack however long the
+subject is. Bounded by a configurable step limit (`MRB_REGEXP_STEP_LIMIT`,
+default 1M) against excessive backtracking and by a stack limit
+(`MRB_REGEXP_STACK_LIMIT`, default 2048) on how tall that stack may
+stand. A search that reaches either raises `RegexpError` naming the limit,
+since what it had found by then is not the answer; one whose stack the
+allocator refuses to grow raises `NoMemoryError` instead, that being a
+different thing to do something about.
 
 The engine is selected automatically at compile time based on
 pattern analysis.
@@ -286,22 +290,71 @@ there.
 #define MRB_REGEXP_STEP_LIMIT 1000000
 #endif
 
-/* Maximum recursion depth of the backtracking engine (C stack) */
-#ifndef MRB_REGEXP_RECURSION_LIMIT
-#define MRB_REGEXP_RECURSION_LIMIT 1000
+/* Maximum height of the backtracking engine's stack (heap) */
+#ifndef MRB_REGEXP_STACK_LIMIT
+#define MRB_REGEXP_STACK_LIMIT 2048
 #endif
 ```
 
 A search that reaches either limit raises `RegexpError`, `step limit over
-(MRB_REGEXP_STEP_LIMIT)` or `recursion limit over
-(MRB_REGEXP_RECURSION_LIMIT)`, rather than answer with what it had found
-by then. The recursion limit is spent per fork and per capture, so a
-repetition of an atomic group or a lookaround spends a few frames per
-iteration, and a long enough run of one reaches it on a pattern that is
-not pathological; a build with the stack for it can set it higher. The
-values a build chose are `Regexp::RECURSION_LIMIT` and `Regexp::STEP_LIMIT`,
-for a program that has to size a subject or a pattern to the build it runs
-on; CRuby has no counterpart, its guard being `Regexp.timeout`.
+(MRB_REGEXP_STEP_LIMIT)` or `stack limit over (MRB_REGEXP_STACK_LIMIT)`,
+rather than answer with what it had found by then. The step limit bounds
+the work one search may do; the stack limit bounds the state it holds while
+doing it: the branches it has not taken yet and the writes it has not taken
+back, an entry each. What a repetition spends per iteration is what it
+holds: one choice point where it captures nothing, and undo records on top
+of that for a capture (two writes to open a group and one to close it) and
+for the record of an iteration that may match empty. A run longer than the
+limit reaches it on a pattern that is not pathological; a build with the
+memory for it can set it higher, and one that wants a smaller ceiling can
+set it lower.
+
+The default is where the state moving off the C stack costs no pattern the
+subject it used to match, and no higher. The limit that preceded it counted
+C frames, and a frame is not an entry: a fork was one frame and is one
+choice point, while a capture was one frame and is three undo records. Of
+the shapes measured across that change the tightest is `(a)*?b`, which
+crossed 498 characters on the old 1,000 frames and crosses 681 on 2,048
+entries; a chain of atomic groups or of lookarounds, which spent two frames
+a link and now spends none once each has closed, is bounded by the pattern
+rather than by this limit either way.
+
+The limit stands between 1 and 16,777,216, and a build that sets it outside
+that fails to compile: at 0 no search could hold one entry, and above the
+ceiling the arithmetic that sizes the arrays stops holding on a 32-bit ABI.
+A low limit is a build's to choose, and what it buys is memory at the price
+of the patterns the engine will match: the gem's tests ask for 49, which is
+where every pattern they take for granted matches, and the assertions that
+reach the engine skip below it while the rest go on running. The two limits
+are set apart from one another as well: a build that turns this one up far
+enough puts it out of the step limit's reach, since filling the stack costs
+a handful of steps an entry, and the tests that pin the stack limit size
+their subjects from it are skipped there. The values a build chose are
+`Regexp::STACK_LIMIT` and `Regexp::STEP_LIMIT`, for a program that has to
+size a subject or a pattern to the build it runs on; CRuby has no
+counterpart, its guard being `Regexp.timeout`.
+
+What the stack limit counts is live entries and not bytes. Two arrays hold
+them, one for the branches and one for the writes, and they grow
+geometrically and keep their capacity for the rest of the search, so a search
+that fills one, backtracks, and then fills the other holds both high-water
+marks at once. Neither is grown past the limit, so the memory one search may
+ask for is bounded by it: at most `MRB_REGEXP_STACK_LIMIT` entries in each
+array, an entry being 32 and 16 bytes respectively on a 64-bit ABI and 24 and
+8 on a 32-bit one, so 96 KiB together at the default on a 64-bit build and
+64 KiB on a 32-bit one. Halving the limit halves that ceiling. The capture
+slots and the per-instruction iteration records a search also holds are sized
+by the pattern rather than by this limit.
+
+A search whose stack the allocator refuses to grow raises `NoMemoryError`
+rather than `RegexpError`. The two are worth telling apart: a limit names the
+knob to turn, where turning `MRB_REGEXP_STACK_LIMIT` up in answer to an
+allocator that had nothing left would only let the next search ask for more.
+
+The macro was called `MRB_REGEXP_RECURSION_LIMIT` while the engine recursed
+once per fork and the limit counted C frames. A build that still defines that
+name fails to compile: the two limits count different things, so an old value
+does not carry over and the build has to choose a new one.
 
 Case folding beyond ASCII is not this gem's to configure. The table is
 core's, carried by any build that defines `MRB_UTF8_STRING` without
--- a/mrbgems/mruby-regexp/include/re_internal.h
+++ b/mrbgems/mruby-regexp/include/re_internal.h
@@ -173,19 +173,95 @@ typedef struct mrb_regexp_pattern {
 #define MRB_REGEXP_STEP_LIMIT 1000000
 #endif
 
-/* Recursion-depth limit for bt_match, which recurses at every fork and
-   every capture: bounds C stack growth on a long subject or a deep pattern. */
-#ifndef MRB_REGEXP_RECURSION_LIMIT
-#define MRB_REGEXP_RECURSION_LIMIT 1000
+/* How tall the backtracking engine's stack may stand in one search: the
+   choice points it has not tried yet and the writes it has not taken back,
+   counted together (see bt_room() in re_exec.c). That stack is on the heap,
+   so a search spends a constant amount of C stack however long the subject
+   is, and what this bounds is what it holds instead. Where
+   MRB_REGEXP_STEP_LIMIT bounds the work one search may do, this bounds the
+   state it may hold while doing it.
+
+   What it counts is live entries, not bytes. The two arrays behind them grow
+   geometrically and keep their capacity for the rest of the search, so a
+   search that fills one, backtracks, and then fills the other holds both
+   high-water marks at once. Neither array is grown past this limit, though
+   (see bt_push()), so the memory one search may ask for is bounded by it:
+   at most this many entries in each array, an entry being 32 and 16 bytes
+   on a 64-bit ABI and 24 and 8 on a 32-bit one, so 96 KiB together at the
+   default on a 64-bit build and 64 KiB on a 32-bit one. The capture slots
+   and the iteration records (see backtrack_exec()) are sized by the pattern
+   and lie outside it.
+
+   The default is where the state moving off the C stack costs no pattern the
+   subject it used to match, and no higher: moving it is one change and
+   letting a search hold more of it is another, and a default above this one
+   would make the second silently. The old limit allowed 1,000 C frames, and
+   a frame is not an entry: a fork was one frame and is one choice point,
+   while a capture was one frame and is three undo records, so what a pattern
+   spends per iteration is what it holds. `(a)*?b` crossed 498 characters on
+   1,000 frames and crosses 681 on 2,048 entries, which is the tightest of
+   the shapes measured; a chain of atomic groups or of lookarounds, which
+   spent two frames a link and now spends none once each has closed, is
+   bounded by the pattern rather than by this limit either way. A build that
+   wants a longer subject to match, or a smaller ceiling on the memory a
+   search may ask for, sets it. */
+#ifdef MRB_REGEXP_RECURSION_LIMIT
+/* The engine no longer recurses per fork, so nothing counts C frames any more
+   and this knob is gone. Its replacement counts entries on a heap stack, and
+   the two measure different things: a value chosen for the old one does not
+   carry over, and a build that means to keep a restriction has to choose a
+   new one rather than have this header guess. */
+#error MRB_REGEXP_RECURSION_LIMIT was replaced by MRB_REGEXP_STACK_LIMIT
+#endif
+#ifndef MRB_REGEXP_STACK_LIMIT
+#define MRB_REGEXP_STACK_LIMIT 2048
 #endif
 
-/* What a search answers when the backtracking engine gave up at one of the
-   two limits before it had an answer (see mrb_re_exec()). The caller raises
-   on it: what the search had found by then is not a shorter or a later
-   match, and reading it as one was the defect. Which of the two it was
-   names the knob to turn. */
-#define RE_OVER_RECURSION_LIMIT (-1)
+/* What a build may set it to. Outside this it bounds nothing.
+
+   The floor is what the engine itself asks: at 0 no search could hold a
+   single entry and every pattern that reaches this engine would answer
+   RegexpError, which is a limit that has stopped being one. Any value from 1
+   up is a build's to choose. A low one is not a broken build but a smaller
+   ceiling on what one search may ask the allocator for, bought by refusing
+   more patterns: an ordinary one holds a handful of entries whatever the
+   subject (ten groups and a backreference hold thirty-three before the first
+   repetition adds any), so a build that sets the limit that low is choosing
+   memory over the patterns it can match. The gem's own tests ask for 49,
+   which is where every pattern they take for granted matches again, and skip
+   below it (see test/backtracking_stack.rb).
+
+   The two limits are set apart from one another as well. Filling the stack
+   costs a handful of steps an entry, so a build that turns this one up far
+   enough puts it out of MRB_REGEXP_STEP_LIMIT's reach: a search that was to
+   stop here stops there instead. Nothing in the engine reads that as an
+   error, the two limits being answers to different questions, but the tests
+   that pin this one size their subjects from it and are skipped there.
+
+   Above the ceiling the arithmetic that sizes the arrays stops holding: a
+   capacity is doubled in 32 bits and multiplied by an entry's width to ask
+   the allocator for bytes, which is 32 bits wide too on a 32-bit ABI, and a
+   limit this high has in any case stopped being one, the two arrays at it
+   standing at hundreds of megabytes.
+
+   A negative value is refused here rather than read as no limit at all: the
+   count it is compared against is unsigned (see bt_room() in re_exec.c), so
+   -1 would stand for the largest ceiling there is. */
+#if MRB_REGEXP_STACK_LIMIT < 1 || MRB_REGEXP_STACK_LIMIT > (1 << 24)
+#error MRB_REGEXP_STACK_LIMIT must stand between 1 and 16777216
+#endif
+
+/* What a search answers when the backtracking engine stopped before it had
+   an answer (see mrb_re_exec()). The caller raises on it: what the search had
+   found by then is not a shorter or a later match, and reading it as one was
+   the defect. Which one it was names what to do about it: a limit names the
+   knob to turn, where RE_NOMEM says the allocator refused and no knob will
+   help. Turning MRB_REGEXP_STACK_LIMIT up in answer to a refused allocation
+   would make the memory it failed to find scarcer still, so the two are kept
+   apart all the way out of the engine. */
+#define RE_OVER_STACK_LIMIT (-1)
 #define RE_OVER_STEP_LIMIT (-2)
+#define RE_NOMEM (-3)
 
 /* Maximum captures */
 #define RE_MAX_CAPTURES 32
--- a/mrbgems/mruby-regexp/include/re_internal.h
+++ b/mrbgems/mruby-regexp/include/re_internal.h
@@ -197,10 +197,10 @@ typedef struct mrb_regexp_pattern {
    letting a search hold more of it is another, and a default above this one
    would make the second silently. The old limit allowed 1,000 C frames, and
    a frame is not an entry: a fork was one frame and is one choice point,
-   while a capture was one frame and is three undo records, so what a pattern
-   spends per iteration is what it holds. `(a)*?b` crossed 498 characters on
-   1,000 frames and crosses 681 on 2,048 entries, which is the tightest of
-   the shapes measured; a chain of atomic groups or of lookarounds, which
+   while a capture was one frame and is up to three undo records, so what a
+   pattern spends per iteration is what it holds. `(a)*?b` crossed 498
+   characters on 1,000 frames and crosses 682 on 2,048 entries, the tightest
+   of the shapes measured; a chain of atomic groups or of lookarounds, which
    spent two frames a link and now spends none once each has closed, is
    bounded by the pattern rather than by this limit either way. A build that
    wants a longer subject to match, or a smaller ceiling on the memory a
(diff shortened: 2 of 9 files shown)


matz:
Thank you for this. I read it, merged it onto master locally, and measured it; the work is sound and the numbers hold up. It needs a rebase before I can take it, and the resolution is not purely textual, so I want to say what I found.

## The conflict

The branch is based on 06ed794af. Since then 5ffcc0034 made `\b` and `\B` read the set `[[:word:]]` holds rather than the ASCII set `\w` names, so that a boundary sits beside a character of any script. That touched the same two `RE_WBOUND` / `RE_NWBOUND` cases the backtracker rewrite moves.

git reports two conflicting regions in `src/re_exec.c`, three lines each. The Pike VM side merges clean. Resolving means taking master's calls and keeping your control flow:

```c
    case RE_WBOUND:
      {
        mrb_bool before = (sp > str) && mrb_re_word_before(str, sp, str_end, binary);
        mrb_bool after = (sp < str_end) && mrb_re_word_at(sp, str_end, binary);
        if (before == after) goto fail;   /* was: return BT_FAIL */
      }
```

`mrb_re_word_before()` and `mrb_re_word_at()` are new inlines in `include/re_internal.h`. Both take `binary`, which the rewritten `bt_match()` already has in scope. The subject's start pointer is what `mrb_re_word_before()` needs to step back over a character, and that is `str` in the backtracker and `s->str` in the Pike VM.

## What I measured on the resolved tree

Correctness first, since an engine rewrite is where I most want evidence that nothing moved. 326 patterns that reach the backtracking engine (backreferences, atomic groups, lookbehind, nested quantifiers) against 19 subjects, comparing master, this branch and CRuby:

```
master vs CRuby differ : 7
branch vs CRuby differ : 7
master vs branch differ: 0
```

No difference introduced, none removed. The seven are the documented lookbehind limitations and my own harness's `Regexp.timeout`.

The longest-run table reproduces: `(a)*?b` 682, `(?:(?>a))*` 2,045, `(?:(?=a)a)*` 2,045, `(?:a(?<=a))*` 2,046.

Peak RSS, on a 300-character run both builds can do, is 4,608 KB on master and 4,288 KB here: a deep C recursion touched more pages than the heap stacks do. That is the number I care most about.

Wall clock at -O3, alternating runs, minimum of five: scan_atomic -15.6%, scan_lookbehind -5.3%, scan_backref -4.5%, scan_lookahead 0.0%, scan_nongreedy +5.0%, gsub480_blk and match_nested within noise. Same signs and the same order of magnitude as your Callgrind table. (I first measured on a debug build and read the scan rows as 25-36% slower, which was my mistake: -O0 charges far too much for the extra calls and the heap indirection.)

Whole suite green on host-debug, the six CI configurations, and clang-asan with UBSan, with 2,509 tests where master has 2,480.

## One thing I want to settle before merging

`Regexp::RECURSION_LIMIT` goes and `MRB_REGEXP_RECURSION_LIMIT` becomes an `#error` rather than an alias. I follow the reasoning, and I think refusing the old macro is right: a number chosen for C frames means nothing as a count of entries, and silently running on the default would be worse than not building. I am asking myself the same question about the constant, which is reachable from Ruby and not only from a build. I will decide that when I take the rebase.

takumin:
Thank you for reading it that closely, and for measuring it yourself.

Rebased onto `1c3e2b761` and force-pushed. The resolution is the one you wrote, in both cases and to the character: master's `mrb_re_word_before()` and `mrb_re_word_at()`, with `goto fail` where the old code returned `BT_FAIL`. The Pike VM side merged clean and `binary` was already in scope, as you say. `git range-diff` against what stood here before shows those two regions and nothing else; the other six commits went across untouched.

Everything in the body is measured again on the rebased tree. What moved:

- The suite is 2,509 assertions where it was 2,507, so every total in the Testing table rises by two, and the skip counts by one: 54 at the default, 105 below the floor.
- `.text`: cxx_abi +2,064, byte-string +2,288, ascii-ctype +1,936, default +2,288. full-debug (+672) and bintest (+2,384) are unchanged.
- The last commit gives back 32 bytes rather than 48. The new word boundary inlines are what changed the code around it.
- `RE_TESTS_NEED_STACK` still asks for 48, on the default box, on `full-core` and on the byte-indexed one; with the guard neutered, 47 leaves one test, 44 one, 36 two and 1 fifty-one, as before. The sweep over 13 limits from 1 to 16,777,216 is KO 0 and Crash 0 at every one.
- Callgrind and the clock stand where they did: scan_atomic -13.3%, scan_backref -6.7%, scan_lookbehind -3.8%, scan_nongreedy +7.4%.

Your correctness result reproduces here on a different corpus. Random patterns over the features this engine carries, CRuby 4.0.6 as the reference, master and this branch run over the same cases. With subjects of at most four characters, 10,000 patterns: of the 9,998 CRuby answered, 9,967 both builds answer as CRuby does, 27 are standing differences, 4 are cases both stop at a limit on, and there is no case where the two builds part. A four character subject reaches no limit, so a second run pairs 2,000 patterns with runs of 100, 1,000 and 3,000 characters: of the 1,982 answered, 1,843 agree in both, 39 are cases master stops at its limit on and this branch answers as CRuby does, 6 are standing differences and 94 are cases both stop on. Nothing where the builds part, and nothing this branch refuses that master answered; what a search gives up on goes from 148 cases to 109.

The peak RSS is in the body now, since you are right that it is the figure worth having. On the longest runs both builds cross, `/(?:(?>a))*/` over 330 characters and `/(?:a)*(b)\1/` over 990, it is 3,616 KB on master and 3,348 KB here, with address space randomisation off and read as the median of 21 runs. The minimum is not usable: page reclaim drops single runs below the baseline. 3,348 KB is what the build holds before the search, so this branch's peak is the interpreter's own, where master's stands 256 KB above its. Massif with the C stack counted says it without the page granularity: 285,104 bytes against 165,608 on the first, and 286,776 against 177,032 on the second, the stack a search touches going from about 146,000 bytes to 7,208 while the two arrays hold 16,508 at their widest.

On the constant, it is yours to decide and I will follow it. What I had in mind is the reasoning you give for the macro: a number of C frames read as a count of entries is a wrong answer rather than a stale one, and code that sizes a subject from `Regexp::RECURSION_LIMIT` would go on running with it. Keeping the old name as an alias for the new one is a line in the last commit if you would rather it stayed.

PR #7372, opened 2026-08-26
mruby-array-ext: write `uniq!`, `|` and `&` once against `ary_memb`, with the threshold at 8

takumin:
## Summary

Review of the first version of this PR measured below its table's first row and found the walk winning there, at a size the set cannot reach without paying an allocation that is constant in `n`. So the walk stays. What goes is the duplication: `uniq`/`uniq!`, `|` and `&` are now written once against `ary_memb`, the membership object #7373 introduced for `-` and `intersect?`, and the strategy choice moves out of the operations and into it. With one writing per operation the threshold is free to sit where the crossover is, and `SET_OP_HASH_THRESHOLD` moves from 32 to 8.

## Changes

| Commit | What |
|---|---|
| `mruby-array-ext: write uniq!, \| and & once against ary_memb` | two new question forms on the membership object; the three operations move onto it; no behavioural change for a contract-honouring type, threshold untouched |
| `mruby-array-ext: lower SET_OP_HASH_THRESHOLD from 32 to 8` | the number moves; every predicate that reads it keeps its shape |

### Two more question forms on the membership object

`ary_memb_has()` answers static membership, which is the question `-` and `intersect?` ask. These three ask two others, so the object learns them rather than each operation writing both strategies again:

- `ary_memb_first()` asks "is this the first time I meet this value?". The set remembers every value it is asked about; the walk remembers nothing and reads the caller's record of what it already kept, bounded by the caller's own count so `uniq!` can ask it about the array it is still compacting. `uniq!` asks it about the kept prefix, `|` about the result it is building. `add_uniq()`, the dedup `|` had written a second time, was this question and goes.
- `ary_memb_take()` is membership that hands each distinct element out at most once: asking spends it, and a duplicate no longer finds it. The set spends by deleting the key; the walk consults the caller's record of what was already taken. `ary_memb_select()` narrows it to one of the init arrays, which is the shape of `&`: each argument narrows the result in turn.

Three shapes move underneath, none observable for a type whose `hash` agrees with its `eql?`. The walk under `&` becomes argument-major, the same passes the set makes, instead of asking every argument about each element. The set under `uniq!` records first occurrences as it walks rather than pre-filling from the receiver and taking elements back out, which drops a full pass. The set under `|` stops copying its sources, since every key it remembers was pushed to the result in the same turn and the result is what keeps it alive; `|` now hands the membership object no arrays at all.

### The threshold moves to 8

Driving the deciding length directly against a build that always walks and one that always builds the set, the walk stops winning at 6 elements for `uniq`, 4 for `|` and 8 for `&`, and the set is ahead from 8, 6 and 12. 8 takes the whole of the set's win and is above `&`'s crossover, which 4 would not be. Below it the walk keeps the two things it is there for: at 2 elements it is 1.25x to 1.74x faster, and it allocates nothing where the set pays its `malloc` calls and table bytes at any size.

## Behaviour

For a type whose `hash` agrees with its `eql?` every answer is unchanged at every length, which the threshold block in the tests pins on both sides. A type that breaks that contract has no answer defined by Ruby, and reads back whichever strategy happened to answer; the lengths whose strategy changed are the deciding lengths 9 through 32, which now read the set's answer rather than the walk's.

## Speed

Instruction counts per call, callgrind `Ir(4,000 calls) - Ir(2,000 calls)` over 2,000, `build_config/default.rb`, master → this PR:

| deciding length | `uniq` | `&` | `\|` |
| --- | ---: | ---: | ---: |
| 2 | 1,525 → 1,567 (1.03x) | 2,440 → 2,523 (1.03x) | 1,556 → 1,618 (1.04x) |
| 4 | 2,961 → 3,020 (1.02x) | 6,140 → 6,234 (1.02x) | 3,566 → 3,640 (1.02x) |
| 8 | 7,590 → 7,683 (1.01x) | 21,886 → 21,992 (1.00x) | 11,687 → 11,815 (1.01x) |
| 16 | 27,496 → **9,987 (0.36x)** | 84,765 → **35,045 (0.41x)** | 42,917 → **15,928 (0.37x)** |
| 32 | 109,598 → **21,340 (0.19x)** | 337,305 → **64,346 (0.19x)** | 168,098 → **39,939 (0.24x)** |

At and below 8 both sides run the same walk, and the 1 to 4 percent there is the framing of the membership object. Above 8 the walk stops being asked, which is the whole win the first version of this PR measured, now without giving the small sizes away.

Wall clock agrees, best of 7 interleaved runs, 100,000 calls:

| benchmark | master | this PR |
| --- | ---: | ---: |
| `uniq`, 16 elements | 160 ms | **50 ms (0.31x)** |
| `uniq`, 32 elements | 650 ms | **120 ms (0.18x)** |
| `&`, 16-element argument | 510 ms | **190 ms (0.37x)** |
| `&`, 32-element argument | 2,020 ms | **360 ms (0.18x)** |
| `\|`, 8 with 8 elements | 260 ms | **90 ms (0.35x)** |
| `\|`, 16 with 16 elements | 1,000 ms | **230 ms (0.23x)** |

```ruby
a = (1..N/2).to_a * 2;                   i = 0; while i < 100000 do a.uniq; i += 1 end
a = (1..N).to_a;   b = (N/2+1..N+N/2).to_a
                                         i = 0; while i < 100000 do a & b;  i += 1 end
a = (1..N/2).to_a; b = (N/4+1..N/2+N/4).to_a
                                         i = 0; while i < 100000 do a | b;  i += 1 end
```

## Size

`.text` of `bin/mruby`, `size -A`, for the five builds of `build_config/ci/gcc-clang.rb`, master and this branch built at the same path in an empty build directory:

| --- | ---: | ---: | ---: |
| `ascii-ctype` | 1,071,686 | 1,070,646 | -1,040 |
| `bintest` | 1,076,822 | 1,075,782 | -1,040 |
| `byte-string` | 1,051,670 | 1,050,630 | -1,040 |
| `cxx_abi` | 1,064,917 | 1,063,877 | -1,040 |
| `full-debug` (-O0) | 1,867,046 | 1,864,214 | -2,832 |

All of it is `mrbgems/mruby-array-ext/src/array.o`, whose `.text` goes from 19,063 to 18,023, exactly the -1,040 of the four optimised rows. No other object moves.

## Testing

`rake test` is green on `build_config/default.rb` (2,367 tests, 0 KO, 0 crash, 0 warning) and across the five builds of `build_config/ci/gcc-clang.rb`, and `rake test:bin` passes its 145 bintest cases with 0 KO. Both commits were run separately; each is green on its own.

No test row is added or removed. The threshold block already asks each operation on both sides of the threshold and every answer survives; its comment now says 8. The blocks that quoted "36 elements > 32, triggers hash approach" and the like now say "past the hash threshold", so the next move of the number does not strand them. One row moved sides rather than being stranded: the below-threshold twin of the three-argument `intersection` narrowing put 10 elements into its arguments, which 8 reads as the hash side, so its arguments shrink to a total of 8 and it keeps asking the walk it was written for.

## Environment

<details>

| | |
|---|---|
| OS | Ubuntu 24.04, Linux 7.0.0-30-generic x86_64 |
| CPU | AMD Ryzen 9 5950X |
| C compiler | Homebrew clang 22.1.8 (`CC`, `CXX` and `LD` all clang) |
| valgrind | 3.27.1 (callgrind) |

Benchmarks were run one core at a time under `taskset -c 7`. Wall clock has 10 ms of resolution here, so the instruction counts are the ones to read for small differences; they are `Ir(2N) - Ir(N)` divided by `N`, which cancels out start-up and parsing.

The compile line for the benchmarked build, with `-MMD -c`, `-ffile-prefix-map`, `-I` and `-o` dropped:

```console
clang -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration
      -Wwrite-strings -Wzero-length-array -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0
      -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL
      -DMRB_USE_COMPLEX -DMRB_USE_BIGINT -DMRB_USE_DEBUG_HOOK
```

And for the five builds the `Size` table comes from:

```console
# ascii-ctype
clang -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration
      -Wwrite-strings -Wzero-length-array -DMRB_USE_ASCII_CTYPE
      -DMRB_REGEXP_PARSE_DEPTH_LIMIT=512 -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0
      -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING
      -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET
      -DMRB_USE_TASK_SCHEDULER

# bintest
clang -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration
      -Wwrite-strings -Wzero-length-array -DMRB_GC_FIXED_ARENA
      -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0 -DMRB_USE_BIGINT -DMRB_USE_COMPLEX
      -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM
      -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET
      -DMRB_USE_TASK_SCHEDULER -DMRB_USE_DEBUG_HOOK

# byte-string
clang -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration
      -Wwrite-strings -Wzero-length-array -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0
      -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL
      -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET -DMRB_USE_TASK_SCHEDULER

# cxx_abi, the one build compiled as C++
clang -g -O3 -Wall -Wundef -Wwrite-strings -Wzero-length-array -x c++ -std=gnu++03
      -DMRB_GC_FIXED_ARENA -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0
      -DMRB_USE_CXX_EXCEPTION -DMRB_USE_CXX_ABI -DMRB_USE_BIGINT -DMRB_USE_COMPLEX
      -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING -DHAVE_MRUBY_IO_GEM
      -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET
      -DMRB_USE_TASK_SCHEDULER

# full-debug, the only -O0 build
clang -std=gnu99 -g -O3 -Wall -Wundef -Werror-implicit-function-declaration
      -Wwrite-strings -Wzero-length-array -g3 -O0 -DMRB_GC_STRESS
      -DMRB_USE_DEBUG_HOOK -DMRBGEM_MRUBY_ARRAY_EXT_VERSION=0.0.0 -DMRB_DEBUG
      -DMRB_USE_BIGINT -DMRB_USE_COMPLEX -DHAVE_MRUBY_ENCODING_GEM -DMRB_UTF8_STRING
      -DHAVE_MRUBY_IO_GEM -DMRB_USE_RATIONAL -DHAVE_MRUBY_REGEXP_GEM -DMRB_USE_SET
      -DMRB_USE_TASK_SCHEDULER

# full-debug puts -g3 -O0 after the toolchain's own -g -O3, so it is the one
# row measured at -O0
```

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency across Array set operations, including union, intersection, difference, and uniqueness handling.
  * Array comparisons now consistently follow `eql?` semantics, regardless of array size.
  * Corrected difference results when values appear across multiple input arrays.
  * Ensured intersections include only elements present in every input array.
  * Ensured block-based uniqueness and related set operations produce consistent results.

### DIFF
--- a/mrbgems/mruby-array-ext/src/array.c
+++ b/mrbgems/mruby-array-ext/src/array.c
@@ -412,7 +412,13 @@ ary_rotate_bang(mrb_state *mrb, mrb_value self)
   return self;
 }
 
-#define SET_OP_HASH_THRESHOLD 32
+/* Above this many deciding elements the khash set repays building it; at and
+   below, the linear walk wins and allocates nothing. Driving the deciding
+   length directly against a build that always walks and one that always
+   builds the set, the walk stops winning at 6 elements for `uniq`, 4 for `|`
+   and 8 for `&`, and the set is ahead from 8, 6 and 12; 8 takes the whole of
+   the set's win without giving any operation's small sizes away. */
+#define SET_OP_HASH_THRESHOLD 8
 
 /* Helper functions for temporary khash sets */
 static void
--- a/mrbgems/mruby-array-ext/src/array.c
+++ b/mrbgems/mruby-array-ext/src/array.c
@@ -461,18 +461,23 @@ ary_get_array_args(mrb_state *mrb, mrb_int argc, const mrb_value **argv_ptr)
   return total_len;
 }
 
-/* Answers one question, "is this value one of the elements I hold?", and
-   hides which of two strategies answered it. A khash set when there are
-   enough elements to repay building one, a walk over the arrays themselves
-   when there are not.
+/* Remembers elements and answers questions about them, hiding which of two
+   strategies did the remembering. A khash set when there are enough elements
+   to repay building one, a linear scan of an array the caller can already
+   see when there are not.
 
    Choosing between those two is what makes this file fast and is worth
    keeping. Writing the surrounding operation twice, once per strategy, is
    what made it wrong: the two copies compared elements with different
    functions until #7352, and `Array#intersection` narrowed by the union of
    its arguments on one side and by each argument on the other until #7368.
-   A caller states its algorithm once, against ary_memb_has(), and never
-   learns which strategy replied. */
+   A caller states its algorithm once, against one of three questions, and
+   never learns which strategy replied:
+
+     ary_memb_has()    is this value one of the elements I hold?
+     ary_memb_take()   the same, but each distinct element is handed out at
+                       most once; asking spends it
+     ary_memb_first()  is this the first time the caller meets this value? */
 typedef struct ary_memb {
   mrb_bool use_set;
   ary_set_t set;
@@ -489,21 +494,25 @@ ary_memb_init(mrb_state *mrb, ary_memb *m, const mrb_value *arys, mrb_int n_arys
   m->use_set = use_set;
   m->n_arys = n_arys;
   m->total_len = total_len;
+  m->arys = arys;
 
   if (!use_set) {
     /* The walk reads the caller's arrays in place. It allocates nothing, which
        is most of why it is worth choosing at all. */
-    m->arys = arys;
     return;
   }
 
-  /* Shared copies hold the elements while kh_put() runs `hash`, which is Ruby
-     code and free to empty the caller's arrays. */
-  mrb_value *copies = (mrb_value *)mrb_alloca(mrb, sizeof(mrb_value) * n_arys);
-  for (mrb_int i = 0; i < n_arys; i++) {
-    copies[i] = mrb_ary_make_shared_copy(mrb, arys[i]);
+  if (n_arys > 0) {
+    /* Shared copies hold the elements while kh_put() runs `hash`, which is
+       Ruby code and free to empty the caller's arrays. A caller that hands
+       over no arrays has to pin the set's keys some other way; `|` holds
+       every key in its result. */
+    mrb_value *copies = (mrb_value *)mrb_alloca(mrb, sizeof(mrb_value) * n_arys);
+    for (mrb_int i = 0; i < n_arys; i++) {
+      copies[i] = mrb_ary_make_shared_copy(mrb, arys[i]);
+    }
+    m->arys = copies;
   }
-  m->arys = copies;
   ary_init_temp_set(mrb, &m->set, total_len);
 }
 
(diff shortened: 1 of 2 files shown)


matz:
Thank you for measuring this. The central claim holds: the walk does lose long
before 32, and I would like the threshold moved. What I would rather not do is
delete the walk, because measuring below your table's first row shows the set
losing there, and it costs an allocation at every size.

**Below n=8.** Your table starts at 8. Instruction counts (callgrind, `-O2`,
30,000 calls) for the sizes under it:

| op | n=2 | n=4 | n=8 | n=16 |
| `uniq` | **1.55x** | 0.72x | 0.38x | 0.19x |
| `\|` | **1.25x** | 0.58x | 0.30x | — |
| `&` | **1.74x** | 0.67x | 0.43x | 0.24x |

**Allocation.** The set costs 3 more `malloc`/`free` pairs and 259 more bytes
per call, and that is constant in n rather than proportional to it, so a
two-element `uniq` pays the same table as a two-thousand-element one. Nothing
is retained, but mruby's order is memory before speed, and this is the size of
array that a program runs through these operations most often.

**Where the crossover is.** I drove `total_len` directly against a build that
always walks and one that always builds the set:

| op | walk wins | set wins |
| --- | --- | --- |
| `uniq` | 6 (1.04x) | 8 (0.72x) |
| `\|` | 4 (1.88x) | 6 (0.67x) |
| `&` | 8 (1.47x) | 12 (0.54x) |

**So: keep both paths and make `SET_OP_HASH_THRESHOLD` 8 rather than 32.**
That takes the whole win your table shows, leaves the small sizes where they
are, and is above `&`'s crossover, which 4 would not be.

On the duplication: you are right that #7352 and #7368 were both two copies
disagreeing, and that is worth removing. I would rather remove it the way
#7373 does, by writing the operation once over a strategy, than by removing
the strategy. If the three can be written that way too, the walk stops being a
second place for the meaning to live without stopping being available.

takumin:
Thank you for the review, and for measuring the sizes my table left out; the crossover numbers below n=8 settled this better than anything I had. Reworked as suggested, in two commits.

The first writes `uniq`/`uniq!`, `|` and `&` once against `ary_memb`, keeping both strategies: `ary_memb_first()` ("is this the first time I meet this value?") carries `uniq!` and `|`, and `ary_memb_take()` with `ary_memb_select()` carries `&`'s per-argument narrowing, so the walk stops being a second place for the meaning to live without stopping being available. The second moves `SET_OP_HASH_THRESHOLD` from 32 to 8.

Re-measured against master: at and below 8 both sides run the same walk, within 1 to 4 percent; at 16 the set answers at 0.36x to 0.41x, and at 32 at 0.19x to 0.24x. The threshold's constant allocation is also down one on each side of it: `|` no longer copies its sources on the set path, since every key the set holds is pinned by the result it was pushed to. The PR body has the updated tables.

PR #7415, opened 2026-08-28
mruby-regexp: state the README as the answers a reader comes for

takumin:
## Summary

`mrbgems/mruby-regexp/README.md` had grown to 616 lines, and it had grown in one
direction: every rule was stated together with the reasoning that had chosen it,
the measurements that placed a default, and what an earlier spelling was called.
That reasoning is already written down, in the commit that made each change,
which is where a reader who wants it goes.

What the README is for is answering a question a reader arrives with, so this
states it as lists that carry the answer and nothing beside it, and fills the
gaps that reading it that way made visible: thirteen methods the gem defines
that no list named, three spellings the parser takes that no list named, and the
optional arguments the match methods take.

442 lines, and no rule is dropped.

## Changes

| Commit | What it does |
| --- | --- |
| `state each syntax entry on one line` | An entry is a spelling and what it stands for; what needs more goes below the list |
| `move the named capture rules in with the syntax` | Numbering is part of what a pattern means, so it sits with the syntax |
| `state each limitation as the answer it gives` | An entry names what a pattern raises here and how CRuby answers it |
| `split the configuration section by what it configures` | A heading each for the search limits, the parse depth limit and what the build decides |
| `say what the differential test asks, not why` | What the corpus covers, what the baseline holds, how to take a new one |
| `name the three spellings the syntax lists missed` | `\h` / `\H`, a `?` after `{n,m}`, `\b` as a backspace inside a class |
| `list the Regexp and MatchData methods the gem defines` | Thirteen methods, `Regexp`'s equality and both of its printings among them |
| `show the optional arguments the match methods take` | The start position, `split`'s limit, `start_with?`'s further patterns |
| `read the parse depth limit back beside the other two` | All three limits are constants, so the sentence naming them names three |

### The lists read as lists

The syntax and escape lists had grown paragraphs where an entry names a
spelling and what it stands for, so they had stopped reading as lists. Each
entry is a line again, and the three things that need more than a line are
stated below: how a quantifier after a quantifier reads, when a digit run is a
backreference and when it is octal, and which recursions are refused at compile
time.

Named captures move up beside the syntax, since how a named group renumbers the
whole pattern is part of what a pattern means, not something the engine
declines to do.

### Limitations answer, Configuration configures

A limitation entry now names what a pattern raises here and how CRuby answers
it, which is what a reader who has just hit one came for; the reasoning that
chose each refusal stays in the commit that made it. The step limit leaves the
list, because a limit a build sets is configuration rather than a difference
from CRuby, and Configuration already states it. So did the engine section,
which had spelled out both search limits a second time.

Configuration held the three limits, the case table and the ctype table under
one heading, and only the limits are a build's to set. Three headings now: the
two search limits, the parse depth limit, and what the build decides. A reader
looking for a knob stops before the two sections that have none.

### The gaps

Reading the file as answers made three absences visible, and each is filled
from the source and the test that covers it:

- Thirteen methods the gem defines stood in no list, among them the whole of
  `Regexp`'s equality (`==`, `eql?`, `hash`) and both of its printings (`to_s`,
  `inspect`). A reader asking whether the gem carries one had to read
  `regexp.c`.
- The parser takes `\h` and `\H`, a `?` after a `{n,m}` repeat, and `\b` as a
  backspace inside a character class. Each has a test of its own and none stood
  in the lists a reader goes to for the answer.
- Every spelling of `match` and `match?` takes a start position, `split` takes a
  limit and `start_with?` takes more than one pattern; the list showed each in
  its one-argument form only. The position is a character index, which is how
  the `index` entries beside them already read it; only the `byteindex` pair
  counts bytes.

## Testing

Documentation only: the change is confined to
`mrbgems/mruby-regexp/README.md`, and no source, test or build file is touched,
so there is nothing to build differently and no size or speed figure to report.

Every entry added was checked against the code that implements it and the test
that covers it: the method lists against the `mrb_define_method` calls in
`src/regexp.c`, the three spellings against `test/regexp_syntax.rb`, and the
optional arguments against the `mrb_get_args` format each method uses.

## Summary by CodeRabbit

* **Documentation**
  * Reorganized and condensed the mruby-regexp README for improved clarity.
  * Added documentation for hexadecimal character escapes and pattern interpretation rules.
  * Expanded examples covering regular expressions, match data, strings, and symbols.
  * Clarified engine architecture, named captures, limitations, configuration, search limits, and parse-depth settings.
  * Added guidance for comparing behavior with CRuby.

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -91,8 +91,10 @@ re = Regexp.new("pattern", Regexp::IGNORECASE)
 re = Regexp.compile("pattern")    # Regexp.new under its other name
 re = /pattern/i                   # literal syntax
 re.match("string")                # => MatchData or nil
+re.match("string", pos)           # => same, searching from pos
 re.match("string") { |md| ... }   # => block result, or nil if no match
 re.match?("string")               # => true/false
+re.match?("string", pos)          # => same, searching from pos
 re =~ "string"                    # => index or nil
 re === "string"                   # => true/false (for case/when)
 re.match(:symbol)                 # a Symbol is matched against its name
@@ -132,8 +134,10 @@ md.regexp                         # => the Regexp that matched
 
 # String methods
 str.match(re)                     # => MatchData or nil
+str.match(re, pos)                # => same, searching from pos
 str.match(re) { |md| ... }        # => block result, or nil if no match
 str.match?(re)                    # => true/false
+str.match?(re, pos)               # => same, searching from pos
 str =~ re                         # => index or nil
 str.sub(re, replacement)          # replace first occurrence
 str.sub(re) { |m| ... }           # replace with block
@@ -145,6 +149,8 @@ str.gsub!(re, replacement)        # => self, or nil if no match
 str.gsub!(re) { |m| ... }         # same, replacing with the block result
 str.scan(re)                      # => array of matches
 str.split(re)                     # => array of parts
+str.split(re, limit)              # => same, at most limit parts; a negative
+                                  #    limit keeps the trailing empty ones
 str[re]                           # => matched substring or nil
 str[re, capture]                  # => capture by index or name
 str.slice(re)                     # => same as str[re]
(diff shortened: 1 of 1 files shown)


matz:
Merged as 41e6b6c70 (rebased onto master). Thanks!

PR #7424, opened 2026-08-30
guard mrb_int overflow in String#slice!

saddamr3e:
UBSan, `"abc".slice!(1, 9223372036854775807)`:

    string.c:2349:11: runtime error: signed integer overflow: 1 + 9223372036854775807 cannot be represented in type 'mrb_int'
    string.c:2357:64: runtime error: signed integer overflow: 1 + 9223372036854775807 cannot be represented in type 'mrb_int'

`beg` and `len` come straight from `mrb_as_int`, so `beg + len` (and later `byte_beg + byte_len`) can wrap. The wrapped negative slips past the `> str_len` clamp, `len` stays at `MRB_INT_MAX`, and `mrb_str_new` is reached with it. On a 64-bit `mrb_int`, `str_check_length` rejects the length; on `MRB_INT32` the same input passes it and `memcpy` reads ~2GB past a small source.

Guard both adds with `mrb_int_add_overflow`, the same form `mrb_str_aset` and `str_bytesplice` already use in `src/string.c`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where using an extremely large length with `String#slice!` could cause integer overflow.
  * Large slice lengths are now safely clamped to the available string content.

* **Tests**
  * Added coverage for slicing with a length near the maximum supported integer value.

### DIFF
--- a/mrbgems/mruby-string-ext/src/string.c
+++ b/mrbgems/mruby-string-ext/src/string.c
@@ -4,6 +4,7 @@
 #include <mruby/class.h>
 #include <mruby/string.h>
 #include <mruby/range.h>
+#include <mruby/numeric.h>
 #include <mruby/internal.h>
 
 #define ENC_ASCII_8BIT "ASCII-8BIT"
@@ -2346,7 +2347,8 @@ mrb_str_slice_bang(mrb_state *mrb, mrb_value self)
   }
 
   if (beg > str_len) return mrb_nil_value();
-  if (beg + len > str_len) {
+  mrb_int end;
+  if (mrb_int_add_overflow(beg, len, &end) || end > str_len) {
     len = str_len - beg;
   }
   if (len < 0) len = 0;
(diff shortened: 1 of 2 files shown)


matz:
Merged as 706cc0f47, with the source half following in c31f98cf1 — my mistake, the merge took your test and left `src/string.c` behind, so CI ran the new assertion against unguarded code for one commit. Sorry for the noise on your PR.

The bug is worse than the report says, and the report was right to look. Your original input stops at `ArgumentError` even on `MRB_INT32`, because `str_check_length()` rejects a length of exactly `MRB_INT_MAX`. Move `beg` to 2 and it slips past:

```
$ mruby -e '"abc".slice!(2, 2147483646)'   # MRB_INT32
ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 2147483646
  #1 str_init_normal_capa ./src/string.c:76
```

Three bytes of source, two gigabytes read. On 64-bit the same shape only fails the allocation. A guard that tests the exact boundary is one the caller steps around by one.

I also swept `src/string.c`, `mruby-string-ext` and `mruby-regexp` for the same raw addition. `re_byte_substr()` in regexp.c has the shape, but every caller passes MatchData capture offsets bounded by the subject, so it cannot overflow. Yours was the only one.

One change to your test. It guarded the width with `(1 << 63) - 1` under `rescue RangeError`, which works without bigint but not with it: on `MRB_INT32` **with** mruby-bigint the expression answers with a big integer instead of raising, the rescue never runs, and `slice!` then raises `RangeError` for the argument size — `Crash: 1`. CI does not build that combination, so it would have stayed hidden. It now finds the width the way `mruby-random`'s tests do, by asking whether the value is one an `mrb_int` holds. Green on all four of 64/32-bit crossed with bigint present/absent.

Thanks for the report and the fix.

saddamr3e:
Thanks for taking it, and for the test fix. The bigint interaction on MRB_INT32 is one I hadn't considered, and the beg=2 variant slipping past str_check_length is a good catch. Glad the guard covers both.

PR #7425, opened 2026-08-31
mruby-struct: take a recursive pair as equal in Struct#== and #eql?

takumin:
## Summary
- A struct whose member references itself always compared unequal: the recursion guard in `mrb_struct_equal` / `mrb_struct_eql` answered false for the pair currently being compared
- Take that pair as equal, as in CRuby's `recursive_equal()`, and let the remaining members decide the outcome, so structs that differ only in their self reference are told apart
- Add a test covering `#==` and `#eql?` on self-referential structs

## Test plan
- [x] `make test` — all 2553 tests pass (2490 OK, 0 KO)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed equality checks for self-referential structs.
  * Recursive structs now correctly compare as equal when their contents match and unequal when members differ.
  * Updated both `==` and `eql?` behavior to align with standard Ruby semantics.

* **Tests**
  * Added coverage for reflexive and structural comparisons involving recursive structs.

### DIFF
--- a/mrbgems/mruby-struct/src/struct.c
+++ b/mrbgems/mruby-struct/src/struct.c
@@ -603,9 +603,10 @@ mrb_struct_equal(mrb_state *mrb, mrb_value s)
     return mrb_false_value();
   }
 
-  /* Check for recursion */
+  /* A pair already being compared is taken as equal, as in CRuby's
+     recursive_equal(), and the other members decide the outcome. */
   if (MRB_RECURSIVE_BINARY_FUNC_P(mrb, MRB_OPSYM(eq), s, s2)) {
-    return mrb_false_value();
+    return mrb_true_value();
   }
 
   mrb_int len = RSTRUCT_LEN(s);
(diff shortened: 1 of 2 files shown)


matz:
Merged as d488e4e55, with Array and Hash folded into the same commit.

Your reading of the guard is right, and it turned out not to be Struct's alone: `src/array.c` and `src/hash.c` carry the same `MRB_RECURSIVE_BINARY_FUNC_P` check answering false, in four more places.

```ruby
a = [1]; a << a
b = [1]; b << b
a == b        # false here, true in CRuby

h = {x: 1}; h[:s] = h
i = {x: 1}; i[:s] = i
h == i        # false here, true in CRuby
```

The one-line change you made fixes all six. I checked the shapes where the assumption could paper over a real difference and none of them does — a cycle through two arrays rather than one, a self reference nested inside an element, a Struct held by an Array, a Hash whose value is a self-referential Array, and lengths that differ (`[1, x]` against `[1, 2, x]`, still false). Seven shapes, all matching CRuby.

Tests for the Array and Hash halves went into `test/t/array.rb` and `test/t/hash.rb`; reverting any one of the three sources turns exactly its own test red.

Thanks — the fix and the reasoning both carried further than the PR claimed.

PR #7426, opened 2026-08-31
Use `mrb_args_pack_positional()` in `prepare_missing()`

dearblue:
Also, removed the `blk` parameter, which is no longer used directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `Kernel#method_missing` argument forwarding.
  * Positional, splatted, keyword, and block arguments are now correctly preserved when handling missing methods.

* **Tests**
  * Added coverage verifying that method names, arguments, keyword arguments, and block results are forwarded correctly.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1369,7 +1369,7 @@ mrb_ci_nregs(mrb_callinfo *ci)
 mrb_value mrb_obj_missing(mrb_state *mrb, mrb_value mod);
 
 static mrb_method_t
-prepare_missing(mrb_state *mrb, mrb_callinfo *ci, mrb_value recv, mrb_sym mid, mrb_value blk, mrb_bool super)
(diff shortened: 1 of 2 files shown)


matz:
Merged as 4207b954e.

I checked the refactor against thirteen `method_missing` shapes on both sides — no arguments, packed and unpacked positional (15 and 16, either side of `CALL_MAXARGS`), keywords alone and with positionals, each of those with a block, `**{}`, and `send`/`public_send` — plus the `NoMethodError` and `super` messages. Byte-identical output before and after.

The part worth a second look was `stack_extend()` going away, since the old code re-read `argv` after it in case the stack moved. `mrb_args_pack_positional()` reads the arguments before anything can move them, so there is nothing left to re-read. A stress run under ASan and UBSan (900 `method_missing` calls with varying arity, keywords and blocks, some at 60 frames deep, with a collection at the end) reports nothing.

Green on all five ci/gcc-clang builds including full-debug's `MRB_GC_STRESS`, plus bintest.

PR #7428, opened 2026-08-31
mruby-regexp: read the last participating group for a duplicate name

takumin:
Stacked on #7427 (its commit is included in this branch).

When a pattern gives one name to several capture groups (e.g. the
alternatives in /(?<x>a)|(?<x>b)/), CRuby's named accessors read the last
group that took part in the match. re_name_to_group() returned the first
entry in the pattern's named-capture table instead, so
/(?<x>a)|(?<x>b)/.match("b")[:x] came back nil instead of "b".

Resolve the name against the match's captures in the shared lookup: the
candidates are walked back to front and the first one that participated
is the answer; when none of them took part the last of them stands in,
which the callers read as a non-participating group (nil or nothing).
Since the lookup is shared, this fixes every named accessor at once:

- MatchData#[] with a String/Symbol
- MatchData#begin and #end
- MatchData#values_at (added in #7427)
- \k<name> in replacement strings

Verified against CRuby.

## Summary by CodeRabbit

* **New Features**
  * Added `MatchData#values_at` for retrieving multiple captured values by index, name, or range.
  * Improved handling of duplicate named capture groups in match results and replacement strings.

* **Bug Fixes**
  * Corrected behavior when resolving named captures, including unmatched and repeated named groups.

* **Documentation**
  * Added usage documentation for `MatchData#values_at`.

### DIFF
--- a/mrbgems/mruby-regexp/README.md
+++ b/mrbgems/mruby-regexp/README.md
@@ -120,6 +120,7 @@ md[1]                             # => "user"
 md[2]                             # => "host"
 md[:name]                         # named capture access
 md.captures                       # => ["user", "host"]
+md.values_at(1, 2)                # => ["user", "host"]
 md.to_a                           # => ["user@host", "user", "host"]
 md.to_s                           # => "user@host" (same as md[0])
 md.size                           # => group count, group 0 included
--- a/mrbgems/mruby-regexp/src/regexp.c
+++ b/mrbgems/mruby-regexp/src/regexp.c
@@ -11,6 +11,7 @@
 #include <mruby/array.h>
 #include <mruby/variable.h>
 #include <mruby/hash.h>
+#include <mruby/range.h>
 #include <mruby/error.h>
 #include <mruby/internal.h>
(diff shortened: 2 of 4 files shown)


matz:
Merged as 59c7fc7e7.

Reproduced on master first: six of the ten shapes I tried answered wrong, and your list of affected accessors is exactly right — `#[]`, `#begin`, `#end`, `#values_at` and `\k<name>` in a replacement all went through the same lookup. `#named_captures` was already correct, which is why the bug could sit there looking narrow.

After the change, all ten match CRuby. I also went deeper on the duplicate-name shapes: a name repeated inside a quantifier (`/(?<r>.)+/`), three alternatives with only the middle one participating, a duplicate where the winning group matches empty, duplicates where *both* participate (nested and adjacent), `captures`, `size`, `to_a` and `names`. Eleven more shapes, all matching CRuby except `MatchData#offset`, which mruby does not have at all.

The gem's 5,175-pattern difftest corpus is byte-identical before and after, so nothing outside duplicate names moved.

PR #7429, opened 2026-08-31
Change `mrb_callinfo::nk` to `kw`

dearblue:
This means separating the semantic representation from the third operand of the OP_SEND* instruction.
Since its introduction, `mrb_callinfo::nk` has always had a value of either 0 or 15 (CALL_MAXARGS).
The reason for this is that when an OP_SEND* instruction is issued, the called method can only observe that either no keyword arguments exist or that packed keyword arguments exist.

This change also eliminates the need for related assertions.

Furthermore, it corrects errors in assertions and comments that assumed `ci->nk` to be in the range 1..14.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyword-argument handling throughout method calls, evaluation, blocks, and data operations.
  * Fixed argument parsing and block/environment indexing for calls that include keyword arguments.
  * Preserved packed-argument behavior while simplifying keyword-argument presence tracking.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -172,8 +172,8 @@ typedef struct mrb_irep mrb_irep;
 #endif
 
 typedef struct {
-  uint8_t n:4;                  /* (15=*) c=n|nk<<4 */
-  uint8_t nk:4;                 /* (15=*) */
+  uint8_t n:4;                  /* number of positional arguments; 15 means packed arguments */
+  uint8_t kw:1;                 /* has keyword arguments (TRUE or FALSE) */
   uint8_t cci;                  /* called from C function */
   uint8_t vis;                  /* 4(ZERO):1(module_function):1(separate module):2(method visibility) */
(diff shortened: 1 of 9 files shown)


matz:
Merged as 1a7572bcf.

`sizeof(mrb_callinfo)` is unchanged at 48 bytes, which was the first thing I checked — the field is in a bitfield that had room.

For behaviour I ran 29 keyword-argument shapes: required and optional keywords, `**rest`, splat at 15 and 16 arguments (either side of `CALL_MAXARGS`), splat combined with keywords and a block, `**{}` and `**h`, `send`, `public_send`, `Method#call`, lambda, proc, `method_missing`, and `Struct` with `keyword_init`. Zero rows differ from master. A stress run under ASan and UBSan (1,800 calls varying arity, keywords and blocks, with a collection) reports nothing.

Green on all five ci/gcc-clang builds plus bintest, and the amalgam still compiles — worth checking since `include/mruby.h` is in the diff.

PR #7430, opened 2026-08-31
mruby-struct: let a bare Struct.new answer a memberless struct

takumin:
## Summary

- `Struct.new` with no arguments raised `ArgumentError: wrong number of arguments (given 0, expected 1+)`, where CRuby answers a memberless struct
- Drop the arity check in `mrb_struct_s_def()` so a bare `Struct.new` yields a class with no members, as in CRuby
- Add a test pinning the memberless `Struct.new` and the size-0 object it builds

## Changes

| File | Change |
| --- | --- |
| `mrbgems/mruby-struct/src/struct.c` | `mrb_struct_s_def()` no longer refuses a zero-argument call; the empty member list flows through the unchanged duplicate scan and accessor definition, which both no-op at length zero |
| `mrbgems/mruby-struct/test/struct.rb` | pins `Struct.new` with no arguments: superclass `Struct`, `members` `[]`, a `new` of size 0 |

## Behaviour

```ruby
# CRuby 4.0.6, and mruby with this PR
c = Struct.new
c.superclass    # => Struct
c.members       # => []
c.new.size      # => 0
c.new.members   # => []

# mruby before
Struct.new      # => ArgumentError: wrong number of arguments (given 0, expected 1+)
```

A call that names one or more members is untouched: the member list is read exactly as before.

## Testing

- [x] `rake test` — all 2561 tests pass (2498 OK, 0 KO, 0 crashes; 63 skipped, unrelated to this change)
- [x] Ran a bare `Struct.new` directly against `bin/mruby` to confirm the memberless struct above

## Summary by CodeRabbit

* **Bug Fixes**
  * `Struct.new` now works without arguments, creating an anonymous struct with no members.
  * Instances of empty structs correctly report zero members and a size of zero.

* **Tests**
  * Added coverage for creating and using empty structs.

### DIFF
--- a/mrbgems/mruby-struct/src/struct.c
+++ b/mrbgems/mruby-struct/src/struct.c
@@ -272,9 +272,6 @@ mrb_struct_s_def(mrb_state *mrb, mrb_value klass)
   mrb_value keyword_init_val = mrb_nil_value();
 
   mrb_get_args(mrb, "*&", &argv, &argc, &b);
-  if (argc == 0) {
-    mrb_raise(mrb, E_ARGUMENT_ERROR, "wrong number of arguments (given 0, expected 1+)");
-  }
 
   /* Check for keyword_init option in arguments */
   if (argc > 0 && mrb_hash_p(argv[argc-1])) {
--- a/mrbgems/mruby-struct/test/struct.rb
+++ b/mrbgems/mruby-struct/test/struct.rb
@@ -11,6 +11,15 @@ assert('Struct.new', '15.2.18.3.1') do
   assert_equal [:m1, :m2], c.members
 end
 
+assert('Struct.new with no arguments', '15.2.18.3.1') do
+  c = Struct.new
+  assert_equal Struct, c.superclass
+  assert_equal [], c.members
+  o = c.new
+  assert_equal [], o.members
+  assert_equal 0, o.size
+end
+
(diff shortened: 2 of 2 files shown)


matz:
Merged as 3d13b7eb7.

Worth recording why the arity check can go, since the answer is not the same in every CRuby: the ruby_3_2 branch reads its arguments with `rb_scan_args(argc, argv, "1*:", ...)` — one required — and raises exactly the error this gem raised. On master it is `"0*:"`, zero required, and `rb_struct_s_def()` carries no arity check of its own. So the PR is tracking the newer spelling, not a longstanding one, and a reader comparing against an older ruby would find the opposite.

I checked sixteen shapes around the change. Everything that moved went from `ArgumentError` to an answer — `superclass`, `members`, `size`, `to_a`, `to_h`, `each`, `inspect` (`"#<struct >"`), `new` with a block, and `new(1)` now reaching the real `struct size differs` — and nothing that worked before changed. `==` between two instances of one memberless class is `true`, which matches CRuby: `rb_struct_equal()` compares `rb_obj_class` first, so two *different* memberless classes still answer `false`, and they do here.

A stress run under ASan and UBSan (500 memberless structs plus structs of 0 to 5 members, with a collection) reports nothing. Green on all five ci/gcc-clang builds plus bintest, and reverting `struct.c` alone turns the new assertion red.

PR #7435, opened 2026-09-01
io.c: read only WSAENOTSOCK as "not a socket" on Windows

matz:
The AppVeyor `Visual Studio 2019 32-bit` job has been failing since build
9150, intermittently, on one test:

```
Errno::EBADF: UDPSocket.new(AF_INET6) => Bad file descriptor - bad file descriptor
  Total: 2783
     OK: 2736
     KO: 0
  Crash: 1
```

The other two AppVeyor jobs and all 22 GitHub Actions checks stay green, so
this is specific to 32-bit Windows.

### Where it comes from

The message is `badfd_error()`'s, which `check_file_descriptor()` in
`mrbgems/mruby-io/src/io.c` raises. `UDPSocket#initialize` reaches it
because `BasicSocket#initialize` calls `super` (so `IO#initialize`, so the
descriptor check) before `self._is_socket = true`, leaving the check to
judge a raw Winsock handle.

That is why the check asks Winsock first: a Winsock handle is not a CRT
file descriptor, and the CRT probes after it would call a perfectly good
socket bad. But it read *any* `getsockopt` failure as the answer "not a
socket". `SO_ERROR` is refused on a socket that is not bound yet by some
stacks, so the socket fell through to the CRT fallback and was rejected as
`EBADF`.

Only `WSAENOTSOCK` reports that the value is not a socket. Every other
error is a socket declining that particular question, so this keeps it.

### Verification

I have no Windows machine, so the point of this PR is AppVeyor's
`Visual Studio 2019 32-bit` job. The change is inside `#ifdef _WIN32`;
`rake CONFIG=host-debug all test:run:serial` on Linux stays at
`KO: 0 / Crash: 0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows compatibility when working with valid Winsock socket handles.
  * Prevented healthy sockets from being incorrectly rejected during file descriptor checks.

### DIFF
--- a/mrbgems/mruby-io/src/io.c
+++ b/mrbgems/mruby-io/src/io.c
@@ -562,6 +562,14 @@ check_file_descriptor(mrb_state *mrb, mrb_int fd)
     if (getsockopt(fdi, SOL_SOCKET, SO_ERROR, (char*)&err, &len) == 0) {
       return;
     }
+    /* A Winsock handle is not a CRT file descriptor, so the probes below
+       cannot vouch for one and would call it bad. Only WSAENOTSOCK says
+       the value is not a socket; every other error is a socket that would
+       not answer this particular question, and treating that as "not a
+       socket" fails IO.new on a healthy one. */
+    if (WSAGetLastError() != WSAENOTSOCK) {
+      return;
+    }
   }
 
   if (fdi < 0 || fdi > _getmaxstdio()) {
--- a/mrbgems/mruby-socket/test/udpsocket.rb
+++ b/mrbgems/mruby-socket/test/udpsocket.rb
@@ -1,26 +1,3 @@
-# TEMPORARY diagnostic, to be removed before this branch is merged. The
-# 32-bit Windows job rejects UDPSocket.new(Socket::AF_INET6) with EBADF from
-# IO#initialize's descriptor check while the AF_INET case beside it passes,
-# and the descriptor Socket._socket hands over is the only thing that differs.
-# Print it.
-assert('DIAGNOSTIC descriptors') do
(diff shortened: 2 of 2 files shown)


matz:
Closing this. The change rested on a wrong reading of the failure.

I assumed `getsockopt(SO_ERROR)` was failing with something other than
`WSAENOTSOCK`, so that `check_file_descriptor()` was mistaking a healthy
socket for a non-socket. Reading only `WSAENOTSOCK` as "not a socket" did not
stop the `Errno::EBADF`, and printing the descriptor from the failing job
showed why: `Socket._socket(Socket::AF_INET6, ...)` succeeded, and the
descriptor it returned wrapped and closed without complaint. Winsock really
was answering `WSAENOTSOCK`, and it was right to. The handle had already been
closed.

The double close was in this gem. `BasicSocket#close` was overridden on
Windows to call `closesocket()`, and it left the descriptor in the `mrb_io`
as it was, so the object still answered `#closed?` with false and
`fptr_finalize()` closed it a second time when it was collected. Windows
hands socket handles out of a pool, so by then the number could name a socket
someone else had opened since.

4af596835 drops that override: `fptr_finalize()` has called `closesocket()`
for a socket on Windows since 949bf6ca4 and clears the descriptor after, and
`IO#close` returns early once it is clear, so there was nothing left for the
override to add. 954e172c7 follows up on two hazards the same investigation
turned up in `check_file_descriptor()`: `_getmaxstdio()` is the stdio stream
limit and says nothing about which descriptors the CRT has, and asking the
CRT about one it does not have ended the process, no invalid parameter
handler being installed.

PR #7491, opened 2026-09-04
task: let an embedder disable the scheduler for one mrb_state

Asmod4n:
mruby-task makes every VM in the process 30% slower. This function disables the scheduler for a VM that does not use tasks, reducing the cost of adding mrb-task as a dependency to 1% instead of 30% for a mrb_state which doesn’t plan to use tasks.

## Summary by CodeRabbit

* **New Features**
  * Added controls to check whether task scheduling is enabled and permanently disable it before tasks are created.
  * Task scheduling is enabled by default.
  * Task creation now reports a `Task::Error` when scheduling is disabled.
  * Disabling the scheduler after tasks exist is rejected with an error.
* **Bug Fixes**
  * Task switching and stopping now respect the scheduler’s enabled state.

### DIFF
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -327,15 +327,16 @@ enum mrb_idx_op_slot {
 struct mrb_task;
 
 typedef struct mrb_task_state {
-  /* Read once per dispatched instruction, so it is the first field: a VM
-     that will never run tasks says so with mrb_disable_task_scheduler() and the
-     dispatch loop's check costs one test of an already-hot word instead
-     of two dependent loads. */
-  mrb_bool enabled;
   struct mrb_task *queues[4];      /* Task queues (dormant, ready, waiting, suspended) */
   volatile uint32_t tick;           /* Current tick count */
   volatile uint32_t wakeup_tick;    /* Next wakeup tick */
   volatile mrb_bool switching;      /* Context switch pending flag */
+  /* Read once per dispatched instruction: a VM that will never run tasks
+     says so with mrb_disable_task_scheduler(), and the dispatch loop's
+     check costs one test of an already-hot word instead of two dependent
+     loads. It sits HERE, beside switching, because the check loads that
+     word first anyway - and because padding already wasted this byte. */
+  mrb_bool enabled;
   struct mrb_task *main_task;       /* Main task wrapper for root context */
   uint8_t scheduler_lock;           /* Lock counter for synchronous execution */
   uint8_t irq_nesting;              /* Depth counter for scheduler-IRQ exclusion */
--- a/include/mruby.h
+++ b/include/mruby.h
@@ -327,6 +327,11 @@ enum mrb_idx_op_slot {
 struct mrb_task;
 
 typedef struct mrb_task_state {
+  /* Read once per dispatched instruction, so it is the first field: a VM
+     that will never run tasks says so with mrb_disable_task_scheduler() and the
+     dispatch loop's check costs one test of an already-hot word instead
+     of two dependent loads. */
+  mrb_bool enabled;
   struct mrb_task *queues[4];      /* Task queues (dormant, ready, waiting, suspended) */
   volatile uint32_t tick;           /* Current tick count */
   volatile uint32_t wakeup_tick;    /* Next wakeup tick */
--- a/mrbgems/mruby-task/include/task.h
+++ b/mrbgems/mruby-task/include/task.h
@@ -124,6 +124,8 @@ MRB_API void mrb_tick(mrb_state *mrb);
 /* Take a VM out of the scheduler for good; see task.c for what it saves.
  * After mrb_open(), before anything runs. Raises if tasks already exist. */
 MRB_API void mrb_disable_task_scheduler(mrb_state *mrb);
+
+/* TRUE while this VM still runs tasks. A VM starts enabled. */
 MRB_API mrb_bool mrb_task_scheduler_enabled_p(mrb_state *mrb);
 MRB_API mrb_value mrb_task_run(mrb_state *mrb);
 MRB_API mrb_value mrb_task_run_once(mrb_state *mrb);
--- a/mrbgems/mruby-task/include/task.h
+++ b/mrbgems/mruby-task/include/task.h
@@ -120,6 +120,11 @@ void mrb_task_mark_all(mrb_state *mrb);
  * Core task scheduler API
  */
 MRB_API void mrb_tick(mrb_state *mrb);
+
+/* Take a VM out of the scheduler for good; see task.c for what it saves.
+ * After mrb_open(), before anything runs. Raises if tasks already exist. */
+MRB_API void mrb_disable_task_scheduler(mrb_state *mrb);
+MRB_API mrb_bool mrb_task_scheduler_enabled_p(mrb_state *mrb);
 MRB_API mrb_value mrb_task_run(mrb_state *mrb);
 MRB_API mrb_value mrb_task_run_once(mrb_state *mrb);
 
--- a/mrbgems/mruby-task/src/task.c
+++ b/mrbgems/mruby-task/src/task.c
@@ -531,6 +531,42 @@ execute_task(mrb_state *mrb, mrb_task *t)
 }
 
 /* Tick handler - called by timer interrupt */
+/* Task::Error, resolved from this VM. Both refusals below raise it, so a
+   caller can rescue the scheduler's own answer instead of matching on
+   RuntimeError and catching whatever else went wrong. */
+static struct RClass *
+task_error_class(mrb_state *mrb)
+{
+  return mrb_class_get_under_id(mrb, mrb_class_get_id(mrb, MRB_SYM(Task)), MRB_SYM(Error));
+}
+
+/* Take this VM out of the scheduler for good.
+ *
+ * A process may hold VMs of both kinds - a pool whose workers run tasks
+ * so a runaway callback can be preempted, and one VM that never does.
+ * The second kind still paid for the first: RETURN_IF_TASK_STOPPED is
+ * expanded at every dispatch, and its context-status arm is read whether
+ * or not any task exists. Disabled, that check is one test of a word the
+ * dispatch loop already has.
+ *
+ * Call it after mrb_open() and before running anything. It refuses once
+ * tasks exist rather than stranding them, since a queue nobody ticks is
+ * a hang, not a saving. */
+MRB_API void
+mrb_disable_task_scheduler(mrb_state *mrb)
+{
+  if (q_ready_ || q_waiting_ || q_suspended_ || q_dormant_) {
+    mrb_raise(mrb, task_error_class(mrb), "mrb_disable_task_scheduler: this VM already has tasks");
+  }
+  mrb->task.enabled = FALSE;
+}
+
+MRB_API mrb_bool
+mrb_task_scheduler_enabled_p(mrb_state *mrb)
+{
+  return mrb->task.enabled;
+}
+
 MRB_API void
 mrb_tick(mrb_state *mrb)
 {
@@ -926,6 +962,16 @@ static mrb_task*
 task_create_common(mrb_state *mrb, const struct RProc *proc,
                    mrb_value name, uint8_t priority)
 {
+  /* Every task is born here - Task.new, mrb_create_task, the C API - so
+     this is the one place a disabled VM has to say no. A task made in a
+     VM the dispatch loop no longer checks would never be preempted and
+     never be switched away from: a hang, arriving later and somewhere
+     else. mrb_disable_task_scheduler() refuses when tasks already exist
+     for the same reason, from the other side. */
+  if (!mrb->task.enabled) {
+    mrb_raise(mrb, task_error_class(mrb), "tasks are disabled for this mrb_state");
+  }
+
   mrb_task *t = task_alloc(mrb);
   t->priority = priority;
   t->status = MRB_TASK_STATUS_READY;
@@ -1884,6 +1930,10 @@ mrb_mruby_task_gem_init(mrb_state *mrb)
 
   /* Initialize HAL (timer and interrupts) */
   mrb_hal_task_init(mrb);
+
+  /* On unless the embedder says otherwise: a VM that will never run
+     tasks calls mrb_disable_task_scheduler() right after mrb_open(). */
+  mrb->task.enabled = TRUE;
 
   /* Initialize main task to NULL and scheduler_lock to 0 */
   mrb->task.main_task = NULL;
--- a/mrbgems/mruby-task/test/task.rb
+++ b/mrbgems/mruby-task/test/task.rb
@@ -430,3 +430,37 @@ assert("closure escaping a synchronously executed proc survives GC") do
   assert_equal "sync-result", result
   $task_escaped_proc3 = nil
 end
+
+assert('a disabled VM refuses to make a task') do
+  # The VM under test has tasks, so it cannot be disabled here - what can
+  # be checked is the other half: TaskTest.with_tasks_disabled runs a
+  # block with the flag off, and Task.new must refuse rather than build
+  # something nothing will ever schedule.
+  caught = nil
+  TaskTest.with_tasks_disabled do
+    begin
+      Task.new { 1 }
+    rescue Task::Error => e
+      caught = e
+    end
+  end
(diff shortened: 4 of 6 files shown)


Asmod4n:
@coderabbitai please review again, your concerns are adressed.

matz:
I could not reproduce the 30%. Here is what I measured, so you can tell me
where my setup differs from yours.

### How I measured

The `#ifdef MRB_USE_TASK_SCHEDULER` means a build with the gem and a build
without it are two different binaries, and comparing those two is dominated
by code layout rather than by the check. In one pair of builds I got the
scheduler build running 24% MORE instructions yet taking FEWER cycles than
the build without it; in another pair the same comparison came out +9%. A 16
point swing from layout alone is larger than the effect being looked for, so
I stopped trusting that comparison.

Your patch gives a clean one. I built libmruby once and linked two drivers
against it that differ only in whether they call
`mrb_disable_task_scheduler()`. Same library, same layout, one runtime flag.

Machine: AMD Ryzen 9 7945HX, gcc 13.3.0, `-O2`, `default` gembox plus
`mruby-task`, load average 0.2, counters from `perf stat`.

### What it says

A tight `while` loop, 120M iterations, 16 alternating runs, medians:

| | enabled | disabled | change |
| --- | --- | --- | --- |
| instructions | 53.681G | 43.121G | **-19.67%** |
| cycles | 10.386G | 10.319G | -0.64% |
| task-clock | 2034ms | 2009ms | **-1.25%** |
| IPC | 5.17 | 4.18 | |

A mixed workload (`fib(29)`, Array operations, 400k Hash stores), 14
alternating runs:

| | enabled | disabled | change |
| --- | --- | --- | --- |
| instructions | 15.902G | 15.315G | -3.69% |
| cycles | 3.793G | 3.662G | -3.45% |
| task-clock | 740ms | 713ms | **-3.74%** |

So the instruction count falls by a fifth and the time falls by one to four
percent. The IPC line is why: what the check costs is a branch that always
goes the same way and a load of a word already in cache, and an out-of-order
core absorbs both into the dependency chain of the loop body. On an in-order
core it would not, and there the 19.67% would land on the clock much more
directly. If your 30% came from a core like that, your number is right and
mine is a statement about this one.

### What I would like to know

- Which CPU, and is it in-order?
- What the workload was.
- Was it wall time, or instructions or cycles?
- `mruby-task` is in no gembox, so a build only has the check if its config
  asks for the gem. Was the 30% against a build of the same config with the
  gem removed, or against something else?

### One note on the patch itself

`mrb_task_state` already wastes 11 bytes of padding: 7 between `switching`
and `main_task`, and 4 between `exception_as_result` and `scheduler_hook`.
Putting `enabled` first adds a new 7-byte hole and grows `mrb_state` from
24776 to 24784. Declaring it right after `switching` instead puts it at
offset 41, leaves `mrb_state` at 24776, and keeps the two flags in one word,
which is the word the check reads first anyway. I measured both layouts:
instruction counts agree to within 0.004%, and the hole placement was equal
or a shade faster in every condition. It passes here across the CI configs,
ASan, word boxing and MRB_INT32.

Asmod4n:
I’ve measured it against a new web server I am writing, responses per second  went down 30%

and then I wrote a while loop with 20 million runs and it also got 30% slower.

The build was with -march=native and -O3 and gcc 16.2

The issue was a dependent load, on a ryzen 9 5950x and on a xeon with avx 512 the same slowdown happened.

the numbers on the xeon where ~19 nanoseconds without the gem added, ~25-26 with and this patch does run at ~20ns

As a hint how fast my web server is: i hit 1 million none pipelined http 1 requests per second without entering the mruby vm, when i enter it once per request it gets 20% slower, with mruby-task added those numbers were 30% slower for each function call.

Asmod4n:
When RETURN_IF_TASK_STOPPED could be rewritten to not have dependent loads this patch wont be needed.

matz:
You were right and I was measuring the wrong build. `-O3 -march=native`
reproduces it here.

I had been building `-O2`. Rebuilt with your flags, same 20M `while` loop,
12 alternating runs, medians on a Ryzen 9 7945HX:

| | per iteration | vs no gem |
| --- | --- | --- |
| no `mruby-task` | 8.59 ns | |
| `mruby-task`, scheduler on | 10.57 ns | **+23.2%** |
| `mruby-task`, scheduler disabled | 9.38 ns | +9.2% |

Instructions +26.3%, cycles +22.7%, task-clock +22.9%. Close enough to your
30%, and the shape matches your Xeon numbers exactly: 19 -> 25/26 -> 20 there,
8.6 -> 10.6 -> 9.4 here.

Why `-O2` hid it: the dispatch loop there runs at IPC 4 to 5 already, and the
extra branch and load fit in the gaps between the loop body's own work, so the
instruction count moved 19.7% and the clock moved 1.3%. `-O3 -march=native`
packs the body tighter, the gaps are gone, and the same added instructions land
on the clock. I should have measured at the optimization level people ship at
before reporting a number, and I should have read your 04:53 comment before
posting mine. Sorry for making you repeat yourself.

That leaves your 05:19 point as the open question. You are right that this
patch is a workaround for the dependent loads rather than a fix for them, and
a rewrite that removes them would serve every build rather than the ones that
call `mrb_disable_task_scheduler()`. The check reads `mrb->task.switching`,
then `mrb->c`, then `mrb->c->status`, and it is that second hop that costs.
Whether the two can be folded into one hot word is worth trying before we
settle for a flag.

If it turns out they cannot, this patch stands on its own merits and I would
take it. In that case please move `enabled` to sit right after `switching`
rather than at the front of `mrb_task_state`: at the front it opens a fresh
7-byte hole and grows `mrb_state` from 24776 to 24784, while after `switching`
it lands at offset 41 in padding that is already wasted, keeps `mrb_state` at
24776, and puts the two flags in one word, which is the word the check loads
first anyway. I measured both layouts and they run the same.

Asmod4n:
Currently evaluating it what would need to happen to not a) make tasks slower and b) reducing their inline code sizes to not blow past L1 cache size.

Asmod4n:
Tried several different ways to make the vm faster when the Task code is compiled in.

Sadly found nothing that would preserve their current behavior while reducing their per run cost.
The only thing that would give a real speed boost is a different implementation: mirroring the state the check reads into one hot word, kept correct at every site that writes it — about 40 write sites across three gems, plus every HAL. It also introduces stale data as a failure mode: a missed update is not a slow path, it is a silent stall.

This change adds no additional speed penalty for adding the mruby-task gem, but gives embedders who dont need them for all vms in a process a noticeable speed increase.

Id say thats the better tradeoff.

Asmod4n:
Oh and one can use fibers again in a vm where tasks are present in the code but disabled

matz:
That one is worth pinning down, because it changes what this patch is for. A
percent or three of dispatch is one argument; getting a language feature back
is another entirely, and the second is the stronger one.

I could not reproduce it here, though. On a build with `mruby-task` and the
scheduler left on, these all behave the same as on a build without the gem:

- a plain `Fiber.new { Fiber.yield 1; ... }` resumed to completion
- a fiber created and finished after a `Task` exists
- a fiber created inside a task body
- resuming a finished fiber (`FiberError` either way), then carrying on
- a fiber finishing inside another fiber, with the outer one continuing past it

So whatever you hit is something I am not spelling. What did the failure look
like, and what was around it?

I ask because there is a real collision in the neighbourhood and I would like
to know whether it is the one you met. `MRB_TASK_STOPPED` is `#define`d to
`MRB_FIBER_TERMINATED`, the same value in the same 4-bit `status` field of
`struct mrb_context` that mruby-fiber writes. `RETURN_IF_TASK_STOPPED` reads
`mrb->c->status == MRB_TASK_STOPPED` on every dispatched instruction, so a
context a fiber has finished reads to the scheduler as a task that stopped.
The guard ahead of it excludes `mrb->c == mrb->root_c`, which is what keeps my
cases above working; a context that is neither the root nor a task is where I
would expect it to bite.

If that is it, then it is a bug in its own right rather than a cost, and the
fix is to stop overloading one field with two meanings, which would serve every
build rather than the ones that call `mrb_disable_task_scheduler()`. Same shape
as your point about the dependent loads: the flag makes the symptom avoidable,
and there may be something underneath worth fixing instead.

A reproducer would settle it. Even a rough one.

Asmod4n:
Ah, i just read the Readme of the mruby-task gem which says one cannot use fibers with it.
https://github.com/mruby/mruby/tree/master/mrbgems/mruby-task#relationship-with-fiber

matz:
That README line is vaguer than it should be, and my measurements say it
overstates the case. It reads "Tasks and Fibers both use `mrb_context` but are
**not compatible**" and then "Do not mix Tasks and Fibers in the same
application", without saying what breaks. You read that as fibers being
unavailable once the gem is in the build, which is a fair reading of it and is
not what happens: on a build with `mruby-task` and the scheduler on, every
fiber I tried behaves exactly as on a build without the gem, including fibers
created inside a task body.

What the README is pointing at, I think, is the collision I mentioned:
`MRB_TASK_STOPPED` is `#define`d to `MRB_FIBER_TERMINATED`, one value in one
4-bit `status` field that both mruby-fiber and the scheduler write. So a
context is either a fiber's or a task's, and a program that makes it both is
where the two stop being distinguishable. That is a narrower claim than "do not
use fibers", and it is the one worth writing down.

So this does not give your patch the second argument I thought it might. It
also means the README owes its readers a better sentence, which is on me rather
than on you, and I will fix it.

Which leaves us where your 05:19 comment put us: the dependent loads. If those
can be folded so the check reads one hot word instead of chasing `mrb->c`, that
serves every build, including the ones running tasks, and your patch is not
needed. If they cannot, the patch stands on its measurements, and I would ask
for `enabled` to sit after `switching` rather than at the front of the struct
so it costs no memory. I would rather try the first before settling for the
second.

Asmod4n:
The dependent load aint the issue actually, its codesize, once the task code is in the functions it blows past L1 cache sizes and makes the vm slower.

I tried to work around that by moving the task code to cold pages by telling the compiler to never inline them, made tasks and the vm 6% slower.

Then i tried a bitmask, no real win their either.

Then i rewrote the whole implementation and reduced the code which checks what tasks are supposed to do by moving the task scheduling code from NEXT to JUMP, that made tasks only a few % slower being inside the codebase cause we would only have to check the state 3 times per run instead of dozens of times and the codesize would only grow a little.

but when users would'nt use that code correctly  the vm could stall since we could end up in an endless loop and it would change how tasks behave, all HALs would need to be rewritten too.


All i really need tasks for is to interrupt the VM mid run, could something like that be added without needing to add that gem?

matz:
Code size is the right answer, and it explains a measurement of mine I had put
down to something else. Thank you for going and finding it.

I measured it here. `mrb_vm_exec` compiled `-O3 -march=native`:

| | size |
| --- | --- |
| without `mruby-task` | 67008 bytes |
| with it | 79662 bytes (**+18.9%**) |

And on the 20M loop, `L1-icache-load-misses` goes up **56%**. That is the part
I had missed: I was reading the IPC rise as the back end absorbing the extra
work, and asking whether the dependent loads were what cost. The front end was
where it went. Instructions +26%, cycles +23%, and the icache misses are why
those two track each other instead of the second staying flat.

Which also says your patch does what it does for a reason other than the one
its comment gives. Short-circuiting on `enabled` skips the work, but the
instructions are still in the function either way, so the size is unchanged and
so is the pressure. That the disabled path still costs 9% over a no-gem build
fits: the code is there whether or not it runs.

> All i really need tasks for is to interrupt the VM mid run, could something
> like that be added without needing to add that gem?

That is a much better question than the one this PR is answering, and I think
the answer is yes. There is no such thing today. `code_fetch_hook` is the only
per-instruction callback and it sits behind `MRB_USE_DEBUG_HOOK`, which is a
debugger's price rather than a server's. Nothing else lets an embedder say stop
from outside a running `mrb_vm_exec`.

What that wants is small: a flag the VM tests where it already tests something,
and a documented way for the check to land, which for a preemptive interrupt
means a signal handler or another thread setting it. Nothing like the scheduler,
the queues, the HAL or the tick. If it can be spelled in a few instructions on
the hot path, the size question does not arise the way it does here.

Tell me what your web server needs from it: is a check between instructions the
requirement, or would between method calls do? And what should happen when it
fires, an exception raised in the running code, or `mrb_vm_exec` returning to
you? Those two answers decide most of the design. I would rather add that than
a switch for a gem you are only carrying to get at one of its side effects.

Asmod4n:
I am running one thread which does all the IO and http work and also cheap mruby vm work and a pool of threads with one mrb_state each for compute heavy tasks I either send irep to them to work on or a c function pointer.

I also give users an argument how long such an irep or c function may work before it's getting canceled mid run.

The main thread can know the addresses of each mrb_state so all that would need would be a volatile value somewhere in mrb_state it could set which could be read by the vm as it does work and then being reset to its default by it.
When the c function runs it would have to read and reset the value to its default.

I guess a volatile value is the only thing which would fit mruby since otherwise we would need to add atomics
to the VM, the value just has to be word aligned for all platforms mruby supports so the instruction of setting and getting it could be done independently of other values in mrb_atate by the cpu. A late read should be okay here but what it should read should be the value it asks for and not something else in the struct when the load would be unaligned and then also touch the bytes surrounding the stop value.

A additional volatile int satisfies all of this.

Asmod4n:
Oh and that volatile int shouldn’t be near a bool and such so that a write form a foreign thread can’t damage those.

Asmod4n:
Sorry, fat fingers.

so my attempts at making mruby-task faster could speed up this change too: the vm should only read this new stop value at L_SENDB_SYM OP_JMP OP_JMPIF OP_JMPNOT OP_JMPNIL.
That way the vm should not get measurably slower and still be able to be preempt where it matters.

Asmod4n:
Sorry for spamming you, but i guess the best thing to do here would be to raise Interrupt when someone asks for a stop and make this a opt in thing.

mruby-task just stops executing and then returns nil, but then cleanup of a function cant work, aka ensure gets skipped by mruby-task during terminate.

Example:
```ruby
t = Task.new do
  begin
    10.times { Task.pass }
    puts "task body finished"
  ensure
    puts "ENSURE RAN"
  end
end

k = Task.new do
  3.times { Task.pass }
  puts "killer terminates it now"
  t.terminate
end

Task.run
puts "t.status = #{t.status}"
```
Output
```
killer terminates it now
t.status = DORMANT
``` 

A raised exception still runs the ensure block of a function. And cruby itself has the Interrupt exception too.

matz:
Implemented as you described, on master in a9151d778.

`mrb_vm_interrupt(mrb)` sets one flag and reads none, so it is safe from a
signal handler or another thread. The VM reads it at `L_SENDB_SYM` and at
`OP_JMP`, `OP_JMPIF`, `OP_JMPNOT` and `OP_JMPNIL`, exactly the five you named,
and nowhere else. It raises `RuntimeError` in the code that was running and
clears the flag, so a caller reads it as an ordinary exception and the VM is
usable afterwards.

What it does here, alarm(1) against `mruby -e`:

| script | result |
| --- | --- |
| `i = 0; while true; i += 1; end` | interrupted |
| `def f; 1; end; i=0; while i < 10**18; f; i+=1; end` | interrupted |
| `loop { }` | interrupted |
| `begin; while true; end; rescue => e; ... end` | `caught: RuntimeError` |
| `1 + 1` | ran to completion |
| after an interrupt, `40 + 2` on the same state | `42` |

The cost, `-O3 -march=native`, 20M iterations, 12 alternating runs, medians:

| | without | with |
| --- | --- | --- |
| instructions | 4.267G | 4.347G (+1.9%) |
| cycles | 0.909G | 0.876G (-3.7%) |
| task-clock | 179ms | 172ms (**-4.0%**) |

`mrb_vm_exec` also went from 67008 to 39060 bytes. I did not expect either and
would not claim them as a win: the five `goto L_RAISE` give gcc cold paths it
did not have, and it relaid the function out around them. What the numbers do
say is that placing the check where you said costs nothing measurable, which
was the point.

Your reading of what was wrong was right twice, and both times it was not what
I was measuring. Thank you for staying with it.

That leaves this PR. Nothing in it is needed for the interrupt, and its own
argument was the 30% you have now traced to code size, which short-circuiting
on a flag does not change since the instructions stay in the function either
way. Unless you want it for something else, I would close it. Say if you would
rather keep it open.

Asmod4n:
Thank you matz, ill close this.

PR #7603, opened 2026-09-15
Add refinements as an opt-in feature (MRB_USE_REFINEMENTS)

shugo:
This adds Ruby refinements to mruby, compiled in only when `MRB_USE_REFINEMENTS` is defined
(`conf.defines << 'MRB_USE_REFINEMENTS'`). Nothing changes in a build without the define.

## What is added

- `Module#refine`, `Module#refinements`, `main.using`, `Module#using`, `Module.used_modules`,
  `Module.used_refinements`
- class `Refinement` (a subclass of `Module`) with `#target`, `#refined_class` and
  `#import_methods`; `#include` and `#prepend` raise `TypeError`, as in Ruby 3.2+
- `#<refinement:C@M>` in `Module#to_s`
- `test/t/refinement.rb` (skipped when the define is off), a `doc/guides/mrbconf.md` entry, and
  updates to `doc/limitations.md` and `doc/guides/language.md`
- the define is enabled in `build_config/host-debug.rb` and in the CI `full-debug` build
  (`MRB_GC_STRESS`) so the feature is exercised

## Semantics

They follow CRuby: a refinement is active lexically from the `using` call to the end of the
file, or of the class or module body. A method defined before the `using` does not see it; a
block written in the scope does. A `refine` block sees all refinements of its owner module,
including ones written later. `super` in a refined method reaches the refined class, and
refinements come before the class's prepended modules.

Honored: sends made by the VM, the operator instructions, `__send__`, `send`, `public_send`,
`respond_to?`, `defined?(obj.m)`, and strings given to `eval`, `instance_eval` and
`class_eval`.

Not honored (documented): `Kernel#method`, `Module#instance_method`, `method_defined?`,
`Symbol#to_proc`, string interpolation's `to_s`, `==` asked by containers, and every
`mrb_funcall()` made from C.

`using` written in a block given a class to run under (`class_eval {}`, `Class.new {}`, a
`refine` block) is kept to that block, as in CRuby.

Known divergences (documented): `using` in an `eval` string installs into the enclosing file
or class body rather than the string; each `mrb_load_*` call is a separate file, so `using` in
one mirb line does not reach the next; at most 2047 scopes can be alive at once.

## Design

- A refinement is a module whose `super` is the class it refines, as in CRuby. `super` in a
  refined method therefore reaches that class through the frame's target class as it does
  today, `alias` in a refine block finds the class's methods, and `R.ancestors` reads
  `[R, C, ...]`. `include`/`prepend`/`extend` with a refinement as the argument raise
  `TypeError`, since `include_module_at` would otherwise splice the class's chain into a module.
- The refinements a scope sees are an `Array` (most recently activated first) that a scope proc
  carries by index. `struct RProc` is already at the object-size cap, so the index is packed
  into the flag bits only a cfunc proc uses (bits 0-6 and 16-19); no object grows. The Arrays
  live in a weak table on `mrb_state` (`refscopes`) that the GC clears after marking; storing
  into a proc goes through the write barrier.
- A method body copies its scope's Array at definition, so a `using` written after it does not
  reach it, while a block reads its scope's dynamically, as CRuby's cref does. `using` never
  mutates an Array in place; only the owner module's own live Array, which refine blocks share,
  is grown.
- Dispatch: at `L_SENDB`, when the name has ever been defined into a refinement (a bloom filter
  on `mid`) and the calling frame carries a scope, `mrb_vm_find_refined_method()` walks the
  ancestry asking each class's refinements before its own table, without reading or writing the
  method cache. Every other send is unchanged; a send loop without `using` measures the same as
  before.
- A refinement of a guarded operator (`Integer#+`, `String#[]`, `==` and the like) disarms that
  operator's fast path for good, the way CRuby disables its `opt_*` instructions. `bop_refresh`
  and `idx_op_refresh` keep the slot disarmed while its refined bit is set.
- `mrb_load_irep()` now marks the top proc it runs `MRB_PROC_CREF` with Object (or the
  context's class) as its class, as `mrb_load_exec()` already does for the proc it compiles, so
  a walk up the `upper` chain ends at the file rather than in whatever frame was running when
  the irep was read. Without it a `using` in one precompiled file leaked into the next (this is
  how `mrbtest` runs its files). This is the one change that applies to every build.

## Verification

| build | result |
|---|---|
| `host-debug` with the define | mrbtest 3096 OK / 0 KO, `-s` stress included |
| full-core + `MRB_GC_STRESS` + the define | 3097 OK / 0 KO |
| full-core without the define | 3069 OK / 0 KO, `refinement.rb` reports one skip |

Footprint (`-O3`, default gembox, x86_64): `libmruby.a` grows by about 9.8 KB (+0.55%),
`mrb_state` by 64 bytes; `struct RProc` and every other object stay the same size, and a program
that never calls `refine` or `using` allocates nothing extra.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Ruby refinement support through the `MRB_USE_REFINEMENTS` build option.
  * Added `Module#refine`, `using`, refinement introspection, and `Refinement#import_methods`.
  * Refinements now apply across method calls, operators, `send`, `super`, blocks, lambdas, and evaluated strings.
  * Added protections against unsupported refinement inclusion and prepending.

* **Documentation**
  * Updated guides to describe refinement availability, behavior, configuration, and limitations.

* **Tests**
  * Added coverage for refinement scoping, dispatch, inheritance, operators, evaluation, errors, and garbage collection.


shugo:
> _🟡 Minor_ · Scope the operator limitation. · `doc/guides/language.md:13-18`
> > `13-18`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > **Scope the operator limitation.** Refinement-enabled builds honor refined primitive operators and disable the related fast paths. Qualify this statement as applying to ordinary operator overrides or builds without `MRB_USE_REFINEMENTS`.

Operators of primitive classes can already be redefined without refinements (the operator instructions check for a redefinition and fall back to a send), so the guide was wrong on master too. I've rewritten both statements to say what actually happens: a redefinition, by def, alias, prepend, or a refinement in an MRB_USE_REFINEMENTS build, turns that operator's fast path off.

